PR #200, opened 2026-04-07
Update sandbox commands: env persists to config, setup reads from config

schurchleycci:
## Summary

Motivated by [this thread](https://circleci.slack.com/archives/C070XFJAKFB) — @webster suggested combining the manual sandbox flow into an auto-build command, and converting the manual steps into tests.

### New `chunk sandbox env setup` command
- Creates a sandbox, runs the environment setup steps from `.chunk/config.json` (written by `chunk sandbox env`), then either runs a given command or drops into an interactive SSH shell.
- Collapses the old "build → export tokens → create → exec" manual flow into one command.
- Reads the environment spec from `.chunk/config.json` by default; accepts stdin as an override.

### Persisted environment spec
- `chunk sandbox env` now saves the detected environment to `.chunk/config.json` (new `Environment` field on `ProjectConfig`) in addition to printing JSON to stdout.
- `chunk sandbox env setup` reads it back instead of re-running detection.

### Environment model: ordered `Steps`
- `envbuilder.Environment` replaces `install` / `system_deps` with ordered `[]Step{Name, Command}`.
- `cimg` sudo substitution and extra-dep install commands are baked into the steps so every consumer (Dockerfile writer, JSON output, remote exec) uses the same resolved commands.
- New `Environment.ResolvedImage()` helper returns `image:version` (or `""`).

### `sandbox build` modes
- `CHUNK_BUILD_COMMANDS=1` — run the env spec's steps directly as local shell commands instead of building a Docker image.
- `CHUNK_CAPTURE_COMMANDS=<path>` (or `1` for `build-steps.json` in `--dir`) — dump the steps as JSON without executing.

### `sandbox create`
- New `--quiet` / `-q` flag prints only the sandbox ID (clean for shell composition).
- Provider selection remains env-var-only (`CHUNK_SANDBOX_PROVIDER`, defaults to `e2b`) — no `--provider` flag.

### Exit code propagation
- New `sandbox.ExitError` carries the remote command's exit code.
- `main.go` exits silently with that code for bare `ExitError`s (output was already printed), and preserves the code for wrapped `ExitError`s while still printing the wrapper's message.
- `sandbox.RunStep` takes `(name, command string)` — no dependency from `internal/sandbox` back into `envbuilder`.

### Tests
- New `acceptance/sandbox_gaps_test.go` covers the `CHUNK_BUILD_COMMANDS` and `CHUNK_CAPTURE_COMMANDS` modes plus failure behavior.

## Test plan
- [x] `task build`, `task test`, `task lint` clean
- [x] `chunk sandbox env --dir .` detects and saves to `.chunk/config.json`
- [x] `chunk sandbox env setup --name test` reads from config without re-detecting
- [x] `chunk sandbox env | chunk sandbox env setup --name test` stdin override still works
- [x] `chunk sandbox create --name test -q` prints only the sandbox ID
- [x] `CHUNK_BUILD_COMMANDS=1 chunk sandbox env --dir . | chunk sandbox build --dir .` runs steps locally
- [x] Failing step in `sandbox build` CHUNK_BUILD_COMMANDS mode surfaces the step name in the error output


main.go
```diff
@@ -10,13 +10,20 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/usererr"
 )
 
+type exitCoder interface {
+	ExitCode() int
+}
+
 var version = "dev"
```
danmux, line 17: Should this use `errors.As` instead of a type assertion? If an `ExitError` ever gets wrapped (e.g. `fmt.Errorf(...: %w, exitErr)`), the type assertion silently falls through to `os.Exit(1)` losing the actual exit code.

internal/sandbox/sandbox.go
```diff
@@ -46,6 +58,28 @@ func AddSSHKey(ctx context.Context, client *circleci.Client, sandboxID, publicKe
...
 
+// StepResult holds the output from executing a single setup step.
+type StepResult struct {
+	Name     string
+	Stdout   string
+	Stderr   string
+	ExitCode int
+}
+
+// RunStep executes a single setup step and returns its result.
```
danmux, line 68: Could `RunStep` take `name, command string` instead of `envbuilder.Step`? This would avoid the new `internal/sandbox` -> `envbuilder` dependency. The `Step` type is only used to extract two strings.

danmux (approved):
Reviewed -- approving.

schurchleycci:
Closing this in favour of a more targeted PR for env persistence

PR #224, opened 2026-04-13
feat: offer shell completion install during chunk init

meeech:
## Summary

- Extracts `completionInstalled()` and `installCompletion()` helpers from the `completion install` command so both `chunk completion install` and `chunk init` can reuse the logic
- Adds Step 5 to `chunk init`: detects shell compatibility, checks if already installed, prompts with `tui.Confirm`, installs on acceptance
- New `--skip-completions` flag for non-interactive / CI use
- Follows the same graceful-failure pattern as other init steps (warn and continue, never fatal)

## Test plan

- [x] `task build` compiles cleanly
- [x] `task test` — new tests in `init_test.go` pass (completionInstalled, installCompletion)
- [x] `task lint` — no issues
- [x] Manual: `chunk init` in a test repo shows completion prompt after settings step
- [x] Manual: `chunk init --skip-completions` skips the prompt
- [x] Manual: `chunk completion install` still works independently

### DIFF
--- a/internal/cmd/completion.go
+++ b/internal/cmd/completion.go
@@ -95,11 +95,10 @@ func installCompletion(streams iostream.Streams) (err error) {
 	line := completionTag + "\n" + sh.source + "\n"
 
 	// Check if already installed.
-	if data, err := os.ReadFile(sh.rcFile); err == nil {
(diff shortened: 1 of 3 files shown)


danmux (approved):
Reviewed — approving.

PR #228, opened 2026-04-14
Redesign auth commands for multi-provider support

schurchleycci:
## Summary

https://linear.app/circleci/issue/FACT-73/simplify-tokenauth-management

- Redesign `auth` commands (`set`, `status`, `remove`) to accept a provider argument (`anthropic`, `circleci`) instead of the previous `login`/`logout` flow
- Add interactive credential input with hidden prompts, validation against live APIs, and confirmation before overwriting existing credentials
- Show all provider statuses in `auth status` (Anthropic, CircleCI, GitHub) with source indicators (env var vs config file) and masked key display
- `auth remove` warns when env vars will still take precedence after removing stored config
- Extend the user config schema: rename `apiKey` → `anthropicAPIKey` and add `circleCIToken`. `Load()` migrates the legacy `apiKey` field forward so existing users don't silently lose their stored key on upgrade; the legacy field is dropped from disk on the next `Save()`.
- `circleci.NewClient` now resolves its token via `config.Resolve` (env vars then config file) so a token stored by `chunk auth set circleci` is actually used. Adds `GetCurrentUser` for token validation during `auth set`.
- Replace `os.Exit(2)` in `config set` with proper cobra error handling (`ErrSilent`)
- Update help text, hints, and docs to reflect the new command signatures

chunk auth set circleci:

Screenshot 2026-04-14 at 4 09 38 PM

chunk auth status:

Screenshot 2026-04-14 at 4 07 49 PM

chunk auth set anthropic:

Screenshot 2026-04-14 at 4 08 11 PM

chunk auth status:

Screenshot 2026-04-14 at 4 08 26 PM

## Test plan

- [x] \`task build\` succeeds
- [x] \`task test\` — all tests pass (includes new coverage for legacy \`apiKey\` migration and \`circleCIToken\` resolution)
- [x] \`task lint\` — 0 issues
- [x] \`chunk --help\` shows \`chunk auth set <provider>\`
- [x] \`chunk config set badkey badvalue\` exits cleanly with code 1
- [x] Acceptance tests cover status with env vars, config file keys, masking, and env-overrides-config behavior
- [x] Existing config containing \`{"apiKey": "..."}\` still authenticates after upgrade and is rewritten as \`anthropicAPIKey\` on next save

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -27,6 +27,11 @@ type UserConfig struct {
 	AnthropicAPIKey string `json:"anthropicAPIKey,omitempty"`
 	CircleCIToken   string `json:"circleCIToken,omitempty"`
 	Model           string `json:"model,omitempty"`
+
+	// LegacyAPIKey reads the pre-rename "apiKey" field so existing users don't
+	// silently lose their stored Anthropic key on upgrade. Migrated into
+	// AnthropicAPIKey by Load and dropped on the next Save (omitempty).
+	LegacyAPIKey string `json:"apiKey,omitempty"`
 }
 
 // ResolvedConfig holds the final resolved values with their sources.
@@ -60,6 +65,10 @@ func Load() (UserConfig, error) {
 	if err := json.Unmarshal(data, &cfg); err != nil {
 		return UserConfig{}, err
 	}
+	if cfg.AnthropicAPIKey == "" && cfg.LegacyAPIKey != "" {
+		cfg.AnthropicAPIKey = cfg.LegacyAPIKey
+	}
+	cfg.LegacyAPIKey = ""
 	return cfg, nil
 }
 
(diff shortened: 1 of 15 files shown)


internal/config/config.go
```diff
@@ -16,22 +17,30 @@ const (
...
+	AnthropicAPIKey string `json:"anthropicAPIKey,omitempty"`
+	CircleCIToken   string `json:"circleCIToken,omitempty"`
+	Model           string `json:"model,omitempty"`
 }
 
 // ResolvedConfig holds the final resolved values with their sources.
 type ResolvedConfig struct {
-	APIKey       string
-	APIKeySource string
-	Model        string
```
danmux, line 31: The JSON key changes from `"apiKey"` to `"anthropicAPIKey"` — any existing user who stored a key via `chunk auth login` or `config set apiKey` will silently lose it after updating. Should there be migration logic in `Load()` that reads the old field and copies it forward?

danmux (approved):
Reviewed — approving.

PR #231, opened 2026-04-15
Add script to sync E2B templates from cimg Docker images

schurchleycci:
## Summary

- Adds `scripts/sync-e2b-templates.sh` which fetches the 3 most recent full-semver tags for `cimg/go`, `cimg/python`, `cimg/node`, and `cimg/rust` from Docker Hub and creates E2B sandbox templates from each
- Supports `--dry-run` flag to preview what would be built without actually creating templates
- Checks for required tools (`e2b`, `jq`, `curl`) at startup and reports a summary of created/failed templates at the end

## Test plan

- [ ] Run `./scripts/sync-e2b-templates.sh --dry-run` to verify Docker Hub API calls return valid semver tags
- [ ] Run `./scripts/sync-e2b-templates.sh` with E2B credentials to confirm templates are created
- [ ] Verify templates appear in `e2b template list`

### DIFF
--- a/scripts/sync-e2b-templates.sh
+++ b/scripts/sync-e2b-templates.sh
@@ -47,6 +47,8 @@ sanitize_version() {
 
 created=0
 failed=0
+cleanup_dirs=()
+trap 'rm -rf "${cleanup_dirs[@]+"${cleanup_dirs[@]}"}"' EXIT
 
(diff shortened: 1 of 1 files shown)


danmux (approved):
Reviewed — approving.

PR #235, opened 2026-04-17
Rename .chunk/sandbox to .chunk/sandbox.json

danmux:
## Summary
- Rename sandbox state file from `.chunk/sandbox` to `.chunk/sandbox.json` (session-keyed: `sandbox.<id>.json`)
- `chunk init` now appends sandbox tracking patterns to `.gitignore` if not present
- Fixes session-keyed sandbox files not being gitignored

### DIFF
--- a/internal/sandbox/active.go
+++ b/internal/sandbox/active.go
@@ -18,12 +18,12 @@ type ActiveSandbox struct {
 // Claude sessions in the same repo each maintain their own active sandbox.
 func sandboxFileName() string {
 	if id := os.Getenv("CLAUDE_SESSION_ID"); id != "" {
-		return "sandbox." + id
+		return "sandbox." + id + ".json"
 	}
-	return "sandbox"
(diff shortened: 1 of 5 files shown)


internal/sandbox/active.go
```diff
@@ -18,12 +18,12 @@ type ActiveSandbox struct {
 // Claude sessions in the same repo each maintain their own active sandbox.
 func sandboxFileName() string {
 	if id := os.Getenv("CLAUDE_SESSION_ID"); id != "" {
-		return "sandbox." + id
+		return "sandbox." + id + ".json"
```
michael-webster, line 21: not critical for this PR but maybe per session id directories would be a better choice for this.

michael-webster (approved):

internal/sandbox/active.go
danmux: will follow up

PR #240, opened 2026-04-17
Defer auth prompts with testable prompter and actionable CI errors

schurchleycci:
https://linear.app/circleci/issue/FACT-99/use-progressive-discovery-to-request-auth-tokens-based-on-whats-needed

## Summary

- Introduces `authprompt.Prompter` (a `func(label string) (string, error)` type) injected into all three `Ensure*Client` functions, making the TTY dependency explicit and the prompt-and-save path fully unit-testable via stub prompters
- Wraps `tui.ErrNoTTY` with `usererr.New` in each provider — CI users see actionable messages like "GitHub token required: set GITHUB_TOKEN or run 'chunk auth set github'" instead of "no interactive terminal available"
- Removes the `sk-ant-` prefix check from both the inline prompt and `auth set anthropic` so the env-var and interactive paths accept the same keys
- Adds unit tests for the full prompt-and-save path, no-TTY errors, and invalid token errors

## Test plan

- [ ] `go test -race ./...` passes
- [ ] `go test -race ./acceptance/...` passes, including the tightened no-TTY assertions
- [ ] Run `chunk build-prompt` in a shell with no `GITHUB_TOKEN` set — confirm error reads "GitHub token required: set GITHUB_TOKEN or run 'chunk auth set github'"
- [ ] Run `chunk auth set anthropic` with a non-`sk-ant-` key — confirm it is accepted (validated by API round-trip)


danmux (commented):
File-level comments inline below.

internal/authprompt/authprompt.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// EnsureCircleCIClient returns a ready-to-use CircleCI client. If a token is
+// already available (env var or config file), it uses it directly. Otherwise
+// it prompts inline once, validates, saves to config, and returns the client.
+func EnsureCircleCIClient(ctx context.Context, streams iostream.Streams, prompter Prompter) (*circleci.Client, error) {
+	rc, err := config.Resolve("", "")
+	if rc.CircleCIToken != "" {
+		return circleci.NewClient()
+	}
```
danmux, line 92: The three `Ensure*Client` functions are structurally identical — prompt, trim, validate, load, save, return. Would a single generic `ensureToken` helper with a provider-specific config reduce the duplication?

internal/anthropic/client.go
```diff
@@ -22,11 +22,11 @@ type Client struct {
 // ANTHROPIC_BASE_URL from the environment.
 func New() (*Client, error) {
 	rc, err := config.Resolve("", "")
-	if err != nil {
-		return nil, fmt.Errorf("resolve config: %w", err)
-	}
 	key := rc.AnthropicAPIKey
```
danmux, line 25: Is it intentional that `config.Resolve` errors are now swallowed when a key IS present? If `Resolve` fails (e.g. corrupt config file) but the key was found via env var, the error is silently dropped. The old code would surface it.

internal/cmd/auth.go
```diff
@@ -154,13 +160,8 @@ func authSetAnthropic(ctx context.Context, io iostream.Streams, anthropicBaseURL
 		return ErrSilent
```
danmux, line 160: The `sk-ant-` prefix check was removed here — was that deliberate? The old code rejected keys that didn't start with `sk-ant-` before hitting the API. Without it, users with e.g. an OpenAI key pasted by mistake will get a confusing API error instead of a clear format hint.

internal/cmd/init.go
```diff
@@ -159,16 +121,16 @@ func writeSettingsExample(dir string, data []byte, streams iostream.Streams) err
 }
```
danmux, line 121: The `--skip-circleci` flag is removed but the `--skip-completions` flag was added (seen in CLI.md). Was `--skip-circleci` in any documented workflows or scripts that would break?

internal/authprompt/authprompt.go
```diff
@@ -0,0 +1,244 @@
...
+	return nil
+}
+
+// EnsureCircleCIClient returns a ready-to-use CircleCI client. If a token is
+// already available (env var or config file), it uses it directly. Otherwise
+// it prompts inline once, validates, saves to config, and returns the client.
+func EnsureCircleCIClient(ctx context.Context, streams iostream.Streams, prompter Prompter) (*circleci.Client, error) {
+	rc, err := config.Resolve("", "")
+	if rc.CircleCIToken != "" {
+		return circleci.NewClient()
```
danmux, line 91: Could this hit a TOCTOU issue? `config.Resolve` is called at line 89, then `circleci.NewClient()` here calls it again independently. A concurrent config write between the two could cause surprising behavior. Probably fine for a CLI but worth noting.

internal/cmd/sandbox.go
```diff
@@ -126,7 +127,7 @@ environment variable (e.g. CHUNK_SANDBOX_PROVIDER=unikraft).`,
 			if err != nil {
 				return err
 			}
-			client, err := circleci.NewClient()
+			client, err := authprompt.EnsureCircleCIClient(cmd.Context(), io, tui.PromptHidden)
```
danmux, line 130: The `resolveOrgID` call happens before `EnsureCircleCIClient`. If the user has no org ID set, they get the org-ID error before being prompted for a CircleCI token. Should the token prompt come first so the error message makes more sense in context?

danmux (approved):

internal/authprompt/authprompt.go
danmux: The three differ in config fields, client return types, validation functions, and URLs — a generic helper needs enough parameters that the indirection isn't worth it. Keeping them separate.

internal/anthropic/client.go
danmux: Intentional. config.Resolve populates rc from env vars before touching the config file, and returns both the partial rc and any file error. If the env var provides the key, proceeding despite a corrupt config file is the right call — the env var is the higher-priority source.

internal/cmd/auth.go
danmux: Good catch — restored the sk-ant- prefix check in authSetAnthropic and added it to the inline prompt path in EnsureAnthropicClient too.

internal/cmd/init.go
danmux: No — it was only used interactively via chunk init. Removed deliberately along with the org picker, which was the whole point of deferring auth.

internal/authprompt/authprompt.go
danmux: Agreed, fine for a CLI — no concurrent config writers in practice.

internal/cmd/sandbox.go
danmux: Current order is deliberate — if org-id is missing the command can't proceed regardless, so fail fast before prompting for auth rather than collecting a token and then erroring.

PR #229, opened 2026-04-14
Add sandbox snapshot create and get commands

hanabel1:
- **`chunk sandbox snapshot create`**: Creates a snapshot of a sandbox. Accepts   
  `--sandbox-id` (falls back to active sandbox) and `--name` (required). Calls `POST
   /api/v2/sandbox/snapshots` on the server with `sandbox_id` and `name` in the     
  request body.                                                                   
  - **`chunk sandbox snapshot get`**: Fetches a snapshot by ID. Calls `GET
  /api/v2/sandbox/snapshots/:id`.
  - **Fake server handlers**: Added `POST /api/v2/sandbox/snapshots` and `GET
  /api/v2/sandbox/snapshots/:id` to `FakeCircleCI` for use in acceptance tests.     
  - **Acceptance tests**: 9 tests covering happy paths, request body shape, active
  sandbox fallback, missing flags, auth errors, and API errors.                     
                                                                                  
  Note: list and restore are not included here — those require new endpoints on the 
  sandbox-provisioner side first.

  ## Test plan                                                                    

  - [ ] `go test ./...` passes
  - [ ] `chunk sandbox snapshot create --sandbox-id <id> --name <name>` creates a
  snapshot and prints the ID                                                        
  - [ ] `chunk sandbox snapshot create --name <name>` uses the active sandbox when
  `--sandbox-id` is omitted                                                         
  - [ ] `chunk sandbox snapshot get <id>` prints the snapshot name and ID

PR #241, opened 2026-04-17
Re-add sandbox snapshot commands + fix flaky WebSocket test

michael-webster:
## Summary

- Re-applies the sandbox snapshot create/get commands from #229 (which was reverted in #238 due to a flaky test failure unrelated to the feature)
- Fixes the flaky `TestExecOverSSHViaWebSocket` test: both sides of a WebSocket/SSH connection may initiate close simultaneously during teardown; `net.ErrClosed` on `Close()` now treated as a successful close rather than surfaced as an error

## Test plan

- [x] `task test` passes (including `TestExecOverSSHViaWebSocket` with `-count=10 -race`)
- [x] `task acceptance-test` passes for snapshot commands
- [x] CI green


internal/sandbox/ssh.go
```diff
@@ -58,6 +58,11 @@ func (c *sshConn) Close() error {
 	if c.cleanup != nil {
 		c.cleanup()
 	}
+	// Both sides may initiate close simultaneously; if the remote end already
+	// closed the connection, treat it as a successful close.
+	if errors.Is(err, net.ErrClosed) {
```
danmux, line 63: Should `errors.Is(err, net.ErrClosed)` also check the inner `sshClient.Close()` error separately? If `sshClient.Close()` returns a real error but `wsConn.Close()` returns `net.ErrClosed`, the real error gets masked.

internal/sandbox/snapshot.go
```diff
@@ -0,0 +1,15 @@
+package sandbox
+
+import (
+	"context"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/circleci"
+)
+
+func CreateSnapshot(ctx context.Context, client *circleci.Client, sandboxID, name string) (*circleci.Snapshot, error) {
```
danmux: These two functions are pure pass-throughs to `client.CreateSnapshot` / `client.GetSnapshot` with no additional logic. What value does this package-level indirection add over calling the client directly from the command?

internal/cmd/sandbox.go
```diff
@@ -475,3 +476,71 @@ Example:
...
+			client, err := circleci.NewClient()
+			if err != nil {
+				return err
+			}
+			snap, err := sandbox.GetSnapshot(cmd.Context(), client, args[0])
+			if err != nil {
+				return err
+			}
+			if snap.Name != "" {
+				io.Printf("%s  %s\n", snap.Name, snap.ID)
```
danmux, line 537: The output format (`Name  ID`) uses two spaces as separator. If this is meant to be machine-parseable or aligned with `sandbox list`, would a consistent format (e.g. table writer or tab) be better?

internal/testing/fakes/circleci.go
```diff
@@ -229,6 +243,61 @@ func (f *FakeCircleCI) handleExec(c *gin.Context) {
...
+		SandboxID string `json:"sandbox_id"`
+		Name      string `json:"name"`
+	}
+	if err := json.NewDecoder(c.Request.Body).Decode(&body); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"message": "Bad request"})
+		return
+	}
+
+	snap := Snapshot{
+		ID:   "snap-new-123",
```
danmux: The fake always returns `snap-new-123` as the ID. `TestSandboxSnapshotCreateUsesActiveSandbox` relies on `sandbox-new-123` from sandbox create. Could these collide if a test asserts on both IDs in the same recorder?

internal/sandbox/ssh.go
michael-webster: Good catch to double-check. `c.Client.Close()` returns a single error — it closes the SSH connection and its underlying transport (the WebSocket) in one call. There's no separate `wsConn.Close()` path that could produce a different error. So the only error we get back is from the transport close, and `errors.Is(err, net.ErrClosed)` precisely identifies 'connection already gone' without masking anything else. If a real SSH-level error occurred during close, it would not wrap `net.ErrClosed` and would be propagated normally.

internal/sandbox/snapshot.go
michael-webster: `List`, `Create`, and `Exec` in `sandbox.go` are identical pass-throughs today (lines 13–23). The `sandbox` package acts as a consistent boundary between `cmd/` and `circleci.Client` — some functions already carry real logic (`AddSSHKey`, `SSH`), and the pass-throughs are there for consistency and to give a place for logic to grow. Snapshot follows the same convention. Happy to change it if you'd prefer cmd call the client directly, but that would make snapshot inconsistent with the rest of the package.

internal/cmd/sandbox.go
michael-webster: `sandbox list` uses the exact same format — see line 98 of this file: `io.Printf("%s  %s\n", s.Name, s.ID)`. Snapshot get matches it deliberately. If we decide to move to a table writer, that should be a single change across all sandbox output, not piecemeal.

danmux (approved):

PR #242, opened 2026-04-18
Add an org selector for sandbox creation

z00b:
Don't make me go to the web to find an ID

### DIFF
--- a/acceptance/sandbox_gaps_test.go
+++ b/acceptance/sandbox_gaps_test.go
@@ -276,6 +276,89 @@ func TestSandboxCreateAPIError403(t *testing.T) {
 	assert.Assert(t, result.ExitCode != 0, "expected non-zero exit for 403 response")
 }
 
+// --- create org picker paths ---
+
+func TestSandboxCreateCollaborationsAPIError(t *testing.T) {
+	cci := fakes.NewFakeCircleCI()
+	cci.CollaborationsStatusCode = 500
+	srv := httptest.NewServer(cci)
+	defer srv.Close()
+
+	env := testenv.NewTestEnv(t)
+	env.CircleCIURL = srv.URL
+
+	result := binary.RunCLI(t, []string{
+		"sandbox", "create",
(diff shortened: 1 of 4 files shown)


internal/cmd/sandbox.go
```diff
@@ -122,9 +123,28 @@ The sandbox backend defaults to e2b. Override with the CHUNK_SANDBOX_PROVIDER
 environment variable (e.g. CHUNK_SANDBOX_PROVIDER=unikraft).`,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			io := iostream.FromCmd(cmd)
-			resolvedOrgID, err := resolveOrgID(orgID)
-			if err != nil {
-				return err
+			resolvedOrgID, orgErr := resolveOrgID(orgID)
+			if orgErr != nil {
+				client, err := circleci.NewClient()
+				if err != nil {
```
danmux: A new `circleci.NewClient()` is created here for the collaborations call, then another one at line 150 for the actual sandbox create. Could the first client be reused?

acceptance/sandbox_gaps_test.go
```diff
@@ -276,6 +276,89 @@ func TestSandboxCreateAPIError403(t *testing.T) {
...
+	assert.Equal(t, len(collabReqs), 1, "expected collaborations endpoint to be called")
+
+	// The error must not be the generic org-id-missing error — the code
+	// must have reached and attempted the TUI picker.
+	combined := result.Stdout + result.Stderr
+	assert.Assert(t, !strings.Contains(combined, "--org-id"),
+		"expected TUI error, not org-id error, got: %s", combined)
+}
+
 // --- list error paths ---
```
danmux, line 357: This asserts the error does NOT contain `--org-id`. If the TUI error message ever changes to mention `--org-id` as a fallback hint, this test breaks for the wrong reason. Would asserting on a positive string (e.g. the actual TUI error text) be more robust?

danmux (approved):

PR #243, opened 2026-04-20
Update python dependencies

z00b:
Clean up some vulns.


lavin-ds (approved):

acceptance/repos.json
```diff
@@ -24,7 +24,7 @@
       "test": "cargo test --workspace --exclude serde --exclude serde_core --exclude serde_test_suite",
       "system_deps": [],
       "image": "cimg/rust",
-      "image_version": "1.94.1"
+      "image_version": "1.95.0"
```
danmux, line 27: Rust bumped from 1.94.1 to 1.95.0 — is this tested against the acceptance suite or just a version pin update? If the serde tests exercise compiler-specific behavior, a minor Rust bump could surface surprises.

harness/uv.lock
```diff
@@ -127,19 +127,19 @@ requires-dist = [
 
 [[package]]
 name = "claude-agent-sdk"
-version = "0.1.56"
+version = "0.1.63"
```
danmux, line 130: claude-agent-sdk jumped from 0.1.56 to 0.1.63 — seven patch versions. Any breaking changes in the SDK changelog worth calling out? Pre-1.0 semver doesn't guarantee patch compatibility.

acceptance/repos.json
z00b: these only get updated when the harness process successfully completes after detecting a change. it was a result of testing that the other dep update didn't break anything.

harness/uv.lock
z00b: i'll take a look, again, the harness ran successfully so i think it's safe enough to merge.

PR #253, opened 2026-04-21
Show config file path in auth token prompts

danmux:
## Summary
- Show where tokens will be saved before prompting (path + mode 0600)
- Show where tokens were saved in the success message
- Extract `printSaveHint`/`printSaved` helpers to keep the format in one place

### DIFF
--- a/internal/authprompt/authprompt.go
+++ b/internal/authprompt/authprompt.go
@@ -82,6 +82,20 @@ func ValidateAPIKey(ctx context.Context, apiKey, baseURL string) error {
 	return nil
 }
 
+func PrintSaveHint(streams iostream.Streams, label string) {
+	if cfgPath, err := config.Path(); err == nil {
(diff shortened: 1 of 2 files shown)


internal/authprompt/authprompt.go
```diff
@@ -82,6 +82,20 @@ func ValidateAPIKey(ctx context.Context, apiKey, baseURL string) error {
...
+	if cfgPath, err := config.Path(); err == nil {
+		streams.ErrPrintln(ui.Dim(fmt.Sprintf("%s will be saved to user config (%s, mode 0600)", label, cfgPath)))
+	}
+}
+
+func PrintSaved(streams iostream.Streams, label string) {
+	msg := label + " saved"
+	if cfgPath, err := config.Path(); err == nil {
+		msg = fmt.Sprintf("%s saved to user config (%s)", label, cfgPath)
+	}
```
michael-webster, line 95: worth reporting an error here if this fails or do the other PRs handle that?

michael-webster (approved):
approved mod one question

PR #269, opened 2026-04-22
feat: Stop hook redesign with --if-changed, attempt tracking, and worktree support

michael-webster:
## Summary

- Adds a `Stop` hook to `.claude/settings.json` that runs `chunk validate --if-changed` after every session, skipping when the working tree is clean
- `--if-changed` mode detects untracked files (via `git status --porcelain`), acquires an advisory lock to prevent concurrent runs, tracks consecutive failures per content hash, and exits 2 to re-signal the agent
- Attempt tracking stops re-signaling after `stopHookMaxAttempts` consecutive failures on the same content, with a `.chunk/force-validate` sentinel to bypass the limit
- Settings merge logic extended to handle `Stop` hook groups alongside existing `PreToolUse` support
- Removes validate result caching (replaced by attempt-based logic)

## Code review fixes (this PR)

- `HasUncommittedChanges` switched to `git status --porcelain` to catch untracked files
- `TrackFailedAttempt` now accepts a pre-computed hash to avoid spurious resets from concurrent commits between `RunAll` and the attempt counter update
- `ComputeContentHash` exported so the cmd layer computes it once before `RunAll`
- `TryLock` gains an `io.Writer` param so filesystem errors surface as diagnostics rather than silent fail-open
- `runCommand` stdout restored to `streams.Out`; Stop hook path creates a `hookStreams` with `Out: streams.Err` to route all output into the agent feedback block
- `stopTimeout` capped at 600s in `settings.Build`
- `--if-changed` flag help documents `CLAUDE_WORKING_DIR` worktree behaviour

## Test plan

- [ ] `task test` passes (804 tests)
- [ ] `chunk validate --if-changed` skips on clean tree, runs and exits 2 on failure
- [ ] Stop hook in `.claude/settings.json` triggers at session end


michael-webster:
Superseded by #271 (remove validate caching) and #272 (Stop hook implementation), which split this PR into two focused changes.

PR #272, opened 2026-04-22
feat: add stop hook with unified execution and git-change detection

michael-webster:
## Summary

Adds a Claude Code Stop hook that runs configured validation commands at session end. Hook and direct `chunk validate` invocations share the same execution path — flags like `--remote`, `--sandbox-id`, `--cmd`, and named commands work in both contexts.

### How it works

- **Hook detection**: reads the Claude Code JSON payload from stdin when stdin is not a terminal; a non-empty `session_id` confirms hook context. No environment variable needed.
- **Git-change detection**: skips entirely when `git status --porcelain` is clean — no unnecessary runs at session end
- **Attempt tracking**: after N consecutive failures the hook exits 0 and tells the agent to ask the user for help. Configurable via `stopHookMaxAttempts` in `.chunk/config.json` (default 3)
- **Exit code 2**: signals Claude Code to re-invoke the agent on failure
- **stdout → stderr routing**: all command output goes to stderr so it appears in the feedback block

### Design

The hook lifecycle is a thin wrapper around the shared `runValidate` dispatcher rather than a separate execution path:

```
detect hook → redirect streams → git check → runValidate (same as direct) → WrapHookResult
```

Key validate package exports: `HasGitChanges`, `WrapHookResult`, `HookExitError`.
`RunHook` removed — replaced by `WrapHookResult` + shared dispatcher.
`CHUNK_HOOK_ENABLE` env var removed — stdin detection is sufficient.
Duplicated SSH session setup consolidated into `openSSHSession`.

## Test plan

- [x] `task test` passes
- [x] Hook skips when working tree is clean
- [x] Hook runs and exits 2 on failure (dirty tree)
- [x] After max attempts, stops re-signaling and prints guidance
- [x] Acceptance tests for hook mode (dirty + clean tree via stdin payload)
- [x] Verified end-to-end in this repo with deliberate failing test


internal/cmd/usererr.go
```diff
@@ -23,17 +23,19 @@ func notAuthorized(action string, err error) error {
 }
 
 func sshSessionError(err error) error {
-	if e, ok := errors.AsType[*sandbox.KeyNotFoundError](err); ok {
+	var knf *sandbox.KeyNotFoundError
```
danmux: clod base regression!

(caught me out before)

internal/task/collect.go
```diff
@@ -23,6 +23,7 @@ type ProjectDetailFunc func(ctx context.Context, slug string) (*circleci.Project
 // CollectRunConfig drives the interactive form to build a RunConfig.
 // It takes already-fetched projects and collaborations plus injected UI
 // and data-fetch dependencies so the logic is testable without a TTY.
+// envOrgID is the value of CIRCLECI_ORG_ID from the environment (pass "" to skip the mismatch check).
```
danmux: possible regression too ? best to not leak env var knowledge into packages

internal/validate/attempts.go
```diff
@@ -0,0 +1,77 @@
...
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// ForceHookFilePath is the sentinel file that, when present, bypasses the
+// max-attempts guard so the Stop hook always re-signals the agent.
+// Create with: touch .chunk/force-validate
+// Remove when done debugging.
+const ForceHookFilePath = ".chunk/force-validate"
```
danmux: feels like this should be passed in?

internal/task/collect.go
michael-webster: ahh i've been used to these being created in draft

internal/validate/attempts.go
michael-webster: yeah this was wip forgot to me specific enough about creating draft PRs

.claude/settings.json
```diff
@@ -6,9 +6,6 @@
       "Bash(chunk:*)"
     ]
   },
-  "env": {
-    "CHUNK_HOOK_ENABLE": "1"
-  },
   "hooks": {
     "PreToolUse": [
```
schurchleycci, line 10: Should these be using `chunk validate` too?

schurchleycci: If so, we may need to do `chunk validate || exit 2` because the exit code 1 that chunk validate returns on a failure doesn't block the commit (only exit code 2 blocks)

docs/HOOKS.md
schurchleycci, line 64: Same question here

docs/HOOKS.md
```diff
@@ -53,6 +70,13 @@ Generated by `chunk init`. Runs each command directly before commits:
           {"type": "command", "command": "cd ${CLAUDE_PROJECT_DIR:-.} && task test", "timeout": 300}
         ]
       }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {"type": "command", "command": "chunk validate --if-changed", "timeout": 420}
```
schurchleycci: I noticed in testing that if you introduce a breaking change `chunk validate` runs (as expected) and then revert that change `chunk validate` doesn't rerun since the `git status` is clean. I assume this is expected and is a bit of an artificial scenario that in practice wouldn't happen much, but just wanted to raise it in case.

schurchleycci (approved):

.claude/settings.json
michael-webster: possibly, i think these were configured to run much more targeted checks

michael-webster: i updated it to use change for lint and test, but kept format separate.

docs/HOOKS.md
michael-webster: yeah i think it's a possible scenario, to account for. i'd expect in most cases though that would be part of a larger model pass where other edits are occurring.

docs/HOOKS.md
michael-webster: i updated it with the lint and test examples, but let format be run separately. we aren't trying to replace every instance you'd use claude hooks and formatting is a little different.

PR #277, opened 2026-04-22
[FACT-110] Sandbox set up in chunk cli

hanabel1:
## Summary
  - Adds `chunk sandbox setup` — detects the tech stack, syncs files to a sandbox, runs the install command, and creates a snapshot
  - Runs the install in a login shell (`bash -l`) so PATH entries from `~/.profile` (e.g. `/usr/local/go/bin`) are available
  - `cd`s to the synced workspace directory before install so package managers can find their manifest files (`go.mod`, `package.json`, etc.)

  ## Test plan
  - [ ] `chunk sandbox setup --dir .` on a Go project with Go pre-installed — install runs and snapshot is created
  - [ ] `--skip-sync` skips the sync step
  - [ ] `--skip-snapshot` skips snapshot creation
  - [ ] No active sandbox + `--name` + `--org-id` creates a new sandbox first

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -714,145 +714,35 @@ Example:
 			// Step 2: Resolve or create sidecar.
 			var scDisplayName string
 			if sidecarID == "" {
-				active, err := sidecar.LoadActive()
-				if err != nil {
-					return &userError{msg: "Could not load the active sidecar.", suggestion: configFilePermHint, err: err}
-				}
-				if active != nil {
-					sidecarID = active.SidecarID
-					scDisplayName = active.Name
-					status(iostream.LevelInfo, fmt.Sprintf("using active sidecar %s", sidecarID))
-				} else {
-					if name == "" {
-						return &userError{
-							msg:        "No active sidecar and --name not provided.",
-							suggestion: "Pass --name to create a new sidecar, or run 'chunk sidecar use <id>'.",
-							errMsg:     "no active sidecar and --name not provided",
-						}
-					}
-					resolvedOrgID, err := resolveOrgID(orgID, orgPicker(cmd.Context(), client))
-					if err != nil {
-						return err
-					}
-					provider := os.Getenv(config.EnvSidecarProvider)
-					if provider == "" {
-						provider = defaultProvider
-					}
-					status(iostream.LevelStep, fmt.Sprintf("Creating sidecar %q...", name))
-					sc, err := sidecar.Create(cmd.Context(), client, resolvedOrgID, name, provider, "")
-					if err != nil {
-						if err := notAuthorized("create sidecars", err); err != nil {
-							return err
-						}
-						return &userError{
-							msg:        "Could not create the sidecar.",
-							suggestion: "Check your network connection and try again.",
-							err:        err,
-						}
-					}
-					sidecarID = sc.ID
-					scDisplayName = sc.Name
-					if err := sidecar.SaveActive(sidecar.ActiveSidecar{SidecarID: sc.ID, Name: sc.Name}); err != nil {
-						streams.ErrPrintf("warning: could not save active sidecar: %v\n", err)
-					}
-					status(iostream.LevelDone, fmt.Sprintf("Created sidecar %s (%s)", sc.Name, sc.ID))
+				var resolveErr error
+				sidecarID, scDisplayName, resolveErr = sidecarSetupResolveSidecar(cmd.Context(), client, orgID, name, status, streams)
+				if resolveErr != nil {
+					return resolveErr
 				}
 			}
 
 			// Step 3: Ensure SSH key exists (generate if missing and no explicit key given).
-			if identityFile == "" {
-				keyPath := sidecar.DefaultKeyPath()
-				if _, err := os.Stat(keyPath); os.IsNotExist(err) {
-					status(iostream.LevelStep, fmt.Sprintf("Generating SSH key at %s...", keyPath))
-					if err := sidecar.GenerateKeyPair(keyPath); err != nil {
-						return &userError{msg: "Could not generate SSH key.", err: err}
-					}
-					status(iostream.LevelDone, "SSH key generated")
-				}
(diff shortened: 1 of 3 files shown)


danmux (commented):
-

internal/cmd/sandbox.go
```diff
@@ -668,3 +669,258 @@ func newSandboxSnapshotGetCmd() *cobra.Command {
...
+	ctx context.Context,
+	client *circleci.Client,
+	sandboxID, sbDisplayName, snapshotName string,
+	status iostream.StatusFunc,
+) error {
+	if snapshotName == "" {
+		if sbDisplayName != "" {
+			snapshotName = sbDisplayName + "-setup"
+		} else {
+			snapshotName = sandboxID[:8] + "-setup"
```
danmux: Could `sandboxID` ever be shorter than 8 chars? If `--sandbox-id` is passed directly (skipping `sandboxSetupResolveSandbox`), `sbDisplayName` stays empty and this slice is the fallback. A short or malformed ID would panic with index-out-of-range. A guard like `min(len(sandboxID), 8)` would prevent this. This is reachable from user input (`--sandbox-id` flag), so it should be defended.

internal/cmd/sandbox.go
```diff
@@ -668,3 +669,258 @@ func newSandboxSnapshotGetCmd() *cobra.Command {
...
+	if err != nil {
+		if sessErr := sshSessionError(err); sessErr != nil {
+			return sessErr
+		}
+		return err
+	}
+	// Run from the synced workspace directory so package managers can
+	// find their manifest files (go.mod, package.json, etc.).
+	workspaceCmd := env.Install
+	if active, err := sandbox.LoadActive(); err == nil && active != nil && active.Workspace != "" {
```
danmux: This calls `LoadActive()` a second time — the first call is in `sandboxSetupResolveSandbox` (step 2). Two problems: (1) if `SaveActive` failed at line 792 (only a warning), this returns `nil` and silently skips the `cd` into the workspace directory, so the install runs in the wrong location; (2) errors from `LoadActive` here are silently swallowed by the `err == nil` guard. The workspace path was already available when the active sandbox was resolved — thread it through as a parameter instead of re-reading from disk.

internal/cmd/sandbox.go
```diff
@@ -668,3 +669,258 @@ func newSandboxSnapshotGetCmd() *cobra.Command {
...
+					suggestion: "Check the directory contains a supported project.",
+					err:        err,
+				}
+			}
+			status(iostream.LevelInfo, fmt.Sprintf("stack: %s", env.Stack))
+			status(iostream.LevelInfo, fmt.Sprintf("install: %s", env.Install))
+
+			// Step 2: Resolve or create sandbox.
+			var sbDisplayName string
+			if sandboxID == "" {
```
danmux: When `--sandbox-id` is passed explicitly this block is skipped and `sbDisplayName` stays empty. That means `sandboxSetupSnapshot` always falls through to the `sandboxID[:8]` branch — which itself can panic if the ID is short (see comment on line 912). Should the display name be resolved from the API when an explicit ID is given, or at minimum should the snapshot default name handle the missing display name more safely?

internal/sandbox/session.go
```diff
@@ -22,6 +26,43 @@ const (
 	knownHostsFile = "chunk_ai_known_hosts"
 )
 
+// DefaultKeyPath returns the default SSH private key path used by chunk.
+func DefaultKeyPath() string {
+	return filepath.Join(os.Getenv("HOME"), ".ssh", defaultKeyName)
```
danmux: `os.Getenv("HOME")` is empty on some platforms and less idiomatic than `os.UserHomeDir()`. If `HOME` is unset the returned path would be `/.ssh/chunk_ai`, which fails in a confusing way. Additionally, `OpenSession` (line 81 in this same file) resolves the `.ssh` dir via `os.Getenv(config.EnvHome)` — so there is an inconsistency even within the same file. Should this use `config.EnvHome` or `os.UserHomeDir()` for consistency?

internal/cmd/sidecar.go
```diff
@@ -668,3 +669,258 @@ func newSandboxSnapshotGetCmd() *cobra.Command {
...
+			if err := sandboxSetupEnsureSSHKey(identityFile, status); err != nil {
+				return err
+			}
+
+			// Step 4: Sync files to sandbox.
+			if !skipSync {
+				if err := sandboxSetupSync(cmd.Context(), client, sandboxID, identityFile, authSock, status); err != nil {
+					return err
+				}
+			}
```
danmux, line 738: The `RunE` body runs a 6-step orchestration workflow inline (~35 lines) and the helper functions (`sandboxSetupResolveSandbox`, `sandboxSetupRunInstall`, etc.) are all defined in the `cmd` file. This is effectively business logic that can't be tested without importing `cmd/`. Would it make sense to move the orchestration and helpers into `internal/sandbox/` (or a new `internal/sandboxsetup/` package) so `RunE` just parses flags and calls a single setup function?

internal/cmd/sandbox.go
```diff
@@ -668,3 +669,258 @@ func newSandboxSnapshotGetCmd() *cobra.Command {
...
+			msg:        "No active sandbox and --name not provided.",
+			suggestion: "Pass --name to create a new sandbox, or run 'chunk sandbox use <id>'.",
+			errMsg:     "no active sandbox and --name not provided",
+		}
+	}
+	resolvedOrgID, err := resolveOrgID(orgID, orgPicker(ctx, client))
+	if err != nil {
+		return "", "", err
+	}
+	provider := os.Getenv(config.EnvSandboxProvider)
```
danmux: `os.Getenv(config.EnvSandboxProvider)` is read inside this helper rather than hoisted to the `RunE` layer and passed as a parameter. The env var read is buried in what is effectively business logic. Pass `provider` as a parameter for consistency with the project's layering conventions.

danmux (approved):

PR #280, opened 2026-04-23
Retry sync and keep default error detail

danmux:
fixes git drift :

```
Error: An unknown error occurred.

sync failed: fatal: Could not parse object '3616bdf40d49059c0244d1181be0ef002328cb11'.
```


michael-webster (approved):
This works, we could also probably recalcuate the merge base but just rm and re-sync works.

danmux:
> This works, we could also probably recalcuate the merge base but just rm and re-sync works.

Yah i bet theres plenty more elegant ways to do sync :)

PR #288, opened 2026-04-27
Remote validation: route commands to sidecar with auto-creation

michael-webster:
## Summary

- `chunk validate` now routes commands marked `\"remote\": true` in `.chunk/config.json` to the active sidecar automatically — no `--remote` flag required
- When `--remote` is passed, all commands run on the sidecar (existing explicit override)
- `chunk validate --remote` auto-creates a sandbox if no active sidecar is set, then saves it for future runs
- Org ID is persisted in `.chunk/config.json` (`orgID` field) so sandbox creation skips the interactive picker on subsequent runs — set it with `chunk config set orgID <id>`
- In Stop hook context, if no sidecar is active and org is configured, one is created automatically; otherwise falls back to local with a warning
- Status messages show which commands are routed to the sidecar vs running locally

## New config fields

**Per-command** (`.chunk/config.json`):
```json
{ "name": "test", "run": "task test", "remote": true }
```

**Project-level** (`.chunk/config.json`):
```json
{ "orgID": "your-circleci-org-id" }
```
Or: `chunk config set orgID <id>`

## Test plan

- [ ] `chunk validate test` with active sidecar routes test to sidecar
- [ ] `chunk validate test` with no sidecar runs locally with warning
- [x] `chunk validate` (run all) splits remote/local commands correctly
- [x] `chunk validate --remote` creates sandbox when none exists
- [x] `chunk config set orgID <id>` persists to `.chunk/config.json`
- [x] Stop hook auto-creates sandbox when org ID is configured

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -85,6 +85,8 @@ func newConfigSetCmd() *cobra.Command {
 						projCfg.Validation = &config.ValidationConfig{}
 					}
 					projCfg.Validation.SidecarImage = value
+				default:
+					return fmt.Errorf("internal: unhandled project config key %q", key)
 				}
 				if err := config.SaveProjectConfig(workDir, projCfg); err != nil {
 					return &userError{msg: "Could not save project configuration.", suggestion: configFilePermHint, err: err}
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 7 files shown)


internal/cmd/validate.go
schurchleycci, line 1: Claude review:

> In the split remote/local path, runErr is only overwritten by the local error when runErr == nil. If remote commands fail and local commands also fail, only the remote error surfaces — the local failure is lost. A user trying to understand why validate failed wouldn’t see the full picture.
> 
> Consider using errors.Join:
> 
> if err := mapValidateError(validate.RunAll(ctx, workDir, localCfg, statusFn, streams)); err != nil {
>     runErr = errors.Join(runErr, err)
> }

internal/cmd/config.go
schurchleycci, line 1: Claude review:

> The switch key block has cases for "orgID" and "validation.sidecarImage" but no default branch. Any key in ValidProjectConfigKeys that doesn’t match a case will call SaveProjectConfig with an unmodified config and then print a success message — silently lying to the user. This is a silent data-loss trap for any future key added to the map without a matching switch case.
> 
> Suggested fix:
> 
> default:
>     return fmt.Errorf("internal: unhandled project config key %q", key)

schurchleycci (approved):

PR #292, opened 2026-04-28
Add --bundle flag to chunk sidecar sync

michael-webster:
## Summary

Two commits:

**1. Wire stdin through `chunk sidecar ssh` when piped**

Detects when the process stdin is a pipe and forwards it to the remote command. Enables direct bundle transfer via `cat bundle | chunk sidecar ssh`. Detection moved to the `cmd/` layer (the architectural boundary for I/O) rather than reaching for `os.Stdin` inside the business package.

**2. Add `--bundle` to `chunk sidecar sync` for GitHub-free incremental sync**

Introduces `BundleSync` which syncs the local repo using `git bundle` rather than the merge-base patch approach — the branch does not need to be pushed to GitHub first.

- First sync sends a full bundle of HEAD; initialises a bare repo on the sidecar if absent
- Subsequent syncs send only `<lastRef>..HEAD` (incremental), skipping the bundle entirely when HEAD has not advanced
- Working-tree changes are applied on top as a patch after each bundle fetch
- `LastSyncedRef` is stored in `.chunk/sidecar.json` to drive the incremental path
- `SaveActive` uses an atomic temp-file rename to prevent concurrent sync races
- `HeadRef` and `CreateBundle` accept a `cwd` so they always operate on the right repo
- `LoadActive` errors are propagated rather than silently falling back to a fresh state

Opt in via `--bundle` flag or set `bundleSync: true` in `.chunk/config.json`. The `sidecar setup` command reads the same config key.

## Test plan

- [ ] `chunk sidecar sync --bundle` on a fresh sidecar performs a full bundle sync
- [ ] Re-running `chunk sidecar sync --bundle` skips the bundle and only applies the working-tree patch
- [ ] `chunk validate test --remote` passes after a bundle sync
- [ ] `cat bundle | chunk sidecar ssh -- git fetch ...` works (stdin piping)
- [ ] Setting `bundleSync: true` in `.chunk/config.json` uses bundle sync without the flag


michael-webster:
closing this out, in favor of a broader redesign of the sync with the sidecar lifecycle.

PR #296, opened 2026-04-28
Track active snapshot after sidecar setup

michael-webster:
## Summary

- After `chunk sidecar setup` snapshots the sidecar, the snapshot ID is now persisted to `.chunk/snapshot.json` (session-keyed, same pattern as `sidecar.json`)
- Snapshot ID is also written to `validation.sidecarImage` in `.chunk/config.json` so `chunk validate --remote` auto-creates new sidecars from the snapshot instead of from scratch
- Active sidecar is cleared after snapshotting — the snapshotted sidecar is a template, not a general-purpose worker

## New flow

```
chunk sidecar setup        # create → install → snapshot → clear active sidecar
                           # writes .chunk/snapshot.json + config.json validation.sidecarImage
chunk sidecar create       # user creates a fresh worker sidecar (from snapshot via --image)
chunk validate --remote    # auto-creates worker from snapshot if no active sidecar
```

## Notes

- `ValidationConfig` struct is also added on `webster/remote-validation-commands` — trivial merge conflict when both land
- Phase 4 (annotating `sidecar list` with snapshot info) intentionally skipped — would require local tracking we don't want to rely on

## Test plan

- [x] `SaveActiveSnapshot` / `LoadActiveSnapshot` round-trip
- [x] `ClearActiveSnapshot` removes file
- [x] Session-keyed filename via `CLAUDE_SESSION_ID`
- [x] Git-root-bounded walk
- [ ] `chunk sidecar setup` end-to-end with real sidecar

### DIFF
--- a/internal/sidecar/snapshot.go
+++ b/internal/sidecar/snapshot.go
@@ -1,0 +1,97 @@
+package sidecar
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/config"
+)
+
+// ActiveSnapshot holds the most recently created snapshot for a project.
+type ActiveSnapshot struct {
+	ID   string `json:"id"`
+	Name string `json:"name,omitempty"`
+}
+
+func snapshotFileName() string {
+	if id := os.Getenv(config.EnvClaudeSession); id != "" {
+		return "snapshot." + id + ".json"
+	}
+	return "snapshot.json"
+}
+
+// LoadActiveSnapshot walks up from cwd looking for .chunk/snapshot.json. Returns nil if not found.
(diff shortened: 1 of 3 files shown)


internal/sidecar/snapshot.go
```diff
@@ -0,0 +1,97 @@
...
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/config"
+)
+
+// ActiveSnapshot holds the most recently created snapshot for a project.
+type ActiveSnapshot struct {
```
michael-webster, line 13: this completely mirrors the existing sidecar tracking code which will get us unblocked.

danmux (approved):

PR #298, opened 2026-04-29
Update chunk-sidecar skill to improve first-run and validation UX

schurchleycci:
## Summary

- **`chunk init` now offers to install agent skills** as a new step (prompted, default yes). Skips silently when all skills are already current. Adds `--skip-skills` flag to match the existing `--skip-hooks` / `--skip-completions` pattern.
- **`chunk-sidecar` skill updated to v1.1.0** with UX improvements based on real session experience:
  - Step 2 checks `validation.sidecarImage` in config first — creates from snapshot automatically if set, otherwise goes straight to full setup (Step 3). Removes the "list existing sidecars and ask" path.
  - Step 3 setup flow no longer gated on the user explicitly asking to prep — triggers automatically when no snapshot image is configured. After snapshotting, immediately creates a clean sidecar from the snapshot rather than continuing with the original.
  - Validation throughout uses `chunk validate` (per-command routing) instead of `chunk validate --remote`, which was incorrectly sending all commands to the sidecar including `test-changed` whose `{{CHANGED_PACKAGES}}` template only resolves locally.
  - Always passes `--org-id` to `chunk sidecar create` since interactive org selection doesn't work in Claude sessions.
  - New troubleshooting entries: SSH/API timeouts with recovery steps, and missing dependencies or binaries not on `$PATH`.

## Test plan

- [ ] Run `chunk init` in a fresh repo — confirm the "Install agent skills?" prompt appears and installs skills
- [ ] Run `chunk init` again — confirm the skills step is skipped silently (all current)
- [ ] Run `chunk init --skip-skills` — confirm no skill prompt
- [ ] Run `chunk skill install` — confirm it reports `updated chunk-sidecar` (version bump to 1.1.0)
- [ ] Start a new session with no active sidecar and `sidecarImage` set — confirm skill creates from snapshot + syncs without prompting for full setup
- [ ] Start a new session with no active sidecar and no `sidecarImage` — confirm skill proceeds to Step 3 automatically

### DIFF
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -201,10 +201,10 @@ func installSkillsStep(streams iostream.Streams) {
 			continue
 		}
 		for _, name := range r.Installed {
-			streams.ErrPrintln(ui.Dim(fmt.Sprintf("  installed %s for %s", name, r.Agent)))
+			streams.ErrPrintln(ui.Success(fmt.Sprintf("Installed %s skill for %s", name, r.Agent)))
 		}
 		for _, name := range r.Updated {
-			streams.ErrPrintln(ui.Dim(fmt.Sprintf("  updated %s for %s", name, r.Agent)))
+			streams.ErrPrintln(ui.Success(fmt.Sprintf("Updated %s skill for %s", name, r.Agent)))
 		}
 	}
 }
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -196,7 +196,7 @@ func installSkillsStep(streams iostream.Streams) {
 	if homeDir == "" {
 		return
 	}
-	for _, r := range skills.InstallSidecar(homeDir) {
+	for _, r := range skills.InstallByName(homeDir, "chunk-sidecar") {
 		if r.Skipped {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


michael-webster (approved):

internal/cmd/init.go
```diff
@@ -190,8 +191,49 @@ func ensureGitignoreEntries(workDir string, streams iostream.Streams) error {
...
+	}
+	yes, confirmErr := tui.Confirm("Install agent skills?", true)
+	if confirmErr != nil {
+		streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not confirm: %v", confirmErr)))
+		return
+	}
+	if !yes {
+		return
+	}
+	for _, r := range skills.Install(homeDir) {
```
michael-webster: per slack, it would be nice but not a blocker to have some kind of separation between skills we want to install e.g. for sidecars vs optional ones like code review

michael-webster (approved):
approved mod one non blocking comment

internal/skills/install.go
```diff
@@ -140,6 +148,46 @@ func installForAgent(agent Agent) AgentInstallResult {
 	return result
 }
 
+func installSidecarForAgent(agent Agent) AgentInstallResult {
```
michael-webster: we can cut some of this down. i'd either say do one of the following:
* put the skills we also want in a `skills/builtin` and the extras in `skills/extra`
* just add a function to install a skill by name instead of looping over them and seeing if they match

PR #303, opened 2026-05-04
Use an initial image that works

z00b:
Using the `/chunk-sidecar` skill took a lot of steps to get here, so use it as the starting point for future sidecars on this project.


schurchleycci (approved):
Looks like this works!

schurchleycci:
Actually hang on, didn't notice that chunk validate ran locally, not using this image

Edit: got this working, I wasn't passing the image from config properly.

PR #312, opened 2026-05-05
Auto-create sidecar from config image for per-command remote routing

schurchleycci:
## Summary

- When `validation.sidecarImage` is configured and a command is marked `remote: true`, `chunk validate` now auto-creates a sidecar even when not running in a Stop hook context
- Previously, auto-creation only fired in the Stop hook path (`hook != nil`); running `chunk validate` directly would fall through to "no active sidecar found — remote commands will run locally"
- Extracted the per-command sidecar resolution into `resolvePerCommandSidecarID` to make the branching logic easier to follow

## Test plan

- [ ] `TestValidateAutoCreatesSidecar` — verifies a sidecar is created using the configured image, the correct org is passed, and the new sidecar is actually used (add-key called on it)
- [ ] Existing `TestValidateRunRemoteUsesSSH` continues to pass (explicit `--sidecar-id` path unaffected)

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -150,7 +150,7 @@ func newValidateCmd() *cobra.Command {
 				}
 				statusFn(iostream.LevelInfo, fmt.Sprintf("running all commands on sidecar %s", sidecarID))
 			} else if cfg.HasRemoteCommands() {
-				resolvePerCommandSidecarID(cmd.Context(), &sidecarID, orgID, image, workDir, hook, streams, statusFn)
+				resolveSidecar(cmd.Context(), &sidecarID, orgID, image, workDir, hook, streams)
 			}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/validate.go
```diff
@@ -376,6 +365,27 @@ func resolveImage(name string, cfg *config.ProjectConfig) string {
 	return ""
 }
 
+// resolvePerCommandSidecarID fills sidecarID for per-command remote routing
+// (i.e. when --remote is not set but some commands have Remote:true).
+// It uses the active sidecar when available, auto-creates one when a sidecar
+// image is configured or the caller is a Stop hook, and warns otherwise.
+func resolvePerCommandSidecarID(ctx context.Context, sidecarID *string, orgID, image, workDir string, hook *hookContext, streams iostream.Streams, statusFn iostream.StatusFunc) {
```
michael-webster: Non blocking nits, but let's call this resolveSidecar. Also consider making the args list more compact.

z00b (approved):

PR #313, opened 2026-05-05
[FACT-129] Delete sidecar after successful snapshot

hanabel1:
Avoids leaking the build sidecar instance after a snapshot is captured. Clears local active-sidecar state when the deleted sidecar matches. Delete failures emit a warning but do not fail the snapshot command.

Screenshot 2026-05-05 at 4 23 00 PM


internal/cmd/sidecar.go
```diff
@@ -648,6 +654,19 @@ func newSidecarSnapshotCreateCmd() *cobra.Command {
 				return err
 			}
 			io.ErrPrintf("%s\n", ui.Success(fmt.Sprintf("Created snapshot %s", snap.ID)))
+
+			if err := sidecar.Delete(cmd.Context(), client, sidecarID); err != nil {
+				io.ErrPrintf("Warning: could not delete sidecar %s: %v\n", sidecarID, err)
+				io.ErrPrintf("Delete it manually in CircleCI to avoid leaking the build instance.\n")
```
michael-webster: we might want to change the wording here, since "leaking the build instance" is a little scary.

maybe `Delete the sidecar manually, or wait for it to expire`?

README.md
```diff
@@ -87,6 +87,8 @@ chunk sidecar snapshot create --name checkpoint
 chunk sidecar create --name new-sidecar --image <snapshot-id>
 ```
 
+`snapshot create` deletes the source sidecar once the snapshot is captured to avoid leaking the build instance — launch a new sidecar from the snapshot to resume work.
```
michael-webster: i'd reword this as well. just to make it clear that sidecars created as part of the snapshot process are not reusable

acceptance/sidecar_snapshot_test.go
```diff
@@ -87,6 +87,60 @@ func TestSidecarSnapshotCreateUsesActiveSidecar(t *testing.T) {
...
+	cci.DeleteStatusCode = 500
+	srv := httptest.NewServer(cci)
+	defer srv.Close()
+
+	env := testenv.NewTestEnv(t)
+	env.CircleCIURL = srv.URL
+
+	result := binary.RunCLI(t, []string{
+		"sidecar", "snapshot", "create",
+		"--sidecar-id", "sb-222",
```
michael-webster, line 134: might be worth using a random id here?

michael-webster (approved):
Left a few nits on the wording, but the logic here looks right.

internal/sidecar/sidecar.go
```diff
@@ -18,6 +18,10 @@ func Create(ctx context.Context, client *circleci.Client, orgID, name, image str
 	return client.CreateSidecar(ctx, orgID, name, image)
 }
 
+func Delete(ctx context.Context, client *circleci.Client, sidecarID string) error {
```
michael-webster: **High** — `sidecar.Delete` is a pass-through wrapper with no added value.

This function delegates directly to `client.DeleteSidecar` with no additional logic, making it a one-implementation wrapper. The review standards flag this pattern: avoid new functions/structs that wrap a single type without adding behaviour.

The concrete maintenance cost: callers in `cmd/` must import `internal/sidecar` and reason about whether logic lives in `sidecar.Delete` or `client.DeleteSidecar` when they are identical. Compare with how `client.CreateSnapshot` is called directly in `RunE` on line 648 — could `client.DeleteSidecar` follow the same pattern and let this wrapper be dropped?

internal/cmd/sidecar.go
```diff
@@ -648,6 +654,19 @@ func newSidecarSnapshotCreateCmd() *cobra.Command {
...
+				return nil
+			}
+			io.ErrPrintf("%s\n", ui.Success(fmt.Sprintf("Deleted sidecar %s", sidecarID)))
+
+			if active, lerr := sidecar.LoadActive(); lerr == nil && active != nil && active.SidecarID == sidecarID {
+				if cerr := sidecar.ClearActive(); cerr != nil {
```
michael-webster, line 666: **Medium** — inconsistent capitalisation on warning strings.

The delete-failure path (line 659) emits `"Warning: could not delete sidecar…"` (capital W), while the clear-active failure here emits `"warning: could not clear active sidecar state…"` (lowercase w). The acceptance test asserts on the capital-W form so there is no test failure, but the output is visually inconsistent to users.

```suggestion
			io.ErrPrintf("Warning: could not clear active sidecar state: %v\n", cerr)
```

michael-webster (commented):
Two findings from the automated review:

PR #315, opened 2026-05-07
sidecar sync: retry SSH on boot, simplify transient error detection, resolve base from sidecar

schurchleycci:
## Summary

- **SSH retry on boot**: `Sync` now retries `OpenSession` with exponential backoff (2s initial, 15s max interval, 90s max elapsed) on transient `net.Error` failures, so freshly-created sidecars don't require the user to re-run the command while SSH is still starting. Emits a status message on the first retry so users know what's happening.
- **Simpler `isTransientSSHError`**: replaced the blocklist approach (exclude known-permanent errors, retry everything else) with an allowlist (`net.Error` only). The allowlist has a safer default — unknown errors are not retried.
- **Remote `origin/HEAD` as sync base**: replaced local `gitutil.MergeBase()` with `git rev-parse origin/HEAD` on the sidecar, falling back to `HEAD` if `origin/HEAD` is not configured. This gives a more reliable base when the sidecar was freshly cloned, and removes the `git fetch` step that was adding latency on every sync.
- **Tests** for `isTransientSSHError` covering transient and non-transient cases.

## Test plan

- [x] `task test` passes (860 tests, 0 failures) — verified on sidecar
- [x] New `TestIsTransientSSHError` covers timeout, connection refused, wrapped net error, auth failure, status error, key-not-found, and generic error cases

### DIFF
--- a/internal/sidecar/sync.go
+++ b/internal/sidecar/sync.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
@@ -53,7 +54,7 @@ func persistWorkspace(ctx context.Context, workspace string) error {
 func Sync(ctx context.Context,
 	client *circleci.Client, sidecarID, identityFile, authSock, workdir string, status iostream.StatusFunc) error {
 
-	session, err := openSessionWithRetry(ctx, client, sidecarID, identityFile, authSock)
+	session, err := openSessionWithRetry(ctx, client, sidecarID, identityFile, authSock, status)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


internal/sidecar/sync.go
```diff
@@ -198,3 +204,39 @@ func syncWorkspace(ctx context.Context, status iostream.StatusFunc, org, repo, r
 	}
 	return nil
 }
+
+// openSessionWithRetry calls OpenSession, retrying on transient errors to give
+// a newly-created sidecar time to finish booting before its SSH service is ready.
+func openSessionWithRetry(ctx context.Context, client *circleci.Client, sidecarID, identityFile, authSock string, status iostream.StatusFunc) (*Session, error) {
```
z00b, line 221: @danmux or @pete-woods do we have a standard pattern or lib for retries? this feels just like something we would have solved elsewhere...

schurchleycci: This is using the standard lib for retries now! Thanks @danmux for the link

internal/sidecar/sync.go
```diff
@@ -155,10 +159,20 @@ func syncWorkspace(ctx context.Context, status iostream.StatusFunc, org, repo, r
 
 	status(iostream.LevelInfo, fmt.Sprintf("Synchronising local %s/%s to remote: %s...", org, repo, repoPath))
 
-	base, err := gitutil.MergeBase()
+	// Prefer origin/HEAD (set when the remote advertises a default branch), fall
+	// back to HEAD for repos where origin/HEAD is not configured.
+	baseCmd := fmt.Sprintf(
```
michael-webster, line 164: Where was the git fetch happening? We were only calling `rev-parse` and `merge-base` which shouldn't be doing a fetch?

michael-webster:
General question on this, do we want to make the ssh token upload happen as start of the sidecar create?

PR #317, opened 2026-05-07
Store per-sidecar workdir in .chunk/config.json

michael-webster:
## Summary

Agents calling `setup` with a `--workdir` flag, or `validate` without one, could resolve to different remote directories. This fixes that by relying on the workspace already persisted to the active sidecar state file.

`sidecar.Sync` already called `persistWorkspace` which saves the resolved path to `ActiveSidecar.Workspace` in the sidecar state file. `openSSHSession` already fell back to `active.Workspace`. So the fix is to wire them together properly rather than adding a separate map in `.chunk/config.json`.

- `setup` now accepts `--workdir` (consistent with `sync`); the resolved path is persisted to the active sidecar state via the existing `persistWorkspace` call in `Sync`
- `setup --skip-sync` reads workspace from active sidecar state
- `openSSHSession` resolution order: `--workdir` flag → `active.Workspace` → `./workspace`

## Test plan

- [ ] All existing tests pass (`go test -race ./...`)
- [ ] Manual: `chunk sidecar setup --workdir /workspace/my-repo` then `chunk validate` picks up the same path without a flag

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,6 @@ chunk sidecar snapshot create --name checkpoint
 chunk sidecar create --name new-sidecar --image <snapshot-id>
 ```
 
-**Note:** `snapshot create` consumes the source sidecar — it is deleted once the snapshot is captured and cannot be reused. To resume work, launch a new sidecar from the snapshot with `chunk sidecar create --image <snapshot-id>`.
-
 ### Context Generation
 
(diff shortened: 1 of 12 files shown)


michael-webster:
Abandoning — approach needs rethinking after rebase conflicts with main.

PR #323, opened 2026-05-11
Add --env/--env-file flags to validate and sidecar setup

schurchleycci:
## Summary

- Adds \`-e\`/\`--env KEY=VALUE\` and \`--env-file\` flags to \`chunk validate\` and \`chunk sidecar setup\`, forwarding vars into the remote SSH session
- \`--env-file\` defaults to \`.env.local\` and is loaded automatically if present (silently ignored if the file doesn't exist); pass an explicit path to override
- Env vars are resolved through the secrets layer before forwarding
- Flag precedence: \`--env\` wins over file values
- \`sidecar ssh\` refactored to use the shared \`resolveEnvVars\` helper (was inlining equivalent logic)

## Implementation notes

- \`resolveEnvVars\` lives in \`internal/cmd/env.go\` (shared by all three commands)
- For \`validate\`, \`--env\` syntax is validated eagerly so bad values are caught immediately even in dry-run/local mode; file loading and secrets resolution are deferred until a sidecar is actually in use
- \`internal/testing/fakes/ssh.go\` extended with \`EnvVars()\` to capture SSH env requests for test assertions

## Test plan

- [ ] \`TestOpenSSHSessionPassesEnvVars\` — verifies env vars arrive over SSH
- [ ] \`TestValidateEnvFlagBadValue\` — \`--env BADVALUE\` rejected immediately, including in dry-run mode
- [ ] \`TestParseEnvPairs/empty key returns error\` — \`=value\` is rejected

### DIFF
--- a/internal/cmd/env.go
+++ b/internal/cmd/env.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/secrets"
+	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
+)
+
+const defaultEnvFile = ".env.local"
+
+// resolveEnvVars builds the env var map from --env flags and --env-file. Flags win over file.
+func resolveEnvVars(ctx context.Context, workDir, envFile string, envVarsFlag []string) (map[string]string, error) {
+	flagVars, err := sidecar.ParseEnvPairs(envVarsFlag)
+	if err != nil {
+		return nil, &userError{msg: fmt.Sprintf("invalid --env value: %s", err), err: err}
+	}
+	var fileVars map[string]string
+	if envFile != "" {
+		path := envFile
+		if !filepath.IsAbs(path) {
+			path = filepath.Join(workDir, path)
+		}
+		fileVars, err = sidecar.LoadEnvFileAt(path)
+		if err != nil {
+			return nil, &userError{msg: fmt.Sprintf("load %s: %s", envFile, err), err: err}
+		}
+	}
+	envVars := sidecar.MergeEnv(fileVars, flagVars)
+	if len(envVars) > 0 {
+		envVars, err = secrets.ResolveAll(ctx, envVars, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return envVars, nil
+}
--- a/internal/cmd/env.go
+++ b/internal/cmd/env.go
@@ -32,7 +32,7 @@ func resolveEnvVars(ctx context.Context, workDir, envFile string, envVarsFlag []
 	if len(envVars) > 0 {
 		envVars, err = secrets.ResolveAll(ctx, envVars, nil)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("resolve secrets: %w", err)
 		}
 	}
 	return envVars, nil
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -820,7 +820,17 @@ Example:
 			}
 
 			// Step 6: Run setup steps over SSH.
-			if err := sidecarSetupRunSetup(cmd.Context(), client, sidecarID, identityFile, authSock, env, envVars, streams, status); err != nil {
+			opts := sidecarRunSetupOpts{
+				client:       client,
+				sidecarID:    sidecarID,
+				identityFile: identityFile,
+				authSock:     authSock,
+				env:          env,
+				envVars:      envVars,
+				streams:      streams,
+				status:       status,
+			}
+			if err := sidecarSetupRunSetup(cmd.Context(), opts); err != nil {
 				return err
 			}
 
@@ -931,17 +941,20 @@ func sidecarSetupSync(
 	return err
 }
 
-func sidecarSetupRunSetup(
-	ctx context.Context,
-	client *circleci.Client,
-	sidecarID, identityFile, authSock string,
-	env *envbuilder.Environment,
-	envVars map[string]string,
-	streams iostream.Streams,
-	status iostream.StatusFunc,
-) error {
-	if len(env.Setup) == 0 {
-		status(iostream.LevelInfo, "No setup steps detected, skipping")
+type sidecarRunSetupOpts struct {
+	client       *circleci.Client
+	sidecarID    string
+	identityFile string
+	authSock     string
+	env          *envbuilder.Environment
+	envVars      map[string]string
+	streams      iostream.Streams
+	status       iostream.StatusFunc
+}
+
+func sidecarSetupRunSetup(ctx context.Context, opts sidecarRunSetupOpts) error {
+	if len(opts.env.Setup) == 0 {
+		opts.status(iostream.LevelInfo, "No setup steps detected, skipping")
 		return nil
 	}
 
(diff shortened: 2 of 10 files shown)


michael-webster (commented):
## Review Summary

**Files reviewed**: 7
**Issues identified**: 4

---

### 1. Silent no-op: `--env` flags ignored on local runs without user feedback
**`internal/cmd/validate.go:149`** | **High**

When no sidecar is resolved, `resolveEnvVars` is skipped entirely — env vars are validated for syntax but never forwarded. A user who passes `--env FOO=bar` on a purely local `chunk validate` run gets no error and no warning; their flag silently has no effect. Consider emitting a warning via `streams.ErrPrintf` when `len(envVarsFlag) > 0` or `envFile != ""` and `sidecarID == ""`.

---

### 2. `nil` secrets delegate passed to `secrets.ResolveAll` without comment
**`internal/cmd/env.go:32`** | **High**

`secrets.ResolveAll(ctx, envVars, nil)` passes `nil` as the resolver delegate. If `ResolveAll` panics or errors on a `nil` delegate, failures will surface only when env vars are actually present — non-deterministic in tests. The old `sidecar.go` inline code also passed `nil`, but the refactor into a shared helper amplifies the blast radius across all three commands. Is `nil` a documented/supported code path in `secrets.ResolveAll`?

---

### 3. `sidecarSetupRunSetup`: `dir` may be `""` when resolving relative `--env-file` paths
**`internal/cmd/sidecar.go:776`** | **High**

`resolveEnvVars(cmd.Context(), dir, envFile, envVarsFlag)` passes `dir` (the `--dir` flag), which is `""` when not supplied. `sidecar ssh` explicitly calls `os.Getwd()` first; `sidecar setup` does not. With `dir == ""`, `filepath.Join("", ".env.local")` resolves relative to process cwd — which may work by accident today but is inconsistent and fragile.

Suggested fix: resolve `dir` to `os.Getwd()` when empty, mirroring how `sidecar ssh` handles it.

---

### 4. SSH fake env payload parsing: missing bounds check between name and value length fields
**`internal/testing/fakes/ssh.go:177`** | **Medium**

The manual SSH wire-format decoder checks `int(nameLen)+8 > len(req.Payload)` before reading the value length field. On a 32-bit platform, a `uint32` `nameLen` value > `math.MaxInt32` could silently overflow `int`, bypassing this check and causing a panic. Additionally, `TestOpenSSHSessionNoEnvVars` is listed in the PR test plan but is not present in `validate_test.go` — the no-env-vars path through `openSSHSession` appears untested.

internal/cmd/validate_test.go
```diff
@@ -0,0 +1,96 @@
...
+	cmd.SetOut(os.Stderr)
+	cmd.SetErr(os.Stderr)
+	cmd.SetArgs([]string{"--project", dir, "--env", "BADVALUE"})
+
+	err := cmd.Execute()
+	assert.Assert(t, err != nil)
+	assert.Assert(t, strings.Contains(err.Error(), "BADVALUE"), "got: %v", err)
+}
+
+func TestValidateEnvLocalRunIgnoresEnvFile(t *testing.T) {
```
michael-webster: I don't think we really need this particular test.

internal/testing/fakes/ssh.go
```diff
@@ -161,6 +174,28 @@ func (s *SSHServer) handleSession(ch ssh.Channel, requests <-chan *ssh.Request)
 		switch req.Type {
 		case "env":
 			_ = req.Reply(true, nil)
+			if len(req.Payload) < 4 {
+				s.t.Errorf("ssh fake: env payload too short (%d bytes)", len(req.Payload))
+				continue
+			}
+			nameLen := binary.BigEndian.Uint32(req.Payload[:4])
```
michael-webster, line 178: What are we checking here exactly?

michael-webster (commented):
So the code here is fine, but I'm not really understanding the intent.

Remember, we want the user flow to be "chunk validate" and stuff just works. we do the right thing to load environment settings, create a sidecar if necessary and so on. The env overrides _can_ be useful to debug I guess, but in terms of e.g. the hook configuration there probably won't be individual env flags.

internal/cmd/validate_test.go
schurchleycci: Removed!

schurchleycci:
> So the code here is fine, but I'm not really understanding the intent.
> 
> Remember, we want the user flow to be "chunk validate" and stuff just works. we do the right thing to load environment settings, create a sidecar if necessary and so on. The env overrides _can_ be useful to debug I guess, but in terms of e.g. the hook configuration there probably won't be individual env flags.

There's a specific problem I ran into running `chunk validate` with web-ui-consolidated:

- `node_modules` isn't part of the sidecar sync because it's gitignored
- as a result, installing dependencies in the sidecar is required
- we have private dependencies, so an npm token is needed in the sidecar to install them

So in order to get a sidecar where `chunk validate` will work, you'd have to manually ssh into the sidecar with e.g. your `NPM_TOKEN` forwarded, install the dependencies, then snapshot. And even after snapshotting, if you need to refresh the dependencies in the sidecar you'd have to do the same thing.

Thinking through this again, I think there's two separate gaps:
- `chunk sidecar setup` may need env vars in order to run the setup steps (so adding the env forwarding to that command feels reasonable?)
- `chunk sidecar sync` doesn't cover refreshing stale dependencies, so the `chunk sidecar sync` -> `chunk validate` flow won't help you if dependency updates are what you're trying to test. But perhaps that's better solved by doing frequent automatic builds or something similar?

All that to say, maybe we should just pare this down to adding env forwarding to setup and leaving validate as is?

michael-webster (approved):

internal/cmd/env.go
```diff
@@ -0,0 +1,39 @@
...
+		fileVars, err = sidecar.LoadEnvFileAt(path)
+		if err != nil {
+			return nil, &userError{msg: fmt.Sprintf("load %s: %s", envFile, err), err: err}
+		}
+	}
+	envVars := sidecar.MergeEnv(fileVars, flagVars)
+	if len(envVars) > 0 {
+		envVars, err = secrets.ResolveAll(ctx, envVars, nil)
+		if err != nil {
+			return nil, err
```
michael-webster: no we need to wrap this error?

internal/cmd/sidecar.go
```diff
@@ -387,8 +366,8 @@ func newSidecarSSHCmd() *cobra.Command {
 	cmd.Flags().StringVar(&sidecarID, "sidecar-id", "", "Sidecar ID (defaults to active sidecar)")
 	cmd.Flags().StringVar(&identityFile, "identity-file", "", "SSH identity file")
 	cmd.Flags().StringArrayVarP(&envVarsFlag, "env", "e", nil, "KEY=VALUE pairs to set in the remote session (repeatable)")
-	cmd.Flags().StringVar(&envFile, "env-file", "", "Env file to load (default .env.local when flag is present)")
-	cmd.Flags().Lookup("env-file").NoOptDefVal = ".env.local"
+	cmd.Flags().StringVar(&envFile, "env-file", defaultEnvFile, "Env file to load (default: .env.local; pass a path to override)")
```
michael-webster, line 369: Just trying to check what I think is the intent here:
- this line is setting `.env.local` as the default value that we'll resolve
- the `NoOptDefVal` will let someone pass `--env-file` on the CLI without an argument defaulting to our const?

internal/cmd/env.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/secrets"
+	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
+)
+
+const defaultEnvFile = ".env.local"
```
michael-webster, line 12: since we're moving to _always_ loading i wonder if we want to have this come from an env file _specifically_ for chunk? not blocking just a thought

internal/cmd/sidecar.go
```diff
@@ -832,8 +813,14 @@ Example:
 				}
 			}
 
-			// Step 5: Run setup steps over SSH.
-			if err := sidecarSetupRunSetup(cmd.Context(), client, sidecarID, identityFile, authSock, env, streams, status); err != nil {
+			// Step 5: Resolve env vars for SSH execution.
+			envVars, err := resolveEnvVars(cmd.Context(), dir, envFile, envVarsFlag)
```
michael-webster, line 816: Instead of creating extra args consider putting these into opts or a struct.

internal/cmd/env.go
michael-webster: at a minimum include some docs that `.env.local` files will be automatically read and used to passed to the running Chunk sidecar.

michael-webster (approved):
A few small nits, fine to merge if these are addressed.

internal/cmd/sidecar.go
schurchleycci: I think `NoOptDefVal` was a remnant of the old approach where we only loaded the env file if you passed the flag. I've removed it

PR #327, opened 2026-05-12
Add --all flag to sidecar list (user-scoped by default)

schurchleycci:
## Summary

- The sidecar API now returns only the current user's sidecars by default; passing `all=true` returns all org sidecars but requires org admin privileges
- Adds `--all` flag to `chunk sidecar list` to opt into org-wide listing
- Improves the 403 error message when `--all` is used without admin rights: _"Listing all sidecars requires org admin privileges."_

https://linear.app/circleci/issue/FACT-146/sidecar-list-should-show-users-sidecars-not-orgs-unless-all-is-passed

## Changes

- `internal/circleci/client.go` — `ListSidecars` accepts `all bool`; conditionally sends `all=true` query param
- `internal/sidecar/sidecar.go` — threads `all` through `List`
- `internal/cmd/sidecar.go` — adds `--all` flag; context-aware auth error message
- Tests updated and extended to cover both the default (user-scoped) and `--all` paths

## Test plan

- [ ] `chunk sidecar list` — returns only your sidecars
- [ ] `chunk sidecar list --all` — returns all org sidecars (or clear 403 message if not admin)
- [ ] `task test && task lint` — all pass

### DIFF
--- a/internal/circleci/circleci_test.go
+++ b/internal/circleci/circleci_test.go
@@ -62,7 +62,7 @@ func TestListSidecars(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("filters by org", func(t *testing.T) {
-		sidecars, err := client.ListSidecars(ctx, "org-1")
+		sidecars, err := client.ListSidecars(ctx, "org-1", false)
 		if err != nil {
 			t.Fatalf("unexpected error: %v", err)
 		}
@@ -75,7 +75,7 @@ func TestListSidecars(t *testing.T) {
 	})
 
 	t.Run("empty result", func(t *testing.T) {
-		sidecars, err := client.ListSidecars(ctx, "org-nonexistent")
+		sidecars, err := client.ListSidecars(ctx, "org-nonexistent", false)
 		if err != nil {
 			t.Fatalf("unexpected error: %v", err)
 		}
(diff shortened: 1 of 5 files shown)


michael-webster (commented):
## Review Summary

**Files reviewed**: 5
**Issues identified**: 2

---

### 1. `sidecar.List` is a pass-through wrapper with no added value
**`internal/sidecar/sidecar.go:13`** | **Medium**

After this change, `sidecar.List` does nothing beyond forwarding all arguments directly to `client.ListSidecars`. The review standards flag functions that wrap a single thing without adding behavior. Is there any reason this indirection layer should stay? Callers in `cmd/` could call `client.ListSidecars` directly, which would eliminate a layer of plumbing (the `all bool` parameter had to be threaded through this function for no gain).

---

### 2. New tests use raw `t.Fatalf`/`t.Errorf` instead of `gotest.tools/v3/assert`
**`internal/circleci/circleci_test.go:98`** | **Medium**

The new test cases ("omits all param by default", "sends all=true when requested") use raw `if err != nil { t.Fatalf(...) }` patterns. The project standard requires `gotest.tools/v3/assert` — `sidecar_test.go` already uses `assert.NilError` and `assert.Equal` correctly. Could these be updated to match?

michael-webster (approved):
Left a few small comment from the chunk-review LGTM mod those

schurchleycci:
> Left a few small comment from the chunk-review LGTM mod those

Got these fixed up so I'll merge now

PR #329, opened 2026-05-13
Add support for disabling hooks in a claude executable way

michael-webster:
We previously only had an env var to disable hooks. This was hard to work with in a given claude session, exposing a way via cli to disable hooks for dev, can be discovered by agents via the CLI interface.


schurchleycci (approved):
Claude only found a couple issues but none of them seemed relevant in practice. So LGTM!

PR #330, opened 2026-05-13
Hash session+branch for sidecar dedup (FACT-176)

michael-webster:
## Summary

- Sidecar state files are now named `sidecar.<sessionID>-<hash8>.json` where `hash8 = sha256(sessionID + ":" + branch)[:4]`, so concurrent sessions and different branches each get isolated state without the raw branch name appearing in the filename.
- Auto-named sidecars (created by `chunk validate` when no active sidecar exists) now use the same `<base>-<sessionID>-<hash8>` scheme instead of embedding the raw branch string directly. Falls back to the sanitised branch name when no session ID is present.

## Test plan

- [ ] `go test -race ./internal/sidecar/...` — unit tests for `sidecarFileName` hash stability and per-branch uniqueness
- [ ] `go test -race ./internal/cmd/...` — unit tests for `sidecarAutoName` with/without session ID
- [ ] Acceptance tests: `TestValidateHookMode_SessionIsolation` verifies two concurrent sessions see separate state files

### DIFF
--- a/acceptance/validate_test.go
+++ b/acceptance/validate_test.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/crypto/ssh"
 	"gotest.tools/v3/assert"
 
+	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
 	"github.com/CircleCI-Public/chunk-cli/internal/testing/binary"
 	testenv "github.com/CircleCI-Public/chunk-cli/internal/testing/env"
 	"github.com/CircleCI-Public/chunk-cli/internal/testing/fakes"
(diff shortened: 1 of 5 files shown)


schurchleycci:
Two things from claude:

**`currentBranch` logic is duplicated in three places** — the same `git -C <dir> rev-parse --abbrev-ref HEAD` pattern appears in `internal/sidecar/active.go` (as `currentBranch()`), inlined inside `sidecarAutoName()` in `internal/cmd/validate.go`, and again as `gitCurrentBranch()` in the acceptance test. Since `cmd/validate.go` already imports `internal/sidecar`, the simplest fix is to export `currentBranch` as `sidecar.CurrentBranch` and use it in both call sites. The acceptance test helper can stay local.

**`TestSidecarFileNameCases` hollow assertion for the both-present case** — the test checks structural properties of the filename (prefix, suffix, hash length) rather than the actual value. Since the hash is deterministic, consider pinning it directly using `hashFor()`:
```go
{"sess-1", "main", "sidecar.sess-1-" + hashFor("sess-1", "main") + ".json"},
```
Regressions become immediately visible without having to parse the filename apart.

schurchleycci (approved):
The comments from Claude seem worth addressing, but otherwise this looks good!

PR #332, opened 2026-05-13
Detect missing auth settings and exit hooks

michael-webster:
Detect missing CircleCI auth settings and exits code 1 to indicate a failure.


internal/cmd/usererr.go
```diff
@@ -25,6 +23,16 @@ const (
 
 const suggestionReauth = "Check your CircleCI token and try again."
 
+// silentExitError carries an exit code. main.go detects ExitCode() and calls
+// os.Exit directly, skipping further error printing. Use after writing your
+// own message to stderr.
+type silentExitError struct{ code int }
```
michael-webster, line 29: using this since we are controlling the output for the agent.

hanabel1 (approved):

hanabel1:
don't know if I'm missing it, but where is the chunk auth check command ..

michael-webster:
> don't know if I'm missing it, but where is the chunk auth check command ..

Ah stale PR description. I decided to instead detect the missing settings directly.

PR #335, opened 2026-05-13
Write .circleci/test-suites.yml during chunk init

hanabel1:
- `chunk init` now writes a starter `.circleci/test-suites.yml` for the user, so they don't have to write it from scratch.
  - Picks the right template based on the project: `go.mod` → gotestsum, `pyproject.toml` → pytest. Other toolchains are skipped.
  - Creates the `.circleci/` directory if it doesn't exist. Never overwrites an existing `test-suites.yml`.
  - New `--skip-test-suites` flag to turn this off.
  
  ## Tests
  - [x] `go test -race ./internal/validate/ -run TestTestSuitesTemplate` passes
  - [x] `go test -race ./acceptance/ -run "TestInit.*TestSuites"` passes
  - [x] Manual: in a fresh Go repo with no `.circleci/`, run `chunk init` → `.circleci/test-suites.yml` is created
  - [x] Manual: re-run `chunk init` → existing file is preserved, "already exists" message shown
  - [x] Manual: `chunk init --skip-test-suites` → no file written
  - [ ] Manual: in a repo with neither `go.mod` nor `pyproject.toml` → no file written

### DIFF
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -209,8 +209,36 @@ func installSkillsStep(streams iostream.Streams) {
 	}
 }
 
+// writeTestSuites scaffolds .circleci/test-suites.yml for CircleCI Smarter
+// Testing whenever the toolchain has a known template, creating .circleci/
+// if it does not already exist. It never overwrites an existing
+// test-suites.yml.
+func writeTestSuites(workDir string, streams iostream.Streams) error {
+	template := validate.TestSuitesTemplate(workDir)
+	if template == "" {
+		return nil
+	}
+
+	circleDir := filepath.Join(workDir, ".circleci")
+	path := filepath.Join(circleDir, "test-suites.yml")
+	if _, err := os.Stat(path); err == nil {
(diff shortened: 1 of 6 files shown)


internal/cmd/init.go
```diff
@@ -340,6 +375,7 @@ hook config files.`,
 	cmd.Flags().BoolVar(&skipValidate, "skip-validate", false, "Skip validate command detection")
 	cmd.Flags().BoolVar(&skipCompletions, "skip-completions", false, "Skip shell completion installation")
 	cmd.Flags().BoolVar(&skipSkills, "skip-skills", false, "Skip agent skill installation")
+	cmd.Flags().BoolVar(&skipTestSuites, "skip-test-suites", false, "Skip CircleCI test-suites.yml generation")
```
michael-webster: let's make this default to true since we have limited coverage on the language and add it in the agent instructions

also, can we include some kind of skill with smarter testing docs to have the agent generate?

michael-webster (approved):
I think we might want to have this off by default, but I don't feel that strongly. Approving to unblock.

I would also add in some agentic onboarding instructions in our skill to show the agent how to write the test-suites.yml

PR #336, opened 2026-05-13
Forward CIRCLE_TOKEN env vars to remote validate commands

hanabel1:
## Summary
- Commands run on a sidecar via `chunk validate --remote` couldn't reach CircleCI APIs (e.g. smarter-testing endpoints) because no host credentials crossed the SSH boundary.
- `openSSHSession` now forwards `CIRCLE_TOKEN` / `CIRCLECI_TOKEN` from the host environment through the existing `ExecOverSSH` `envVars` hook, mirroring how the local validate path picks them up implicitly.
- Added a small unit test covering the new `hostForwardEnv` helper.

## Test plan
- [x] `task fmt`, `task lint`, `go test -race ./...` all pass.
- [x] End-to-end manual check: `CIRCLE_TOKEN=… chunk validate --remote --cmd 'printf "CIRCLE_TOKEN length: %s\n" "${#CIRCLE_TOKEN}"'` prints the expected non-zero length (70-char personal token reached the remote shell).
- [ ] Reviewer to confirm `AcceptEnv` on the sidecar sshd doesn't need widening (test above succeeded without changes).

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -150,15 +150,7 @@ func newValidateCmd() *cobra.Command {
 			}
 
 			if dryRun {
-				if inlineCmd != "" {
-					cmdName := name
-					if cmdName == "" {
-						cmdName = "custom"
-					}
-					statusFn(iostream.LevelInfo, fmt.Sprintf("%s: %s", cmdName, inlineCmd))
-					return nil
-				}
-				return mapValidateError(validate.RunDryRun(cfg, name, statusFn))
+				return runValidateDryRun(cfg, name, inlineCmd, statusFn)
 			}
 
 			// Hook: fail early when CircleCI auth is missing and remote commands need it.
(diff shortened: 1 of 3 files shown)


internal/cmd/validate.go
```diff
@@ -328,6 +329,24 @@ func openSSHSession(ctx context.Context, sidecarID, identityFile, workdir string
 	return execFn, dest, nil
 }
 
+// hostForwardEnv collects host environment variables that should be forwarded
+// into commands running on the sidecar. CIRCLE_TOKEN / CIRCLECI_TOKEN let
+// remote validate commands authenticate to CircleCI APIs (e.g. smarter-testing
+// endpoints) using the caller's credentials, mirroring the local behavior
+// where these are picked up implicitly from the environment.
+func hostForwardEnv() map[string]string {
```
michael-webster: do we have something to this already for the normal command execution?

internal/cmd/validate.go
```diff
@@ -328,6 +329,24 @@ func openSSHSession(ctx context.Context, sidecarID, identityFile, workdir string
...
+	env := map[string]string{}
+	for _, name := range []string{config.EnvCircleToken, config.EnvCircleCIToken} {
```
michael-webster: I think we want to pull this off the config object. The circle token can be read from a file/keychain so it won't necessarily be in the environment

michael-webster (changes requested):
Left a comment, but I think we want this to use the config we load instead of assuming an env var. We really don't want to depend on the API token being in an env var.

internal/cmd/validate.go
```diff
@@ -318,8 +318,10 @@ func openSSHSession(ctx context.Context, sidecarID, identityFile, workdir string
 	cwd, _ := os.Getwd()
 	_, repo, _ := gitremote.DetectOrgAndRepo(cwd)
 	dest := sidecar.ResolveWorkspace(ctx, workdir, repo)
+	rc, _ := config.Resolve("", "")
```
michael-webster: we should probably return an error if we can't resolve the config.

michael-webster (approved):
One follow up comment to return an error if we can't resolve the config. But once that's fixed this is good to merge.

PR #337, opened 2026-05-14
Add chunk sidecar snapshot list command

stiyyagura0901:
Adds `chunk sidecar snapshot list --org-id <id>` so users can see existing snapshots before creating new ones, avoiding 409 name-conflict errors.

## Changes
- `ListSnapshots` client method hitting `GET /api/v2/sidecar/snapshots`
- `chunk sidecar snapshot list` subcommand with `--org-id` and `--json` flags
- Fake handler + acceptance tests

### DIFF
--- a/acceptance/sidecar_snapshot_test.go
+++ b/acceptance/sidecar_snapshot_test.go
@@ -220,6 +220,7 @@ func TestSidecarSnapshotMissingToken(t *testing.T) {
 	}{
 		{"create", []string{"sidecar", "snapshot", "create", "--sidecar-id", "sb-111", "--name", "snap"}},
 		{"get", []string{"sidecar", "snapshot", "get", "snap-abc"}},
+		{"list", []string{"sidecar", "snapshot", "list", "--org-id", "org-abc"}},
 	}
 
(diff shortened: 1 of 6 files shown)


internal/cmd/validate.go
```diff
@@ -77,6 +77,52 @@ func runValidateList(workDir string, jsonOut bool, streams iostream.Streams, sta
 	return validate.List(cfg, statusFn)
 }
 
+// applyHookContext sets up the context and streams for hook invocations.
+// It returns the updated context and streams, and resets per-session state
+// when the hook is running for the first time (not a retry).
+func applyHookContext(ctx context.Context, hook *hookContext, streams iostream.Streams) (context.Context, iostream.Streams) {
```
michael-webster: why is this included?

stiyyagura0901: It got added to fix a lint failure on the original commit. `newValidateCmd` was pushed over the `gocyclo` complexity limit.

michael-webster (approved):

michael-webster (approved):

PR #339, opened 2026-05-18
Redesign sidecar sync: baseline bootstrap + incremental patch

michael-webster:
## Summary

Replaces the current sync mechanism (merge-base recalculation + GitHub clone on every first sync) with a two-phase design:

- **Bootstrap** (`sidecar setup` or stale baseline): clones from GitHub if workspace absent, otherwise `git fetch`; checks out exact local HEAD when pushed or merge-base otherwise; applies a patch of all local changes; stores `BaselineRef` + `BaselineBranch`
- **Incremental sync** (`sidecar sync`, same branch, valid baseline): generates `git diff <BaselineRef> --binary` locally, resets sidecar to `BaselineRef`, applies patch — no GitHub access needed after initial setup

Staleness detection re-bootstraps automatically when the branch changes, no baseline is stored, or `BaselineRef` is no longer an ancestor of `HEAD` (rebase/force-push).

**Also fixes `MergeBase()` fallback**: was returning `git rev-parse origin/HEAD` (the remote tip, potentially ahead of local HEAD); now correctly uses `git merge-base HEAD origin/HEAD`.

**Remote commands** use `sh -c "cd <path> && git ..."` instead of `git -C` which is unavailable on the sidecar's older git version.

## Changes

- `internal/sidecar/active.go` — add `BaselineRef`, `BaselineBranch` fields
- `internal/gitutil/gitutil.go` — add `HeadRef`, `IsAncestor`; fix `MergeBase` fallback
- `internal/sidecar/sync.go` — replace `syncWorkspace`/`Sync` with `Bootstrap`/`Sync`; remove `errApplyFailed` retry logic
- `internal/cmd/sidecar.go` — `sidecarSetupSync` calls `Bootstrap`; `sync` command passes `cwd`
- `internal/sidecar/sync_test.go` — updated for new behaviour
- `acceptance/sidecar_test.go` — sync flag test now runs from a git repo

## Test plan

- [x] All unit and acceptance tests pass (`go test -race ./...`)
- [x] E2E script (`dev-utils/e2e-sync.sh`) passes all 4 scenarios on a live sidecar:
  - Fresh bootstrap (no baseline) → clone/fetch + patch applied
  - Incremental sync (same branch) → patch only, no re-clone
  - Branch switch → re-bootstrap triggered
  - Stale baseline (simulated rebase) → re-bootstrap triggered

### DIFF
--- a/acceptance/sidecar_test.go
+++ b/acceptance/sidecar_test.go
@@ -365,7 +365,11 @@ func TestSidecarsSshSyncFlags(t *testing.T) {
 			env := testenv.NewTestEnv(t)
 			env.CircleCIURL = srv.URL
 
-			result := binary.RunCLI(t, tt.args, env, env.HomeDir)
+			// sync runs git operations (branch detection) before opening SSH, so
+			// the working directory must be a git repo.
+			workDir := gitrepo.SetupGitRepo(t, "test-org", "test-repo")
+
+			result := binary.RunCLI(t, tt.args, env, workDir)
 
 			// Commands should fail at SSH key step, not at flag parsing
 			assert.Assert(t, result.ExitCode != 0, "expected non-zero exit (SSH fails)")
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -409,7 +409,11 @@ func newSidecarSyncCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			err = sidecar.Sync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, newStatusFunc(io))
+			cwd, err := os.Getwd()
+			if err != nil {
+				return fmt.Errorf("sync: %w", err)
+			}
+			err = sidecar.Sync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, cwd, newStatusFunc(io))
 			if err != nil {
 				if _, ok := errors.AsType[*sidecar.RemoteBaseError](err); ok {
 					return &userError{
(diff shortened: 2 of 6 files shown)


schurchleycci:
This seems like a good approach. A few Claude comments:

Issue 1: Growing patch size — Confirmed

In Sync (sync.go:~210), after successfully applying the patch, active.BaselineRef is never updated. The function loads the stored baseline, runs GeneratePatch(active.BaselineRef), resets, applies — then returns. Nothing writes a new BaselineRef back to state. Every subsequent call diffs from the original bootstrap SHA, so the patch grows linearly with accumulated commits. The fix would be to save active.BaselineRef = headSHA before returning.

Issue 2: Double reset in Bootstrap — Confirmed

The sequence in Bootstrap is:

git reset --hard HEAD && git clean -fd (pre-checkout)
git checkout <checkoutSHA>
git reset --hard <checkoutSHA> && git clean -fd (post-checkout, comment: “handles any detached HEAD drift”)
After step 2, HEAD is already at checkoutSHA in detached HEAD state. No other code runs between steps 2 and 3, so there’s nothing that could cause drift. Step 3 is an unconditional extra SSH round-trip with no effect.

Issue 3: Test coverage gaps — Confirmed

sync_test.go has exactly one test function. It covers only the stale-baseline-triggers-Bootstrap path (and specifically the sub-case where MergeBase fails). There is no test for:

The happy-path incremental sync: valid BaselineRef, same branch → patch applied, no re-bootstrap
The branch-changed trigger specifically: BaselineBranch != currentBranch → Bootstrap called
These two are the core decision paths in the new Sync logic (needsBootstrap evaluation) and aren’t exercised at all.

PR #341, opened 2026-05-19
sidecar sync: retry SSH on boot

schurchleycci:
## Summary

- Adds `openSessionWithRetry` to `sidecar.Sync`, retrying on transient `net.Error` values (connection refused, timeout) for up to 90s with exponential backoff — so a freshly-created sidecar doesn't fail immediately while its SSH service is still starting
- `isTransientSSHError` classifies retryable errors; permanent errors (auth failures, missing key) return immediately without retrying
- Adds `cenkalti/backoff/v4` as a direct dependency
- Extracts `hookEarlyExits` in `validate.go` to fix a pre-existing `gocyclo` lint failure (`newValidateCmd` was at complexity 31, limit is 30)

This is the SSH-related subset of `sidecar-sync-retry-fetch`, without the git changes (remote `origin/HEAD` resolution).

## Test plan

- [ ] `TestIsTransientSSHError` covers transient vs permanent error classification
- [ ] `TestSync_NonApplyFailureReturnsImmediately` continues to pass (no regression to existing sync behaviour)
- [ ] `task test` passes (845 tests)

### DIFF
--- a/internal/sidecar/ssh.go
+++ b/internal/sidecar/ssh.go
@@ -16,12 +16,16 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/coder/websocket"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/term"
 
 	"github.com/CircleCI-Public/chunk-cli/internal/closer"
+	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 )
 
 // ExecResult holds the output of a command executed over SSH.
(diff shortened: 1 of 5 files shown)


internal/sidecar/ssh.go
```diff
@@ -336,3 +340,62 @@ func tofuHostKeyCallback(knownHostsPath, host string) ssh.HostKeyCallback {
 		return err
 	}
 }
+
+// waitForSSHReady probes the SSH connection with exponential backoff, retrying
+// on transient errors so a newly-created sidecar has time to finish booting
+// before its SSH service accepts connections.
+func waitForSSHReady(ctx context.Context, session *Session, status iostream.StatusFunc) error {
```
michael-webster, line 347: is this reimplementing the retry mechanism we already have server side? i'm less concerned about the implementation here than i am knowing if the retries are going to help.

schurchleycci: Tbh that's why I'd parked this PR for a bit, I wanted to go back and check. Upon further review I don't think this PR will actually solve any of the issues we've been seeing - I'll close it!

PR #347, opened 2026-05-20
Store credentials in system keychain

michael-webster:
## Summary

- Adds a `internal/keyring` package wrapping `go-keyring` with a 3-second timeout and service keys namespaced by provider + base URL (supporting multiple server instances)
- `config.Resolve` now reads from the system keychain between env-vars and config-file layers when `--insecure-storage` is not set
- `auth set/remove/status` commands read and write from the keychain by default; `--insecure-storage` falls back to the config file
- All `ensure*Client` helpers now receive the `*cobra.Command` so they can read `--insecure-storage` without it being threaded as a separate parameter
- Acceptance tests prepend `--insecure-storage` automatically so they never trigger a keychain prompt

## Review notes

The four commits are intended to be reviewed in order — each builds cleanly on the previous one:

1. **`internal/keyring` package** — pure library, no command changes
2. **`--insecure-storage` + `config.Resolve`** — flag wiring and priority order, no UI changes
3. **`auth` commands** — the user-facing keychain behaviour
4. **`cmd` threading + test bypass** — mechanical call-site updates

## Test plan

- [x] `task test` passes (all packages including acceptance)
- [x] `chunk auth set circleci` stores token in keychain; `security find-generic-password -s "com.circleci.chunk:circleci:https://circleci.com"` confirms entry
- [x] `chunk auth remove circleci` removes it; subsequent lookup returns not-found
- [x] `chunk auth set circleci --insecure-storage` writes to config file instead; keychain entry absent
- [x] With `CIRCLE_TOKEN` set, `chunk auth remove circleci` reports "No CircleCI token stored" (nothing to remove from storage)

### DIFF
--- a/internal/authprompt/authprompt.go
+++ b/internal/authprompt/authprompt.go
@@ -11,6 +11,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/config"
 	"github.com/CircleCI-Public/chunk-cli/internal/github"
 	hc "github.com/CircleCI-Public/chunk-cli/internal/httpcl"
+	"github.com/CircleCI-Public/chunk-cli/internal/keyring"
 	"github.com/CircleCI-Public/chunk-cli/internal/version"
 )
 
@@ -114,8 +115,12 @@ func ResolveGitHubClient(rc config.ResolvedConfig, logStatus func(string)) (*git
 	})
 }
 
-// SaveCircleCIToken persists a CircleCI token to the config file.
-func SaveCircleCIToken(token string) error {
+// SaveCircleCIToken persists a CircleCI token. When secureStorage is true it uses
+// the system keychain; when false it falls back to the config file.
+func SaveCircleCIToken(token, baseURL string, secureStorage bool) error {
+	if secureStorage {
+		return keyring.Set(keyring.ServiceCircleCI(baseURL), token)
+	}
 	cfg, err := config.Load()
 	if err != nil {
 		return fmt.Errorf("load config: %w", err)
@@ -127,8 +132,12 @@ func SaveCircleCIToken(token string) error {
 	return nil
 }
 
-// SaveAnthropicKey persists an Anthropic API key to the config file.
-func SaveAnthropicKey(key string) error {
+// SaveAnthropicKey persists an Anthropic API key. When secureStorage is true it uses
+// the system keychain; when false it falls back to the config file.
+func SaveAnthropicKey(key, baseURL string, secureStorage bool) error {
+	if secureStorage {
+		return keyring.Set(keyring.ServiceAnthropic(baseURL), key)
+	}
 	cfg, err := config.Load()
 	if err != nil {
 		return fmt.Errorf("load config: %w", err)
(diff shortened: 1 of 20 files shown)


internal/authprompt/authprompt.go
```diff
@@ -114,8 +115,12 @@ func ResolveGitHubClient(rc config.ResolvedConfig, logStatus func(string)) (*git
 	})
 }
 
-// SaveCircleCIToken persists a CircleCI token to the config file.
-func SaveCircleCIToken(token string) error {
+// SaveCircleCIToken persists a CircleCI token. When secureStorage is true it uses
+// the system keychain; when false it falls back to the config file.
+func SaveCircleCIToken(token, baseURL string, secureStorage bool) error {
```
michael-webster: Base url is included since we need to support multiple namespaced keys.

internal/cmd/validate.go
```diff
@@ -227,16 +229,21 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 
 	image := resolveImage(name, cfg)
 
+	circleCIClient, err := ensureCircleCIClient(cmd.Context(), cmd, rc, streams, tui.PromptHidden)
```
michael-webster: hoisting this up since we were previously calling resolve further down the tree. this made the already suboptimal dependency on the FS in resolve more difficult with the keychain dependency in the mix.

schurchleycci: From Claude:

> The previous code used a guard that short-circuited when no remote operations were needed (no --remote, no sidecar ID, no remote commands in config). This PR removes that guard and replaces it with an unconditional ensureCircleCIClient call. Any user running chunk validate with purely local commands and no CircleCI token will now receive an auth prompt — or, in non-interactive contexts (CI, hooks), an auth error — where none was required before.
> 
> Suggested fix: restore the conditional before calling ensureCircleCIClient

schurchleycci:
Not sure how big of a deal this is in practice:

> 1. config package imports internal/keyring, violating the documented leaf constraint
internal/config/config.go:15 | High
> 
> The architecture documentation explicitly states: “config/ is a leaf — no imports from other internal/ packages.” This PR introduces "github.com/CircleCI-Public/chunk-cli/internal/keyring" as a dependency of config.
> 
> The concrete consequence: config is imported by nearly every package in the codebase. Pulling an OS-level keychain integration into that leaf now means every test that imports config transitively depends on go-keyring and its platform backends (wincred, godbus/dbus). Tests that previously ran hermetically may behave differently on CI systems without a keychain daemon.
> 
> The fix is to move the keychain reads out of Resolve and into the cmd/ layer, passing the resolved tokens in as parameters — the same pattern used for flag and env values today. Alternatively, Resolve could accept the already-fetched keychain tokens as optional parameters alongside flagAPIKey/flagModel.

schurchleycci:
Running the commands locally everything is working as expected. It would be nice to warn the user if they have a token stored insecurely when they run `chunk auth remove circleci` without the `--insecure-storage` flag but not a blocker.

schurchleycci (approved):

michael-webster:
> Not sure how big of a deal this is in practice:
> 
> > 1. config package imports internal/keyring, violating the documented leaf constraint
> >    internal/config/config.go:15 | High
> > 
> > The architecture documentation explicitly states: “config/ is a leaf — no imports from other internal/ packages.” This PR introduces "github.com/CircleCI-Public/chunk-cli/internal/keyring" as a dependency of config.
> > The concrete consequence: config is imported by nearly every package in the codebase. Pulling an OS-level keychain integration into that leaf now means every test that imports config transitively depends on go-keyring and its platform backends (wincred, godbus/dbus). Tests that previously ran hermetically may behave differently on CI systems without a keychain daemon.
> > The fix is to move the keychain reads out of Resolve and into the cmd/ layer, passing the resolved tokens in as parameters — the same pattern used for flag and env values today. Alternatively, Resolve could accept the already-fetched keychain tokens as optional parameters alongside flagAPIKey/flagModel.

So I think there's an arguable violation here, but the fix is wrong. We'd want to just move all of the config loading into a single place vs having the command layer decide to read from the key ring. But I think this design is out of scope for this PR

michael-webster:
> Running the commands locally everything is working as expected. It would be nice to warn the user if they have a token stored insecurely when they run `chunk auth remove circleci` without the `--insecure-storage` flag but not a blocker.

fixed.

PR #354, opened 2026-05-21
Auto-sync local changes to sidecar before remote validation

schurchleycci:
## Summary

- `chunk validate` now automatically syncs local changes to the sidecar before running any remote commands, whenever a sidecar is in use (`--remote`, `--sidecar-id`, or commands with `Remote:true` in config)
- Fixes a correctness gap where hooks triggering remote validation could run against a stale sidecar workspace instead of the current working tree
- Adds `--no-sync` flag to opt out when the existing sidecar state is intentional (e.g. validating a specific pinned snapshot)

## Test plan

- [ ] Run `chunk validate --remote` with local uncommitted changes — confirm sync runs before commands execute
- [ ] Run `chunk validate --remote --no-sync` — confirm sync is skipped
- [ ] Trigger via Stop hook with a remote-routed command — confirm sidecar reflects current working tree
- [ ] Run on unpushed branch — confirm `RemoteBaseError` surfaces a clear "push your branch" message
- [ ] Run with no sidecar (local-only commands) — confirm sync is not called

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -241,16 +241,9 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return err
 	}
 
-	freshlyCreated := false
-	if opts.remote {
-		// --remote: force all commands to sidecar, creating one if needed.
-		freshlyCreated, err = resolveOrCreateSidecarID(ctx, circleCIClient, &opts.sidecarID, opts.orgID, image, workDir, streams)
-		if err != nil {
-			return err
-		}
-		statusFn(iostream.LevelInfo, fmt.Sprintf("running all commands on sidecar %s", opts.sidecarID))
-	} else if cfg.HasRemoteCommands() {
-		freshlyCreated = resolveSidecar(ctx, circleCIClient, &opts.sidecarID, opts.orgID, image, workDir, hook, streams)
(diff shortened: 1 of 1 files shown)


internal/cmd/validate.go
```diff
@@ -382,6 +382,40 @@ func runValidate(ctx context.Context, client *circleci.Client, rc config.Resolve
...
+func setupRemote(ctx context.Context, client *circleci.Client, opts *validateOpts, image string, cfg *config.ProjectConfig, hook *hookContext, statusFn iostream.StatusFunc, workDir string, streams iostream.Streams) (bool, error) {
+	if opts.remote {
+		created, err := resolveOrCreateSidecarID(ctx, client, &opts.sidecarID, opts.orgID, image, workDir, streams)
+		if err != nil {
+			return false, err
+		}
+		statusFn(iostream.LevelInfo, fmt.Sprintf("running all commands on sidecar %s", opts.sidecarID))
+		return created, nil
+	}
+	if cfg.HasRemoteCommands() {
```
michael-webster: i think we can combine this with the check above?

michael-webster (approved):

internal/cmd/validate.go
```diff
@@ -87,6 +87,7 @@ type validateOpts struct {
 	list         bool
 	save         bool
 	remote       bool
+	noSync       bool
```
michael-webster: approved but i'd remove this.

schurchleycci: Good call, done!

PR #359, opened 2026-05-22
Make workdir consistent with the sidecar structure.

michael-webster:
Chunk sidecars use /home/user as the home dir. This makes the command logic reference that known location instead of relative paths.

## Changes

- Replace hardcoded \`/home/user\` constant with \`sidecarHome()\`, which reads \`CHUNK_SIDECAR_HOME\` env var and falls back to \`/home/user\`. This allows the base path to be overridden if the sidecar image ever changes its OS user.
- \`ResolveWorkspace\` now returns \`(string, error)\` and errors when \`repo\` is empty and no workspace is saved — previously this would silently return the bare home dir, which the retry \`rm -rf\` path would then wipe entirely.
- Update callers in \`sync.go\` and \`validate.go\` to handle the error.
- Add tests for the empty-repo error case and the env var override.

### DIFF
--- a/internal/sidecar/sync.go
+++ b/internal/sidecar/sync.go
@@ -14,21 +14,31 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 )
 
-const workspaceDir = "/home/user"
+// sidecarHome returns the base home directory on the sidecar. It reads
+// CHUNK_SIDECAR_HOME so the default "/home/user" can be overridden when the
+// image uses a different OS user.
+func sidecarHome() string {
+	if h := os.Getenv("CHUNK_SIDECAR_HOME"); h != "" {
+		return h
+	}
+	return "/home/user"
+}
 
 // ResolveWorkspace determines the workspace path. Priority:
-// 1. CLI --workdir flag  2. sidecar.json workspace  3. default /home/user/<repo>.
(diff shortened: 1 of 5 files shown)


internal/sidecar/sync.go
```diff
@@ -14,10 +14,10 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 )
 
-const workspaceDir = "./workspace"
+const workspaceDir = "/home/user"
```
danmux: If `repo` is empty, `ResolveWorkspace` returns bare `/home/user`. The retry path at line 92 runs `rm -rf` on whatever `ResolveWorkspace` returned — previously that would have been `rm -rf ./workspace` (contained), now it's `rm -rf /home/user` (wipes the sidecar home dir). `DetectOrgAndRepo` guards the `Sync` flow, but `ResolveWorkspace` is public — shouldn't this at least validate that the final path is deeper than the base before returning it?

danmux: The old relative `./workspace` worked regardless of which OS user ran the process. `/home/user` assumes the sidecar user is literally `user` — if the image ever changes the default user, or someone invokes this outside the expected sidecar, it silently targets the wrong location. Worth deriving from `$HOME` or an env var instead?

danmux (approved):
clod comments appear to have some merit on a cursory glance

PR #391, opened 2026-05-28
detect Python, Ruby, Java, Gradle toolchains in chunk init

### LINKED ISSUE #322: chunk init writes npm test as fallback for unrecognized toolchains; lacks requirements.txt case
## Summary

`chunk init` writes `"run": "npm test"` to `.chunk/config.json` and wires `npm test` into the Claude Code `PreToolUse` hook when run in a project whose toolchain is not in the deterministic detection list in [`internal/validate/setup.go`](https://github.com/CircleCI-Public/chunk-cli/blob/main/internal/validate/setup.go) and when no `ANTHROPIC_API_KEY` is configured. In practice this fires for Python projects that use `requirements.txt` (rather than `pyproject.toml`), for Ruby (`Gemfile`), for Maven/Gradle, and for any other stack not enumerated in `DetectCommands`.

The same project is correctly identified by `chunk sidecar env`, which uses the detector in `envbuilder/`. The two detection paths disagree.

## Environment

- `chunk` version: 0.7.44 (Homebrew tap `CircleCI-Public/circleci/chunk`)
- OS: macOS 25.4.0 (Apple Silicon)
- `ANTHROPIC_API_KEY`: unset (the user-facing docs describe this as optional for `init`)

## Steps to reproduce

In an empty directory, create the following four files. No `package.json`, no `pyproject.toml`, no `node_modules`, no JavaScript lockfile.

`calculator.py`:

```python
def add(a: int, b: int) -> int:
    return a + b
```

`test_calculator.py`:

```python
from calculator import add

def test_add():
    assert add(1, 2) == 3
```

`requirements.txt`:

```
pytest==8.3.3
```

Initialize git so `chunk init` can resolve VCS metadata:

```sh
git init -q
git add -A
git -c user.email=test@test -c user.name=test commit -q -m initial
git remote add origin https://github.com/example/repro.git
```

Then run, with `ANTHROPIC_API_KEY` unset:

```sh
chunk init --skip-completions --skip-skills </dev/null
```

## Observed behavior

```
Detected repository: example/repro
Detected command: test (npm test)
✓ Wrote .chunk/config.json
✓ Updated .gitignore with sidecar tracking patterns
✓ Wrote .claude/settings.json
✓ Project initialized
```

`.chunk/config.json`:

```json
{
  "commands": [
    { "name": "test", "run": "npm test", "role": "gate" }
  ],
  "vcs": { "org": "example", "repo": "repro" }
}
```

`.claude/settings.json` `PreToolUse` hook (truncated):

```json
{
  "matcher": "Bash(git commit*)",
  "hooks": [
    {
      "type": "command",
      "command": "cd ${CLAUDE_PROJECT_DIR:-.} && npm test",
      "timeout": 60
    }
  ]
}
```

Every subsequent Claude Code commit in this repository invokes `npm test`, which fails because `npm` is not installed and no `package.json` exists.

## Cross-check with `chunk sidecar env`

Running `chunk sidecar env` in the same directory returns the correct stack and commands:

```json
{
  "stack": "python",
  "setup": [
    { "name": "install", "command": "pip install -r requirements.txt" },
    { "name": "test", "command": "pytest" }
  ],
  "image": "cimg/python",
  "image_version": "3.14.3"
}
```

So the binary already contains a working detector for this case; `chunk init` does not use it.

## Root cause

In [`internal/validate/setup.go`](https://github.com/CircleCI-Public/chunk-cli/blob/main/internal/validate/setup.go):

```go
const defaultTestCommand = "npm test"
```

`DetectCommands` switches on a fixed set of root-level files:

- `Taskfile.yml` / `Taskfile.yaml`
- `Makefile`
- `go.mod`
- `Cargo.toml`
- `pyproject.toml`
- `package.json`

If none match and Claude is unavailable, the fallback is:

```go
if claude == nil {
    return []config.Command{{Name: "test", Run: defaultTestCommand, Role: config.RoleGate}}, nil
}
```

Two issues compound here:

1. **`requirements.txt`, `setup.py`, and `Pipfile` are not in the switch.** Many Python projects (and most introductory tutorials) do not use `pyproject.toml`. The signals are recognized further down in `gatherRepoContext`, which adds them to the Claude prompt, but the deterministic path never consults them. The same gap exists for Ruby (`Gemfile`), Java (`pom.xml`, `build.gradle*`), and Clojure (`project.clj`/`deps.edn`) — all listed in `gatherRepoContext` but absent from the switch.

2. **`defaultTestCommand = "npm test"` is a misleading fallback.** When the toolchain cannot be determined, the safer behaviors are (a) emit no commands and surface a warning, or (b) emit an explicit no-op like `echo "no test command configured" && false`. Writing `npm test` for a Python or Ruby project produces a hook that is guaranteed to fail on every commit.

## Suggested fix

In `internal/validate/setup.go`:

1. Add a Python case to the switch that triggers on `requirements.txt`, `requirements-dev.txt`, `requirements-test.txt`, `setup.py`, or `Pipfile`, emitting `pytest` as the test command (consistent with the existing `pyproject.toml` branch).
2. Optionally add Ruby (`Gemfile` → `bundle exec rspec` or `rake test`), Maven (`pom.xml` → `mvn test`), and Gradle (`build.gradle*` → `./gradlew test`) cases to round out the deterministic path.
3. Change the `claude == nil` fallback from `npm test` to either (a) returning an empty command list with a warning logged, or (b) consulting the `envbuilder/` detector that `chunk sidecar env` already uses successfully.

## Impact

`chunk init` is the first command in the documented quick-start. For any non-Node project that does not happen to have `pyproject.toml`, `Cargo.toml`, `go.mod`, a `Makefile`, or a `Taskfile`, the resulting `.chunk/config.json` and Claude `PreToolUse` hook are misconfigured, and the user has to manually edit both files before Claude-driven commits will work.

hanabel1:
Fixes https://github.com/CircleCI-Public/chunk-cli/issues/322

## Summary

`chunk init` was falling back to `npm test` for any toolchain not in its deterministic detection list (e.g. Python with `requirements.txt`, Ruby, Java, Gradle). This meant the generated `.chunk/config.json` and Claude `PreToolUse` hook were broken out of the box for those stacks.

- Add deterministic detection for Python (`requirements.txt`, `setup.py`, `Pipfile`), Ruby (`Gemfile`), Maven (`pom.xml`), and Gradle (`build.gradle`, `build.gradle.kts`) in `DetectCommands`
- Return no commands instead of a broken `npm test` fallback when toolchain is unknown and Claude is unavailable

## Test plan

- [x] All 905 tests pass (`task test`)
- [x] `TestInitDetectsUnknownToolchainNoClaude` updated to verify no commands are returned for unknown toolchains
- [ ] Manual: run `chunk init` in a Python project with only `requirements.txt` — should detect `pytest`
- [ ] Manual: run `chunk init` in an empty project without `ANTHROPIC_API_KEY` — should produce no test command

### DIFF
--- a/acceptance/init_test.go
+++ b/acceptance/init_test.go
@@ -322,6 +322,160 @@ func TestInitDetectsPyprojectCommands(t *testing.T) {
 	assert.Equal(t, test["run"], "pytest")
 }
 
+func TestInitDetectsRequirementsTxtCommands(t *testing.T) {
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
+	assert.NilError(t, os.WriteFile(filepath.Join(workDir, "requirements.txt"), []byte("pytest\n"), 0o644))
+
+	env := testenv.NewTestEnv(t)
+	env.AnthropicKey = ""
+
+	result := binary.RunCLI(t, []string{
+		"init", "--skip-hooks",
+	}, env, workDir)
+
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
+	cfg := readInitConfig(t, workDir)
+	test := commandByName(cfg, "test")
+	assert.Assert(t, test != nil, "expected test command")
+	assert.Equal(t, test["run"], "pytest")
+}
+
+func TestInitDetectsSetupPyCommands(t *testing.T) {
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
+	assert.NilError(t, os.WriteFile(filepath.Join(workDir, "setup.py"), []byte("from setuptools import setup\nsetup(name='test')\n"), 0o644))
+
+	env := testenv.NewTestEnv(t)
+	env.AnthropicKey = ""
+
+	result := binary.RunCLI(t, []string{
+		"init", "--skip-hooks",
+	}, env, workDir)
+
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
+	cfg := readInitConfig(t, workDir)
+	test := commandByName(cfg, "test")
+	assert.Assert(t, test != nil, "expected test command")
+	assert.Equal(t, test["run"], "pytest")
+}
+
+func TestInitDetectsPipfileCommands(t *testing.T) {
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
+	assert.NilError(t, os.WriteFile(filepath.Join(workDir, "Pipfile"), []byte("[packages]\npytest = \"*\"\n"), 0o644))
+
+	env := testenv.NewTestEnv(t)
+	env.AnthropicKey = ""
+
+	result := binary.RunCLI(t, []string{
+		"init", "--skip-hooks",
+	}, env, workDir)
+
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
+	cfg := readInitConfig(t, workDir)
+	test := commandByName(cfg, "test")
+	assert.Assert(t, test != nil, "expected test command")
+	assert.Equal(t, test["run"], "pytest")
+}
+
+func TestInitDetectsGemfileCommands(t *testing.T) {
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
+	assert.NilError(t, os.WriteFile(filepath.Join(workDir, "Gemfile"), []byte("source 'https://rubygems.org'\ngem 'rails'\n"), 0o644))
+
+	env := testenv.NewTestEnv(t)
+	env.AnthropicKey = ""
+
+	result := binary.RunCLI(t, []string{
+		"init", "--skip-hooks",
+	}, env, workDir)
+
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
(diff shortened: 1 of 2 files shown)


schurchleycci:
Generally looks good! A few minor notes from Claude:

> 1. Missing acceptance tests for every new detection path (significant)
> The project follows a strong "integration over mocks" philosophy, and each existing toolchain has a dedicated TestInitDetects* acceptance test. This PR adds 5 new detection paths but only updates an existing test — none of the new paths are covered.
> 
> The pattern is already established — see TestInitDetectsPyprojectCommands at line 306. Each test creates a temp git repo, writes the trigger file, runs chunk init --skip-hooks, and asserts the generated command. These would take ~10 lines each and directly validate what this PR claims to fix.
> 
> 2. Ruby command assumes rake test (minor, but worth flagging)
> bundle exec rake test only works if a Rakefile with a test task exists. Many common Ruby stacks use:
> 
> RSpec → bundle exec rspec
> Rails → bundle exec rails test or bin/rails test
> Minitest standalone → ruby -Itest test/**/*_test.rb
> A Gemfile alone doesn't imply rake test. A more defensive approach would be to also check for the presence of a Rakefile, or to let Claude handle it (since these cases are less uniform). At minimum, a comment explaining the choice would help reviewers.
> 
> 3. ./gradlew test assumes the wrapper is committed (minor)
> ./gradlew test is standard practice and generally fine. However, if the wrapper isn't present (e.g., .gitignore excludes it), the command will fail. Could check for gradlew existence, or fall back to gradle test. Low priority but worth a note.

schurchleycci (approved):

michael-webster (approved):

PR #403, opened 2026-06-15
Generate .codex/hooks.json during chunk init

schurchleycci:
## Summary

- Adds Codex agent support to `chunk init` alongside the existing Claude Code hook setup
- `BuildCodex` generates `.codex/hooks.json` with the same PreToolUse/Stop hook structure as `Build`, but without Claude Code-specific metadata (`$schema`, `_comment`, `permissions`)
- `MergeCodex` preserves user-defined hook groups by merging chunk-managed entries in place — matching the merge-by-identity strategy used for `.claude/settings.json` (PreToolUse merged by matcher, Stop merged by `chunk validate` command identity)

## Test plan

- [ ] `task test` passes
- [ ] `chunk init` on a fresh project creates `.codex/hooks.json` with PreToolUse and Stop hooks
- [ ] Re-running `chunk init` on an existing `.codex/hooks.json` shows a diff and prompts to apply
- [ ] Declining the prompt writes `.codex/hooks.example.json` and leaves `hooks.json` untouched
- [ ] User-added Stop hooks in an existing `hooks.json` are preserved after re-running `chunk init`


internal/cmd/init.go
```diff
@@ -127,6 +127,106 @@ func writeSettings(workDir string, commands []config.Command, streams iostream.S
 	return nil
 }
 
+// writeCodexHooks writes .codex/hooks.json for the project.
+// Uses the same merge/confirm/fallback pattern as writeSettings.
+func writeCodexHooks(workDir string, commands []config.Command, streams iostream.Streams, confirm confirmFunc) error {
```
michael-webster, line 147: can we detect if codex is on the user's path or if there's a settings folder before we install?

schurchleycci: Good call, updated!

internal/settings/settings.go
```diff
@@ -95,3 +100,61 @@ func Build(commands []config.Command) ([]byte, error) {
...
+
+	s := codexHooksJSON{}
+
+	if len(hooks) > 0 {
+		const maxStopTimeout = 600
+		stopTimeout := 30
+		for _, cmd := range commands {
+			t := cmd.Timeout
+			if t == 0 {
+				t = 60
```
michael-webster: do we need to accumulate the timeout like this?

internal/cmd/init.go
```diff
@@ -127,6 +127,121 @@ func writeSettings(workDir string, commands []config.Command, streams iostream.S
...
+			streams.ErrPrintln(line)
+		}
+	}
+
+	apply, confirmErr := confirm("Apply changes to .codex/hooks.json?", false)
+	if confirmErr != nil {
+		streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not confirm: %v", confirmErr)))
+	}
+	if confirmErr != nil || !apply {
+		return writeCodexHooksExample(dir, generated, streams)
```
michael-webster, line 215: do we want to write the example even if the user declines?

michael-webster (approved):
A few suggested UX tweaks e.g. try and detect if codex is present before the install. Otherwise this looks good.

internal/settings/settings.go
schurchleycci: Fixed!

internal/cmd/init.go
schurchleycci: Good catch, fixed this

schurchleycci:
> A few suggested UX tweaks e.g. try and detect if codex is present before the install. Otherwise this looks good.

All good suggestions! I've addressed all of them now 👍

PR #406, opened 2026-06-16
Fix npm not found on bare Ubuntu by adding Node.js system step

hanabel1:
## Summary

- Node/JS/TS projects now get a `system` setup step that installs Node.js via NodeSource before `npm install` / `yarn install` runs
- Uses the detected major version (e.g. `setup_22.x`) so it matches the project's Node version
- Fixes `npm: command not found` on bare Ubuntu sidecars during `chunk sidecar setup`

## Changes

- `envbuilder/envbuilder.go`: add `nodeInstallCmd(major)` helper and handle `"node"` inline in the system step loop using the already-detected `imageVersion`
- `envbuilder/envbuilder_test.go`: add `TestNodeInstallCmd` covering the version-aware path and fallback

## Test plan

- [x] `go test -race ./envbuilder/... -run 'TestNodeInstallCmd|TestDetectCommands'`
- [x] `chunk sidecar env` on a real TypeScript project shows a `system` step with the NodeSource install command
Screenshot 2026-06-16 at 11 58 14 AM

### DIFF
--- a/envbuilder/envbuilder.go
+++ b/envbuilder/envbuilder.go
@@ -390,9 +390,6 @@ func compareVersions(a, b string) int {
 func nodeInstallCmd(major string) string {
 	// Require a purely numeric major version so we never produce an invalid
 	// NodeSource URL (e.g. "unknown" from a failed version detection).
-	if major == "" || major == stackUnknown {
-		major = "22"
-	}
 	if _, err := strconv.Atoi(major); err != nil {
 		major = "22"
 	}
--- a/envbuilder/envbuilder.go
+++ b/envbuilder/envbuilder.go
@@ -388,7 +388,12 @@ func compareVersions(a, b string) int {
 }
 
 func nodeInstallCmd(major string) string {
+	// Require a purely numeric major version so we never produce an invalid
+	// NodeSource URL (e.g. "unknown" from a failed version detection).
 	if major == "" || major == stackUnknown {
+		major = "22"
+	}
+	if _, err := strconv.Atoi(major); err != nil {
 		major = "22"
 	}
 	return fmt.Sprintf(
@@ -1939,9 +1944,6 @@ func DetectEnvironment(ctx context.Context, dir string) (*Environment, error) {
 	for _, dep := range systemDeps {
 		var cmd string
 		if dep == "node" {
-			// Use the detected major version to pick the right NodeSource stream.
-			// sudo is already explicit in nodeInstallCmd, so we skip the cimg
-			// substitution below to avoid double-sudo on apt-get.
 			major := strings.SplitN(imageVersion, ".", 2)[0]
 			cmd = nodeInstallCmd(major)
 		} else if c, ok := extraDepInstalls[dep]; ok {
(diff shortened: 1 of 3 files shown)


internal/cmd/validate.go
```diff
@@ -218,10 +218,8 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 
 	// Validate --env flag syntax before any remote resolution so bad
 	// values are caught immediately regardless of execution mode.
-	if len(opts.envVarsFlag) > 0 {
-		if _, vErr := sidecar.ParseEnvPairs(opts.envVarsFlag); vErr != nil {
-			return &userError{msg: fmt.Sprintf("invalid --env value: %s", vErr), err: vErr}
-		}
+	if err := validateEnvFlag(opts.envVarsFlag); err != nil {
```
schurchleycci, line 219: **Bug:** `validateEnvFlag` runs before the `--dry-run` early-return (line 225), so `chunk validate --dry-run --env BADVALUE` returns an env-parse error instead of printing the would-be commands.

`opts.envVarsFlag` is never consumed in the dry-run path — `runValidateDryRun` takes only `name`/`inlineCmd`/`cfg`/`statusFn` — so this validation is irrelevant to dry-run but still gates it. Consider moving `validateEnvFlag` after the dry-run check, or skipping it when `opts.dryRun` is true.

internal/cmd/validate.go
```diff
@@ -288,6 +286,16 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 	return execErr
 }
 
+func validateEnvFlag(envVarsFlag []string) error {
```
schurchleycci, line 287: **Reuse:** `validateEnvFlag` calls `sidecar.ParseEnvPairs(envVarsFlag)` purely for validation and discards the result. Later, `resolveEnvVars` calls `sidecar.ParseEnvPairs` again on the same input as its first statement — so the same slice is parsed twice on every sidecar execution with `--env` flags.

Consider having `validateEnvFlag` return the parsed pairs so `resolveEnvVars` can accept pre-parsed input, or folding the validation into `resolveEnvVars` and eliminating the pre-flight call.

envbuilder/envbuilder.go
```diff
@@ -387,6 +387,21 @@ func compareVersions(a, b string) int {
 	return 0
 }
 
+func nodeInstallCmd(major string) string {
+	// Require a purely numeric major version so we never produce an invalid
+	// NodeSource URL (e.g. "unknown" from a failed version detection).
+	if major == "" || major == stackUnknown {
```
schurchleycci: **Simplification:** The first guard is dead code. `strconv.Atoi("")` and `strconv.Atoi("unknown")` both return errors, so the `Atoi` check on line 396 already catches every case this `if` handles. The block can be deleted with no behavior change:

```go
func nodeInstallCmd(major string) string {
    if _, err := strconv.Atoi(major); err != nil {
        major = "22"
    }
    return fmt.Sprintf(...)
}
```

Leaving dead guards forces readers to verify that `stackUnknown` is non-numeric before they can conclude the two branches are equivalent.

schurchleycci:
Added a few findings from Claude review!

schurchleycci (approved):
Thanks for addressing @hanabel1! ✅

PR #410, opened 2026-06-17
Route validate to active sidecar when sidecarImage is configured

hanabel1:
## Problem

From the friction log: after running `chunk sidecar snapshot create` and `chunk config set validation.sidecarImage`, users expected `chunk validate` to route to the sidecar. It didn't -`sidecarImage` alone only activated sidecar routing in Stop hook context, not for plain invocations.

## What changed

`validateNeedsSidecar` now returns true when `sidecarImage` is set **and** an active sidecar already exists, regardless of whether we're in a Stop hook. allRemote is set to true on `sidecarImage alone (hook guard removed).

Auto-create remains hook-only - `resolveSidecar` no longer auto-provisions a sandbox just because `image != ""`. Plain `chunk validate` with no active sidecar falls back to local with a warning.

Screenshot 2026-06-17 at 2 16 56 PM

Screenshot 2026-06-17 at 2 20 09 PM

### DIFF
--- a/acceptance/validate_test.go
+++ b/acceptance/validate_test.go
@@ -397,17 +397,16 @@ func TestValidateRunRemoteUsesSSH(t *testing.T) {
 	assert.Equal(t, len(execReqs), 0, "expected 0 HTTP exec requests (SSH should be used)")
 }
 
-func TestValidateAutoCreatesSidecar(t *testing.T) {
-	// Verify that chunk validate (no --remote) auto-creates a sidecar using the
-	// image stored in validation.sidecarImage when no active sidecar exists.
+func TestValidateSidecarImageNoActiveSidecarRunsLocally(t *testing.T) {
+	// Plain chunk validate with sidecarImage configured but no active sidecar must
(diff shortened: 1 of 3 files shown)


schurchleycci:
A couple Claude comments:

>  1. Stale godoc on resolveSidecar (validate.go:576-578)

>   The function comment still says:
> 
>   ▎ "auto-creates one when a sidecar image is configured or the caller is a Stop hook"
> 
>   But the function body now only auto-creates in the hook path (the image != "" branch was removed). The comment needs to be updated to reflect hook-only auto-create.
> 
>   2. sidecar.LoadActive called twice
> 
>   LoadActive is called at line 233 in runValidateCmdE, then called again inside resolveSidecar (line 582). The second call is redundant I/O. Since activeSidecar is already computed by the time resolveSidecar is invoked, it could be passed directly (similar to how hasActiveSidecar bool was added to setupRemote). Not a bug, but worth cleaning up.

schurchleycci (approved):

PR #428, opened 2026-07-10
Add `--project` flag to `chunk task config` for large CircleCI accounts

### LINKED ISSUE #427: `chunk task config` project picker does not scale for large CircleCI accounts
`chunk task config` is the only supported way to create `.chunk/run.json`, but its project-selection step does not scale for CircleCI accounts with many followed projects.

On a verified account with **438 followed projects**, the wizard loads every project into a single `SelectFromList` with no search, filter, or pagination. Navigating to the target repo with arrow keys is impractical.

## Current behavior

1. `chunk task config` calls `GET /api/v1.1/projects` and lists all followed projects alphabetically.
2. The user must scroll through the full list in a TTY (`↑/↓` only).
3. The fallback ("Enter manually") still requires looking up org and project UUIDs elsewhere.

There is no `--project` flag or other way to skip the picker when the project slug is already known (e.g. `gh/CircleCI-Public/chunk-cli`).

## Impact

- **Who:** Users on CircleCI accounts with hundreds or thousands of followed projects.
- **Workaround:** Hand-edit `.chunk/run.json`, or use "Enter manually" after looking up UUIDs in the web UI or API.
- **Cost:** First-time setup is blocked unless the user already knows how to bypass the wizard.

## Reproduction

```bash
chunk auth set circleci
chunk task config
# All followed projects appear in one flat list.
# On a large account, reaching the target project is impractical.
```

Verified on an account with 438 followed projects across 8 orgs. The count and UI implementation (flat list, no search) are sufficient to confirm the problem without completing the full interactive walkthrough.

## Expected behavior

Users should be able to specify the target project without scrolling through the full followed-projects list:

```bash
chunk task config --project gh/CircleCI-Public/chunk-cli
```

The wizard should resolve org and project IDs from the slug and continue to the remaining prompts.

A longer-term improvement would be search or org/repo filtering in the interactive picker, but a slug flag is the highest-leverage fix.

## Proposed fix

- [ ] Add `--project <slug>` to `chunk task config` to skip project selection
- [ ] Resolve org/project IDs via `GetProjectBySlug`
- [ ] Skip `ListFollowedProjects` when `--project` is set
- [ ] Document the flag in `docs/CLI.md`

hidetaka-cci:
- Add `--project <slug>` to `chunk task config` so users can skip the followed-projects picker when they already know the project slug (e.g. `gh/CircleCI-Public/chunk-cli`).
- When `--project` is set, skip `ListFollowedProjects` and `ListCollaborations`, and resolve org/project IDs via `GetProjectBySlug`.
- Fix `GetProjectBySlug` org ID parsing: the v2 API returns `organization_id`, not `org_id`. This was a latent bug surfaced while testing the new flag against the live API.
- Add a regression test with a live API-shaped JSON fixture so the `organization_id` field is not dropped on decode.

Closes #427

## Motivation

`chunk task config` lists every followed project in a single TTY picker with no search or filter. On an account with 438 followed projects, scrolling to the target repo is impractical. Users who already know the project slug — typically from the repo they are standing in — should be able to pass it directly.

Definition ID discovery is out of scope for this PR. The wizard still prompts for definition UUID and other fields interactively.

## Usage

```bash
chunk task config --project gh/CircleCI-Public/chunk-cli
```

Only the project-selection step is skipped. The wizard continues with definition name, definition UUID, default branch, and optional environment ID.

## Changes

| Area | Change |
|------|--------|
| `internal/cmd/task.go` | `--project` flag; skip project/collab fetch when set |
| `internal/task/collect.go` | `CollectOptions.ProjectSlug`; `resolveProjectFromSlug` |
| `internal/circleci/projects.go` | Fix `organization_id` JSON tag on `ProjectDetail` |
| `internal/circleci/circleci_test.go` | Regression test with live API JSON fixture |
| `internal/testing/fakes/circleci.go` | Fake `GET /api/v2/project/:slug` endpoint for tests |
| `docs/CLI.md` | Document `--project` flag |

## Bug fix note

While verifying `--project` against the live API, `GetProjectBySlug` returned an empty org ID. The API response uses `organization_id`, but `ProjectDetail` was tagged as `org_id`. This likely affected the interactive list-selection path as well, not only `--project`. The fix and regression test are included in this PR.

## Commits

1. Add `--project` flag to task config for large CircleCI accounts.
2. Add regression test for `GetProjectBySlug` `organization_id` parsing.

## Test plan

- [x] `go test -race ./internal/task/ ./internal/circleci/ ./internal/cmd/`
- [x] `TestGetProjectBySlug_decodesOrganizationIDFromAPIResponse` — raw JSON fixture matches live API shape; fails if `org_id` tag is used instead of `organization_id`
- [x] Sidecar: sync + remote tests for changed packages
- [x] Manual: `./dist/chunk task config --project gh/CircleCI-Public/chunk-cli` skips project list and proceeds to definition prompts
- [x] Live API: `GetProjectBySlug` returns org ID after `organization_id` fix

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -47,6 +47,7 @@ chunk
 │   │   --json                      # Output as JSON
 │   └── config                      # Set up .chunk/run.json for this repository
 │       --force                     # Overwrite existing configuration without confirmation
+│       --project <slug>            # Skip project picker (e.g. gh/org/repo)
 │
 ├── skill
 │   ├── install                     # Install all skills
--- a/internal/circleci/circleci_test.go
+++ b/internal/circleci/circleci_test.go
@@ -361,6 +361,30 @@ func TestTriggerRun(t *testing.T) {
 	})
 }
 
+func TestGetProjectBySlug(t *testing.T) {
+	fake := fakes.NewFakeCircleCI()
+	fake.ProjectDetails = map[string]fakes.ProjectDetail{
+		"gh/acme/api": {ID: "proj-1", Slug: "gh/acme/api", Name: "api", OrgID: "org-1"},
+	}
+	srv := httptest.NewServer(fake)
+	defer srv.Close()
+
+	client := newTestClient(t, srv.URL)
+	ctx := context.Background()
+
+	t.Run("success", func(t *testing.T) {
+		detail, err := client.GetProjectBySlug(ctx, "gh/acme/api")
+		assert.NilError(t, err)
+		assert.Equal(t, detail.ID, "proj-1")
+		assert.Equal(t, detail.OrgID, "org-1")
+	})
+
+	t.Run("not found", func(t *testing.T) {
+		_, err := client.GetProjectBySlug(ctx, "gh/missing/repo")
+		assert.Assert(t, err != nil)
+	})
+}
+
 func TestAuthRequired(t *testing.T) {
 	// Verify that the fake returns 401 when no Circle-Token header is present.
 	fake := fakes.NewFakeCircleCI()
--- a/internal/circleci/projects.go
+++ b/internal/circleci/projects.go
@@ -28,7 +28,7 @@ type ProjectDetail struct {
 	ID    string `json:"id"`
 	Slug  string `json:"slug"`
 	Name  string `json:"name"`
-	OrgID string `json:"org_id"`
+	OrgID string `json:"organization_id"`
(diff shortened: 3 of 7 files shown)


michael-webster:
We're currently not focused on the chunk cloud tasks. Closing PRs related to that feature.

PR #436, opened 2026-07-14
Add anonymous usage telemetry

hanabel1:
## Summary

Adds anonymous usage telemetry to chunk via Segment. All events are opt-out and no PII is collected.

### `command_invocation`

Fired after every CLI command completes.

| Property | Value | Notes |
|---|---|---|
| `userId` | stable UUID | Generated on first send, persisted as `instanceID` in `~/.config/chunk/config.json` |
| `command` | e.g. `chunk build-prompt` | Full command path; never includes flag values |
| `flags` | e.g. `model,output` | Sorted comma-separated list of flag *names* that were explicitly set; no values |

## What is NOT collected

- Flag values, argument values, file or directory names
- Tokens, API keys, or any credentials
- Error messages or stack traces
- Any personally identifiable information

## Opt-out

Telemetry is disabled by any of:

- `CHUNK_NO_TELEMETRY` environment variable (any non-empty value)
- `NO_ANALYTICS` environment variable set
- `DO_NOT_TRACK` environment variable set
- `CI` environment variable set (set by most CI systems)
- `chunk config set telemetry false` (persisted to `~/.config/chunk/config.json`)

## Implementation

- New `internal/telemetry` package: `Sender` fans events out to a `multiDestination` — a `loggingDestination` (stderr, gated behind `CHUNK_TELEMETRY_LOG`) and/or a `delegateDestination` (production)
- Events are buffered in-process and only flushed on `Sender.Close()`, which re-execs the `chunk` binary as a detached `receive-telemetry` subprocess (`internal/telemetry/delegate.go`, hidden command in `internal/cmd/receivetelemetry.go`) — this keeps Segment's network round trip off the CLI's exit path
- `internal/telemetry/receiver` decodes the JSON payload from stdin inside that subprocess and forwards it to Segment
- `instanceID` UUID is generated and persisted to `~/.config/chunk/config.json` on first send via `config.EnsureInstanceID`
- Every subcommand is instrumented automatically via `telemetry.RecordForSubcommands`, which wraps each `RunE` — no per-command changes needed. `telemetry.DisableTelemetry` opts a command out (used for `receive-telemetry` itself, so receiving telemetry never emits its own)
- The Segment write key (`cmd.writeKey`) is a hardcoded constant, matching circleci-cli's pattern — write keys are push-only, not secret, so checking one into git is safe. **It's currently `""`** pending Segment workspace access (see below); `setupTelemetry` never sends with an empty key, so this is a no-op until then, and dropping in the real key later is a one-line change.
- `setupTelemetry` also checks `testing.Testing()` before sending — see "Bug found during manual testing" below for why this matters independently of the write key being empty.

## Bug found during manual testing (fixed)

While manually verifying the delegate/receiver mechanism end-to-end against a fake local Segment server, I found that a non-empty write key causes a real problem under `go test`: `delegateDestination` re-execs `os.Executable()`, which under `go test` resolves to the compiled test binary (not `chunk`). Invoking it with `receive-telemetry` doesn't run that subcommand — it silently reruns the package's *entire, unfiltered* test suite, which re-triggers the same send from any test that exercises the real CLI pipeline (`hook_test.go`, `validate_test.go`). Each recursive rerun spawns more copies of itself. This produced **over 2000 orphaned background processes** on my machine before I caught and killed them, and is almost certainly the real cause of the `resource temporarily unavailable` test flakiness seen earlier on this branch (previously misattributed to environment noise).

Fixed with `testing.Testing()` in `setupTelemetry`, which blocks sending outright under `go test` — verified by forcing a fake key directly into the `go test` binary (both via `-ldflags` and by hardcoding it in source, matching how the real key will eventually be added) and confirming zero subprocesses spawn either way. This guard is independent of whether the key is build-injected or hardcoded, so it protects us permanently once a real key lands here.

## Test plan

- [x] `task build`
- [x] `go test ./internal/telemetry/... ./internal/config/... ./internal/cmd/...` — all pass, verified zero leaked `receive-telemetry` processes after
- [x] `task lint` — 0 issues
- [x] Built with a fake write key and confirmed `receive-telemetry` correctly posts a valid Segment `/v1/batch` payload to a local fake server
- [x] Hardcoded a fake key directly in source (simulating the real key landing here) and confirmed `testing.Testing()` still prevents any subprocess spawn under `go test`
- [x] `CHUNK_TELEMETRY_LOG=1` logs `[telemetry] track command_invocation` to stderr with correct properties
- [ ] Set `CHUNK_NO_TELEMETRY=1` and confirm no events are emitted (only verified via the `IsTelemetryEnabled` unit tests so far, not a manual run)
- [ ] Once a real write key is added: confirm events actually show up in Segment's debugger for a real command run

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -118,8 +118,12 @@ type UserConfig struct {
 	GitHubToken        string `json:"gitHubToken,omitempty"`
 	Model              string `json:"model,omitempty"`
 	UseSSHIdentityFile bool   `json:"useSSHIdentityFile,omitempty"`
-	NoTelemetry        bool   `json:"noTelemetry,omitempty"`
 	InstanceID         string `json:"instanceID,omitempty"`
+
+	// Telemetry is the persisted telemetry preference: true enables it,
+	// false disables it. nil means no preference has been set, in which
+	// case telemetry defaults to enabled (it is opt-out).
+	Telemetry *bool `json:"telemetry,omitempty"`
 
 	// LegacyAPIKey reads the pre-rename "apiKey" field so existing users don't
 	// silently lose their stored Anthropic key on upgrade. Migrated into
@@ -208,16 +212,20 @@ func Clear(key string) error {
 	return Save(cfg)
 }
 
-// IsTelemetryEnabled reports whether telemetry should be collected, honoring
-// (in order) well-known opt-out environment variables and the persisted
-// noTelemetry preference.
-func IsTelemetryEnabled(cfg UserConfig) bool {
+// IsTelemetry reports whether telemetry should be collected, honoring (in
+// order) well-known opt-out environment variables and the persisted
+// telemetry preference. Telemetry is opt-out: it defaults to enabled when no
+// preference has been set.
+func IsTelemetry(cfg UserConfig) bool {
 	for _, env := range noTelemetryEnvVars {
 		if os.Getenv(env) != "" {
 			return false
 		}
 	}
-	return !cfg.NoTelemetry
+	if cfg.Telemetry == nil {
+		return true
+	}
+	return *cfg.Telemetry
 }
 
 // EnsureInstanceID returns the persisted anonymous instance ID used to
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/sethvargo/go-envconfig"
 
 	"github.com/CircleCI-Public/chunk-cli/internal/keyring"
@@ -58,6 +59,7 @@ const (
 	EnvModel              = "CODE_REVIEW_CLI_MODEL"
 	EnvCircleCIOrgID      = "CIRCLECI_ORG_ID"
 	EnvChunkHooksDisabled = "CHUNK_HOOKS_DISABLED"
+	EnvChunkNoTelemetry   = "CHUNK_NO_TELEMETRY"
 )
 
 // System/standard environment variable names.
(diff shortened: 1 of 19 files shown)


internal/config/config.go
```diff
@@ -109,6 +118,8 @@ type UserConfig struct {
 	GitHubToken        string `json:"gitHubToken,omitempty"`
 	Model              string `json:"model,omitempty"`
 	UseSSHIdentityFile bool   `json:"useSSHIdentityFile,omitempty"`
+	NoTelemetry        bool   `json:"noTelemetry,omitempty"`
```
michael-webster: can we keep the conventions of the cli and just call this Telemetry. We can also add the `IsTelemetry` helper instead of `IsTelemetryEnabled`

internal/config/config.go
```diff
@@ -197,6 +208,40 @@ func Clear(key string) error {
...
+			return false
+		}
+	}
+	return !cfg.NoTelemetry
+}
+
+// EnsureInstanceID returns the persisted anonymous instance ID used to
+// associate telemetry events with a single install, generating and saving
+// one on first run.
+func EnsureInstanceID() (uuid.UUID, error) {
```
michael-webster, line 234: We'd probably want the OS and agent detection if we can add it reasonably. Happy to fast track that as a separate PR though.

michael-webster:
circleci-cli's delegate integration tests avoid the `os.Executable()` recursive-reexec problem structurally rather than with a runtime guard: `TestMain` in `internal/telemetry/sender_test.go` pre-builds a minimal stub binary from `internal/telemetry/testdata/receiverbin/main.go` (which contains only `receiver.Receive(os.Stdin)`) and passes that path as `Binary` when constructing a `Config{Send: true}` sender. The test binary is never involved.

Worth considering as a complement to the `testing.Testing()` guard here: a `testdata/receiverbin` stub + a `TestMain` that builds it would let us write a real end-to-end delegate test (spawn subprocess → assert Segment payload) without any risk of recursive spawning, since the stub has no subcommands at all. The `testing.Testing()` guard would still be a useful safety net for anyone who writes a test that accidentally passes `Send: true` with a real key, but the stub approach makes that class of test safe to write intentionally.

michael-webster (approved):
Approved mod changing the telemetry to be `Telemetry` with true meaning telemetry is enabled and false disabled, instead of the `NoTelemetry` version as well as using a stubbed receiver instead of the testing guard.

PR #438, opened 2026-07-14
Promote env detection to chunk env init, remove sidecar env/build

schurchleycci:
## Summary
The thinking here: if we want to use dockerfiles as the basis for spinning up lots of sidecars, having a command to generate the dockerfile directly, rather than as a side effect of `chunk sidecar env | chunk sidecar build --dir`, may be a smoother experience.

- Adds `chunk env init` as a new top-level command that detects a repository's tech stack and writes `Dockerfile.test` to `--dir` (default), or prints the environment spec as JSON (`--format json`). The detected spec is saved to `.chunk/config.json` for reuse by `chunk sidecar setup`.
- Fixes a stale error message in `validate.go` that referenced `chunk sidecar env build` (a command that never existed); it now correctly suggests `chunk sidecar sync`.

## Before / after

```bash
# Before
chunk sidecar env | chunk sidecar build --dir .

# After
chunk env init --dir .
docker build -f Dockerfile.test -t myapp:test .
```

## Test plan

- [ ] `task test` passes (acceptance + unit)
- [ ] `chunk env init` writes `Dockerfile.test` and prints its path
- [ ] `chunk env init --format json` prints the env spec to stdout
- [ ] `chunk env init --no-save` skips writing `.chunk/config.json`
- [ ] `chunk sidecar setup` still works (reuses cached spec from config)

### DIFF
--- a/harness/environment.py
+++ b/harness/environment.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-environment: harness for `chunk sidecar env` + the sidecar build acceptance test.
+environment: harness for `chunk env detect` + the sidecar build acceptance test.
 Runs against target repos and uses the Claude Code SDK to improve
 internal/envbuilder when tests fail.
 
@@ -176,7 +176,7 @@ def _run_test(repo: TargetRepo, cache_dir: Path, timeout: int, extra_env: dict |
     }
     result = subprocess.run(
         ["go", "test", "-v", "-count=1", f"-timeout={timeout}s",
-         "-run", "TestSidecarsBuildEndToEnd", "./acceptance/"],
+         "-run", "TestEnvDetectEndToEnd", "./acceptance/"],
         cwd=REPO_ROOT,
         capture_output=True,
         text=True,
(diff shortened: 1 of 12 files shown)


internal/cmd/sidecar.go
```diff
@@ -572,153 +566,6 @@ func newSidecarForgetCmd() *cobra.Command {
...
-	cmd := &cobra.Command{
-		Use:   "build",
-		Short: "Generate a Dockerfile from an environment spec and build a test image",
-		Long: `Read a JSON environment spec from stdin (produced by 'chunk sidecar env'),
-write Dockerfile.test to --dir, and build a Docker test image from it.
-
-Example:
-  chunk sidecar env --dir . | chunk sidecar build --dir .`,
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			if tag != "" && !validDockerTag.MatchString(tag) {
```
hanabel1, line 649: we're dropping chunk's tag validation here now that docker build is run by the user instead of by chunk. it'll still reject a bad tag, with its own error - but wasn't sure if it was intentional - just a ux thing i guess!

harness/environment.py
```diff
@@ -277,7 +277,7 @@ def update_envbuilder(
         else:
             prompt = f"""You are debugging an environment detection tool inside the chunk CLI.
 
-`chunk sidecar env` analyses a repository, detects its tech stack, and writes a
+`chunk env detect` analyses a repository, detects its tech stack, and writes a
```
hanabel1: err i might be missing this but i can not find the detect subcommand. naming mismatch?

internal/cmd/sidecar.go
schurchleycci: Yeah, this was intentional - under the assumption we may as well just have the user run docker build themselves

harness/environment.py
schurchleycci: Good catch, this was a naming mismatch - fixed now!

hanabel1 (approved):

schurchleycci:
Closing this for now, can revisit later if need be

PR #439, opened 2026-07-15
Collect OS and detected AI coding agent in telemetry

hanabel1:
## Summary

Follow-up to #436: adds two fields to every telemetry event's `Context` —
the operating system and, if detected, the AI coding agent chunk-cli was
invoked from.

## Events

Existing `command_invocation` event context gains:

| Field | Value | Notes |
|---|---|---|
| `Context.OS.Name` | e.g. `darwin` | `runtime.GOOS` |
| `Context.Extra["codingAgent"]` | e.g. `claude-code`, `cursor` | Omitted entirely when no known agent is detected |

## What is NOT collected

- Any new PII — detection only checks for the *presence* of well-known
  environment variables (`CLAUDECODE`, `CURSOR_TRACE_ID`), never their values

## Implementation

- New `internal/telemetry/agent.go`: `DetectCodingAgent` checks a small,
  intentionally conservative list of environment variables confirmed to be
  set by the agent itself (Claude Code's `CLAUDECODE`, Cursor's
  `CURSOR_TRACE_ID`) and returns a stable agent name, or `""` if none match
  — the common case for a human running `chunk` directly
- `Meta` gains `OS` and `CodingAgent` fields; `toContext` sets
  `analytics.Context.OS.Name` and only populates `Context.Extra["codingAgent"]`
  when an agent was detected
- `internal/cmd/root.go`'s `setupTelemetry` populates both from
  `runtime.GOOS` and `telemetry.DetectCodingAgent()`
- Documented in `docs/ARCHITECTURE.md`

## Test plan

- [x] `task build`
- [x] `task test` (`-race`) — all pass
- [x] `task lint` — 0 issues
- [x] New unit tests for `DetectCodingAgent` (no signal, each known agent,
      first-match-wins when multiple are set) and for `Meta.toContext`
      omitting `codingAgent` when undetected

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -216,6 +216,15 @@ package uses, because both tools currently share the same Segment write
 key/workspace; the `chunk_` prefix keeps chunk-cli's events unambiguous in
 the event stream.
 
+Every event's `Context` also carries the operating system (`runtime.GOOS`)
+and, if detected, the AI coding agent chunk-cli was invoked from (e.g.
+`claude-code`, `cursor`) — see `internal/telemetry/agent.go`'s
+`DetectCodingAgent`, which checks for well-known environment variables those
+agents set in the shells they spawn. The agent list is intentionally
+conservative: only signals confirmed to be set by the agent itself are
+included, to avoid misattributing invocations. Detection returns `""` when no
+known agent is found, the common case for a human running `chunk` directly.
+
 - `internal/cmd/root.go`'s `setupTelemetry` resolves the user's preference
   (opt-out env var → `telemetry` config field, first match wins) and
   attaches a `telemetry.Sender` to the command's context;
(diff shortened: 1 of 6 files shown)


schurchleycci (approved):
A couple small things from Claude:

> 1. envClaudeCode = "CLAUDECODE" — worth a comment citing the source
> 
> The env var name CLAUDECODE is a specific, verifiable contract. It would be easy for someone to accidentally "confirm" a signal that isn't reliably set, so it's worth a short comment (e.g. // set by Claude Code since v...) alongside the constant, the same way the CURSOR_TRACE_ID one might warrant one. Low priority but useful for future maintainers extending the list.
> 
> 2. agentEnvSignals is package-level mutable state
> 
> It's a var, not a const or anything guarded. Nothing currently mutates it, but if a future test tried to append to it, it would leak across tests. Switching to a function returning a slice, or a var named with a comment noting it should be treated as read-only, would make the intent explicit. Again, low priority given the current codebase.

PR #441, opened 2026-07-17
Add auth signup and org create commands

schurchleycci:
## Summary

- Adds `chunk auth signup` to open the CircleCI signup flow via browser OAuth (mirrors `chunk auth login` but routes to the signup page)
- Adds `chunk org create <name>` to create a new standalone CircleCI organization via `POST /api/v2/organization`

## Test plan

- [ ] `chunk auth signup` opens browser to CircleCI signup and saves token on completion
- [ ] `chunk auth signup` with an existing token surfaces a clear error with remediation hint
- [ ] `chunk org create my-org` creates a standalone org and prints the ID and slug
- [ ] `chunk org create` without a name exits with a usage error
- [ ] Both commands respect `--insecure-storage` and `CI` (non-interactive) mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,10 @@ Chunk init will install skills for working with Chunk sidecars. After the init, 
 Create and work in cloud sidecar environments. Sidecars are available to all CircleCI customers, including the free plans. Share feedback in the [CircleCI Discord](https://discord.gg/circleci).
 
 ```bash
-# Authenticate
+# Authenticate (browser OAuth — recommended)
+chunk auth login
+
+# Or set a token directly
 chunk auth set circleci
 
 # Create a sidecar (sets it as active automatically)
@@ -120,7 +123,8 @@ chunk skill install
 ## Commands
 
 ```
-chunk auth set|status|remove               Authentication
+chunk auth login|signup|set|status|remove  Authentication
+chunk org create                           Manage CircleCI organizations
 chunk sidecar list|create|exec|ssh         Manage cloud sidecar environments
 chunk sidecar sync|env|build               Sync files, detect env, build images
 chunk sidecar use|current|forget           Manage active sidecar
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Create and work in cloud sidecar environments. Sidecars are available to all Cir
 # Authenticate (browser OAuth — recommended)
 To create a new account:
 chunk auth signup
-or , to log in to an existing account:
+or, to log in to an existing account:
 chunk auth login
 
 # Or set a token directly
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ Create and work in cloud sidecar environments. Sidecars are available to all Cir
 
 ```bash
 # Authenticate (browser OAuth — recommended)
+To create a new account:
+chunk auth signup
+or , to log in to an existing account:
 chunk auth login
 
 # Or set a token directly
--- a/internal/circleci/orgs_test.go
+++ b/internal/circleci/orgs_test.go
@@ -22,9 +22,9 @@ func TestCreateOrg(t *testing.T) {
 
 		org, err := client.CreateOrg(ctx, "my-org")
 		assert.NilError(t, err)
-		assert.Assert(t, org.ID != "", "expected non-empty org ID")
+		assert.Equal(t, org.ID, "org-new-1")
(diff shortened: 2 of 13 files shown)


internal/circleci/projects.go
```diff
@@ -67,3 +67,25 @@ func (c *Client) GetProjectBySlug(ctx context.Context, slug string) (*ProjectDet
...
+type OrgInfo struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	Slug    string `json:"slug"`
+	VcsType string `json:"vcs_type"`
+}
+
+// CreateOrg creates a new CircleCI organization. vcsType must be one of
+// "circleci" (standalone), "github", or "bitbucket".
+func (c *Client) CreateOrg(ctx context.Context, name, vcsType string) (*OrgInfo, error) {
```
michael-webster: this probably shouldn't go into `project.go`

internal/cmd/org.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/config"
+	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
+	"github.com/CircleCI-Public/chunk-cli/internal/tui"
+	"github.com/CircleCI-Public/chunk-cli/internal/ui"
+)
+
+func newOrgCmd() *cobra.Command {
```
michael-webster, line 14: might be worth making this be a hidden command since we really only have this for onboarding purposes.

internal/circleci/projects.go
michael-webster: since we only intend this to be for standalone i would just make the vcs type be circleci.

schurchleycci: Good call, I've moved it into a new `orgs.go`

internal/cmd/org.go
schurchleycci: Done!

internal/circleci/projects.go
schurchleycci: Good catch, updated this

README.md
```diff
@@ -50,7 +50,10 @@ Chunk init will install skills for working with Chunk sidecars. After the init,
 Create and work in cloud sidecar environments. Sidecars are available to all CircleCI customers, including the free plans. Share feedback in the [CircleCI Discord](https://discord.gg/circleci).
 
 ```bash
-# Authenticate
+# Authenticate (browser OAuth — recommended)
+chunk auth login
```
michael-webster, line 57:
```suggestion
To create a new account:
chunk auth signup
or , to log in to an existing account:
chunk auth login
```
add that as well

internal/cmd/org_test.go
```diff
@@ -0,0 +1,104 @@
...
+	var ue *userError
+	assert.Assert(t, errors.As(err, &ue))
+	assert.Assert(t, strings.Contains(ue.UserMessage(), "failing-org"),
+		"expected org name in error message, got: %s", ue.UserMessage())
+}
+
+func TestOrgCreateRequiresAuth(t *testing.T) {
+	isolateConfig(t)
+	t.Setenv(config.EnvCircleToken, "")
+	t.Setenv(config.EnvCircleCIToken, "")
```
michael-webster, line 77: this is because we support both?

internal/circleci/orgs_test.go
```diff
@@ -0,0 +1,79 @@
...
+		defer srv.Close()
+
+		client := newTestClient(t, srv.URL)
+		ctx := context.Background()
+
+		org, err := client.CreateOrg(ctx, "my-org")
+		assert.NilError(t, err)
+		assert.Assert(t, org.ID != "", "expected non-empty org ID")
+		assert.Equal(t, org.Name, "my-org")
+		assert.Assert(t, org.Slug != "", "expected non-empty org slug")
```
michael-webster: let's change this and the org.ID assert to have real values instead of just not empty.

michael-webster (approved):
A few smaller nits on the testing structure and a code suggestion you can apply to recommend signup and login.

Approving as these can be done whenever. Overall structure looks fine.

internal/cmd/org_test.go
schurchleycci: Yup, that's correct

PR #442, opened 2026-07-17
Add bundle-based sidecar sync as the default

michael-webster:
## Summary

Revives and extends #292.

`chunk sidecar sync` now uses git bundle by default, removing the requirement to push the branch to GitHub before syncing:

- **First sync**: sends a full bundle of HEAD to the sidecar, initialising a bare repo there if absent
- **Subsequent syncs**: sends only `<lastRef>..HEAD` (incremental), skipping the bundle entirely when HEAD has not advanced
- **Uncommitted changes**: applied on top as a patch after each bundle transfer
- **Opt out**: pass `--checkout` to fall back to the previous checkout/patch approach (requires branch pushed to GitHub)

Both `chunk sidecar sync` and `chunk sidecar setup` use bundle sync by default. The `bundleSync` config field has been removed since bundle sync is now the default.

Also included: stdin piping through `chunk sidecar ssh` — when the process stdin is a pipe it is forwarded to the remote command, enabling patterns like `cat bundle | chunk sidecar ssh -- git fetch ...`.

## Test plan

- [x] `chunk sidecar sync` on a fresh sidecar performs a full bundle sync
- [x] Re-running `chunk sidecar sync` skips the bundle and only applies the working-tree patch
- [x] `chunk sidecar sync --checkout` falls back to the checkout/patch approach
- [ ] `chunk validate test --remote` passes after a bundle sync
- [ ] `cat bundle | chunk sidecar ssh -- git fetch ...` works (stdin piping)

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -88,7 +88,7 @@ chunk
 │   │   --sidecar-id <id>           # Sidecar ID (defaults to active sidecar)
 │   │   --public-key <key>          # SSH public key string
 │   │   --public-key-file <path>    # Path to public key file
-│   ├── ssh                         # SSH into sidecar
+│   ├── ssh                         # SSH into sidecar (stdin forwarded when piped)
 │   │   --sidecar-id <id>           # Sidecar ID (defaults to active sidecar)
 │   │   --identity-file <path>      # SSH identity file
 │   │   -e / --env KEY=VALUE        # Set env var in remote session (repeatable)
@@ -97,6 +97,7 @@ chunk
 │   │   --sidecar-id <id>           # Sidecar ID (defaults to active sidecar)
 │   │   --identity-file <path>      # SSH identity file
 │   │   --workdir <path>            # Destination path on sidecar (auto-detected when omitted)
+│   │   --checkout                  # Sync via git checkout/patch instead of bundle (requires branch pushed to GitHub)
 │   ├── env                         # Detect tech stack and print environment spec as JSON
 │   │   --dir <path>                # Directory to analyse (default: .)
 │   │   --no-save                   # Print only, do not save to .chunk/config.json
@@ -154,6 +155,12 @@ chunk
 - `chunk init` uses Claude to auto-detect the test command for the project.
   It generates `.claude/settings.json` with pre-commit hooks. It never touches
   CircleCI — tokens are prompted inline only when a command actually needs them.
+- `sidecar sync` sends a full git bundle on first use, then incremental bundles
+  (`<lastRef>..HEAD`) on subsequent syncs. The branch does not need to be pushed
+  to GitHub. Pass `--checkout` to fall back to the git checkout/patch approach
+  (requires the branch to be pushed).
+- `sidecar ssh -- <cmd>` forwards stdin when the process stdin is a pipe, enabling
+  patterns like `cat bundle | chunk sidecar ssh -- git fetch ...`.
 - Commands that require a CircleCI token (`task run`, `task config`, `sidecar *`,
   `validate --sidecar-id`) prompt for it inline at the point of need rather than
   failing with an error.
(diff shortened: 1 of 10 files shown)


docs/CLI.md
```diff
@@ -143,20 +143,22 @@ chunk
...
   Non-interactive sessions (agents, CI) should set `orgID` in project config or
   pass `--org-id` / `CIRCLECI_ORG_ID`.
 - `chunk init` uses Claude to auto-detect the test command for the project.
+- `sidecar sync` sends a full git bundle on first use, then incremental bundles
+  (`<lastRef>..HEAD`) on subsequent syncs. The branch does not need to be pushed
+  to GitHub. Pass `--checkout` to fall back to the git checkout/patch approach
+  (requires the branch to be pushed).
+- `sidecar ssh -- <cmd>` forwards stdin when the process stdin is a pipe, enabling
+  patterns like `cat bundle | chunk sidecar ssh -- git fetch ...`.
   It generates `.claude/settings.json` with pre-commit hooks. It never touches
```
schurchleycci: I think 162-163 are supposed to be directly below `chunk init`

docs/CLI.md
```diff
@@ -98,6 +97,7 @@ chunk
 │   │   --sidecar-id <id>           # Sidecar ID (defaults to active sidecar)
 │   │   --identity-file <path>      # SSH identity file
 │   │   --workdir <path>            # Destination path on sidecar (auto-detected when omitted)
+│   │   --bundle                    # Sync via git bundle (no GitHub access required)
```
schurchleycci: This should be --checkout now that the default behaviour is swapped, right?

internal/sidecar/sync.go
```diff
@@ -116,6 +117,151 @@ func Sync(ctx context.Context,
...
+	client *circleci.Client, sidecarID, identityFile, authSock, workdir, cwd string, status iostream.StatusFunc) error {
+
+	session, err := OpenSession(ctx, client, sidecarID, identityFile, authSock)
+	if err != nil {
+		return err
+	}
+
+	_, repo, err := gitremote.DetectOrgAndRepo(cwd)
+	if err != nil {
+		return fmt.Errorf("bundle sync: %w", err)
```
schurchleycci: BundleSync doesn't wrap NoOriginRemoteError (correctness)

Sync wraps the DetectOrgAndRepo error in &NoOriginRemoteError{}, which the command layer catches to show a user-friendly message. BundleSync wraps it as a plain fmt.Errorf instead, so that check silently falls through to a generic error. sync.go:134:

_, repo, err := gitremote.DetectOrgAndRepo(cwd)
if err != nil {
    return fmt.Errorf("bundle sync: %w", err)  // should be &NoOriginRemoteError{Err: err}
}

internal/sidecar/sync.go
```diff
@@ -116,6 +117,151 @@ func Sync(ctx context.Context,
...
+	}
+
+	active, err := LoadActive(ctx)
+	if err != nil {
+		return fmt.Errorf("bundle sync: load active sidecar: %w", err)
+	}
+	if active == nil {
+		active = &ActiveSidecar{SidecarID: sidecarID}
+	}
+	lastRef := active.LastSyncedRef
```
schurchleycci, line 154: When active.SidecarID != sidecarID, the code forces a full bundle by setting lastRef = "" — correct. But it never updates active.SidecarID = sidecarID before calling SaveActive. So the saved file still has the old sidecar ID, and every subsequent sync will detect a mismatch and force a full bundle forever. sync.go:154–157:

lastRef := active.LastSyncedRef
if active.SidecarID != sidecarID {
    lastRef = "" // sidecar changed; force full bundle
    // missing: active.SidecarID = sidecarID
}

schurchleycci (approved):

PR #446, opened 2026-07-22
Add chunk watch TUI dashboard with sidecar activity log

hanabel1:
## Summary

Adds `chunk watch`, a live TUI dashboard showing active sidecars, sync state, and recent activity in a split-pane layout. 
- `--all` flag watches all known projects, not just the current direcotyr
- Sidecars are reused across sessions instead of spawning a new one per Claude Code session
- Adds an append only JSONL event log (internal/eventlog) written by chunk sidecar sync and chunk validate to feed the TUI
- Adds iostreamLevelError and emits done/error completion events from validate so the spinner clears correctly. 

https://github.com/user-attachments/assets/60134a5e-8b99-4944-aecd-2eaf3b428ff5

## Implementation

**New packages:**
- `internal/eventlog` — append-only JSONL log; `Log.Wrap` decorates any `StatusFunc` to mirror calls into the log tagged with sidecar ID, op, and branch
- `internal/tui/watch` — BubbleTea model: left pane lists sidecars with sync/running state, right pane shows events filtered to the selected sidecar grouped by op-run; polls every 2s and tails the log incrementally

**Wiring changes:**
- `internal/cmd/sidecar.go` — wraps `syncFn` with the event log before passing to `BundleSync`/`Sync`
- `internal/cmd/validate.go` — wraps `statusFn` with the event log before `runValidate`; emits `LevelDone`/`LevelError` after the run so the TUI knows when validation is complete
- `internal/cmd/root.go` — registers `newWatchCmd()`
- `internal/iostream/status.go` — adds `LevelError` constant
- `internal/ui/format.go` — adds `ErrFailure` for red stderr rendering
- `go.mod` — promotes `charm.land/lipgloss/v2` from indirect to direct dependency

## Test plan

- [ ] `task build && ./dist/chunk watch` — opens TUI; shows sidecars, updates every 2s
- [ ] `./dist/chunk sidecar sync` — sync events appear in right pane for the active sidecar
- [ ] `./dist/chunk validate` — validate events appear; spinner clears after `✓ validate passed`
- [ ] Validate failure — spinner clears with `✗ validate failed` in red
- [ ] `j`/`k` — navigates sidecars; right pane updates to show selected sidecar's events
- [ ] `q` / `Ctrl-C` — exits cleanly
- [ ] `task test` — all tests pass

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -16,6 +16,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/envbuilder"
 	"github.com/CircleCI-Public/chunk-cli/internal/circleci"
 	"github.com/CircleCI-Public/chunk-cli/internal/config"
+	"github.com/CircleCI-Public/chunk-cli/internal/eventlog"
 	"github.com/CircleCI-Public/chunk-cli/internal/gitutil"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
@@ -495,11 +496,19 @@ func newSidecarSyncCmd() *cobra.Command {
 			if cwdErr != nil {
 				return fmt.Errorf("sync: %w", cwdErr)
 			}
+			syncFn := newStatusFunc(io)
+			if dataDir, dirErr := sidecar.StateDir(); dirErr == nil {
+				scName := ""
+				if as, _ := sidecar.LoadActive(cmd.Context()); as != nil && as.SidecarID == sidecarID {
+					scName = as.Name
+				}
+				syncFn = eventlog.WrapFromDir(dataDir, syncFn, eventlog.OpSync, sidecarID, scName, sidecar.CurrentBranch(cwd))
+			}
 			useBundle := !checkout
 			if useBundle {
-				err = sidecar.BundleSync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, cwd, newStatusFunc(io))
+				err = sidecar.BundleSync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, cwd, syncFn)
 			} else {
-				err = sidecar.Sync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, newStatusFunc(io))
+				err = sidecar.Sync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, syncFn)
 			}
 			if err != nil {
 				if _, ok := errors.AsType[*sidecar.NoOriginRemoteError](err); ok {
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -288,56 +288,75 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 
 	statusFn(iostream.LevelStep, "chunk validate")
 
-	// Wire event log around the actual validate run (best-effort; never blocks on failure).
-	if dataDir, dirErr := sidecar.StateDir(); dirErr == nil {
-		scName := ""
-		if activeSidecar != nil && activeSidecar.SidecarID == opts.sidecarID {
-			scName = activeSidecar.Name
-		}
-		statusFn = eventlog.WrapFromDir(dataDir, statusFn, eventlog.OpValidate, opts.sidecarID, scName, sidecar.CurrentBranch(workDir))
+	freshlyCreated, err := setupRemote(ctx, circleCIClient, opts, image, cfg, activeSidecar, statusFn, workDir, streams)
+	if err != nil {
+		return err
 	}
 
-	execErr := func() error {
-		freshlyCreated, err := setupRemote(ctx, circleCIClient, opts, image, cfg, activeSidecar, statusFn, workDir, streams)
-		if err != nil {
-			return err
-		}
-		// Only load env vars and resolve secrets when a sidecar is actually
-		// being used — avoids parsing .env.local or hitting secrets APIs on
-		// purely local runs.
-		envVars, err := loadSidecarEnvVars(ctx, circleCIClient, opts, workDir, statusFn)
-		if err != nil {
-			return err
-		}
-		return runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
-	}()
+	// Wire event log only when a sidecar is involved. The wrap goes here — after
+	// setupRemote fills opts.sidecarID but before loadSidecarEnvVars — so that
+	// sync and env-resolve status events are captured. Skipping when there is no
+	// sidecar avoids writing events with an empty sidecar_id that the TUI would
+	// filter out and never display.
+	statusFn = wrapEventLogStatusFn(statusFn, opts.sidecarID, activeSidecar, workDir)
+
+	// Only load env vars and resolve secrets when a sidecar is actually
+	// being used — avoids parsing .env.local or hitting secrets APIs on
+	// purely local runs.
+	envVars, err := loadSidecarEnvVars(ctx, circleCIClient, opts, workDir, statusFn)
+	if err != nil {
+		return err
+	}
+
+	execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
+	return finishValidate(cmd, hook, execErr, start, opts.sidecarID, cfg, statusFn, streams)
+}
+
+// wrapEventLogStatusFn wraps statusFn with event log recording when a sidecar
+// is active. Returns statusFn unchanged when no sidecar is involved, so callers
+// with empty sidecar IDs never write events with a blank sidecar_id.
+func wrapEventLogStatusFn(statusFn iostream.StatusFunc, sidecarID string, activeSidecar *sidecar.ActiveSidecar, workDir string) iostream.StatusFunc {
+	if sidecarID == "" {
+		return statusFn
+	}
+	dataDir, err := sidecar.StateDir()
+	if err != nil {
+		return statusFn
+	}
+	scName := ""
+	if activeSidecar != nil && activeSidecar.SidecarID == sidecarID {
+		scName = activeSidecar.Name
+	}
+	return eventlog.WrapFromDir(dataDir, statusFn, eventlog.OpValidate, sidecarID, scName, sidecar.CurrentBranch(workDir))
+}
+
+// finishValidate reports the validate outcome and handles hook exit codes.
+func finishValidate(cmd *cobra.Command, hook *hookContext, execErr error, start time.Time, sidecarID string, cfg *config.ProjectConfig, statusFn iostream.StatusFunc, streams iostream.Streams) error {
 	if execErr != nil {
 		statusFn(iostream.LevelError, fmt.Sprintf("done in %s (failed)", ui.FormatDuration(time.Since(start))))
 	} else if hook == nil {
 		statusFn(iostream.LevelStep, fmt.Sprintf("done in %s", ui.FormatDuration(time.Since(start))))
 	}
-
-	if opts.sidecarID != "" {
+	if sidecarID != "" {
 		if execErr != nil {
 			statusFn(iostream.LevelError, "validate failed")
 		} else {
 			statusFn(iostream.LevelDone, "validate passed")
 		}
 	}
-
-	if hook != nil {
-		maxAttempts := cfg.StopHookMaxAttempts
-		if maxAttempts <= 0 {
-			maxAttempts = validate.DefaultMaxAttempts
-		}
-		hookErr := validate.WrapHookResult(hook.sessionID, execErr, maxAttempts, streams.Err)
-		if hookErr == nil && execErr == nil {
-			_, _ = fmt.Fprintf(cmd.OutOrStdout(), "%s\n", ui.Success(fmt.Sprintf("chunk validate passed (%s)", ui.FormatDuration(time.Since(start)))))
-			return nil
-		}
-		return hookErr
+	if hook == nil {
+		return execErr
 	}
-	return execErr
+	maxAttempts := cfg.StopHookMaxAttempts
+	if maxAttempts <= 0 {
+		maxAttempts = validate.DefaultMaxAttempts
+	}
+	hookErr := validate.WrapHookResult(hook.sessionID, execErr, maxAttempts, streams.Err)
+	if hookErr == nil && execErr == nil {
+		_, _ = fmt.Fprintf(cmd.OutOrStdout(), "%s\n", ui.Success(fmt.Sprintf("chunk validate passed (%s)", ui.FormatDuration(time.Since(start)))))
+		return nil
+	}
+	return hookErr
 }
 
 func validateEnvFlag(envVarsFlag []string) error {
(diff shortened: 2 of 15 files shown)


michael-webster:
🔥

internal/tui/watch/model.go
```diff
@@ -0,0 +1,643 @@
...
+	"███   ██         ██   ██████",
+	"███     █████████     ██████",
+	logoBlankRow,
+	"███                   ████",
+	"  ███                ██",
+	"    █████████████████",
+}
+
+// color helpers — always enabled (watch requires a TTY)
+func fg(code, s string) string { return "\x1b[" + code + "m" + s + "\x1b[0m" }
```
michael-webster: can we use lipgloss instead of defining raw escape codes?

internal/cmd/sidecar.go
```diff
@@ -478,11 +479,21 @@ func newSidecarSyncCmd() *cobra.Command {
 			if cwdErr != nil {
 				return fmt.Errorf("sync: %w", cwdErr)
 			}
+			syncFn := newStatusFunc(io)
+			if dataDir, dirErr := sidecar.StateDir(); dirErr == nil {
```
michael-webster, line 500: we duplicate this a few times, can you create a helper in the event log that closes over all the work to figure out the file path and handle any errors loading the sidecar

internal/tui/watch/model.go
```diff
@@ -0,0 +1,643 @@
...
+		snap := hasSnapshotFile(p.DataDir)
+		sidecars := loadSidecars(p.DataDir, p.ProjectRoot, snap)
+		allSidecars = append(allSidecars, sidecars...)
+
+		if p.Log == nil {
+			continue
+		}
+		offset := m.offsets[i]
+		if offset == 0 {
+			recent, _ := p.Log.Recent(recentEvents)
```
michael-webster: do we need to handle these errors at all?

internal/tui/watch/model.go
```diff
@@ -0,0 +1,643 @@
...
+		}
+	}
+
+	return dataMsg{sidecars: allSidecars, events: allEvents, offsets: newOffsets}
+}
+
+// loadSidecars reads all sidecar*.json files from dataDir, deduplicates by ID.
+func loadSidecars(dataDir, projectRoot string, snap bool) []sidecarInfo {
+	matches, _ := filepath.Glob(filepath.Join(dataDir, "sidecar*.json"))
+	head := headRef(projectRoot)
```
michael-webster: from claude review: headRef calls exec.Command("git", "-C", dir, "rev-parse", "HEAD") inside loadData, which fires every poll tick. For N projects this spawns N git subprocesses every 2 seconds
  indefinitely. Should headRef be called once per poll at the cmd layer and passed in as state, rather than re-invoked inside the model?

seems like we don't need to poll quite so aggressively here?

internal/sidecar/active.go
```diff
@@ -126,7 +168,51 @@ func SaveActiveTo(ctx context.Context, dir string, a ActiveSidecar) error {
 	}
 	root, _ := projectRoot()
 	branch := CurrentBranch(root)
-	return os.WriteFile(filepath.Join(dir, sidecarFileName(session.IDFromCtx(ctx), branch)), data, 0o644)
+	if err := os.WriteFile(filepath.Join(dir, sidecarFileName(session.IDFromCtx(ctx), branch)), data, 0o644); err != nil {
+		return err
```
michael-webster, line 175: Could `saveDir()` and `config.ProjectDataDir()` produce different subdirectory paths for the same project root? Both write a `project-root` breadcrumb, but `AllProjectRoots` discovers projects by iterating `config.AppData()` subdirs. If the two functions hash or clean the project path differently, projects registered via `watch.go` (which calls `config.ProjectDataDir(abs)` directly) would never appear in `--all` output — silently producing an empty list with no error.

michael-webster (changes requested):

internal/cmd/sidecar.go
hanabel1: Yes, added [here](https://github.com/CircleCI-Public/chunk-cli/pull/446/changes#diff-32b406521e8d8262bc359d2f67a990c7868df17173fbc0e5497dc096c2a9e9c4R88).

internal/sidecar/active.go
hanabel1: Fixed [here](https://github.com/CircleCI-Public/chunk-cli/pull/446/changes#diff-28e29eebdcfb06bc9b6d01c91e25190499eaeb45433f9185a94ec2229b259cbaR59) - ProjectDataDir now runs EvalSymlinks before hashing so both code paths always land in the same directory.

internal/tui/watch/model.go
hanabel1: err they're intentionally ignored - log reads are best effort and a missing file shouldn't crash the tui

internal/cmd/validate.go
```diff
@@ -281,8 +284,25 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return err
 	}
 
+	// Wire event log around the actual validate run (best-effort; never blocks on failure).
+	if dataDir, dirErr := sidecar.StateDir(); dirErr == nil {
+		scName := ""
+		if activeSidecar != nil && activeSidecar.SidecarID == opts.sidecarID {
+			scName = activeSidecar.Name
+		}
+		statusFn = eventlog.WrapFromDir(dataDir, statusFn, eventlog.OpValidate, opts.sidecarID, scName, sidecar.CurrentBranch(workDir))
```
danmux: The wrap is installed *after* `setupRemote` and `loadSidecarEnvVars`, so sidecar create/sync/env-resolve status never reaches the log — which is the slow part of a validate and the bit a live dashboard most wants to show. `eventlog.OpSetup` is declared and rendered in `opTag` but nothing ever emits it. Move the wrap above `setupRemote`?

internal/cmd/validate.go
```diff
@@ -281,8 +284,25 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
...
+	}
+
 	execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
 
+	if opts.sidecarID != "" {
```
danmux: Completion events are gated on `opts.sidecarID != ""` but the wrap at 293 isn't, so a purely local validate writes events with an empty `sidecar_id` and never a terminal `done`/`error`. The TUI filters on `e.SidecarID == id`, so those are invisible noise that still grows the log. Skip the wrap too when there's no sidecar?

internal/sidecar/active.go
```diff
@@ -84,6 +85,47 @@ func LoadActive(ctx context.Context) (*ActiveSidecar, error) {
 	return LoadActiveFrom(ctx, dir)
 }
 
+// LoadAnyActive returns the most recently modified sidecar state file for the
```
danmux, line 88: `LoadAnyActive` picks the newest `sidecar*.json` by mtime, dropping both dimensions `findSidecarFile` keys on (session **and** branch). So switching branches promotes the other branch's sidecar, and `SaveActive` then re-keys that state under the current branch. Is cross-branch reuse intended, or should this be branch-scoped and only session-agnostic?

Also takes no `ctx` while every sibling loader does — worth keeping the signature consistent?

internal/config/paths.go
```diff
@@ -46,14 +46,20 @@ func AppData() (string, error) {
...
+// ensures callers that discover the root via different means — git's
+// --show-toplevel vs. a manual filesystem walk — always hash the same string.
 func ProjectDataDir(projectRoot string) (string, error) {
 	base, err := AppData()
 	if err != nil {
 		return "", err
 	}
-	sum := sha256.Sum256([]byte(filepath.Clean(projectRoot)))
+	clean := filepath.Clean(projectRoot)
+	if resolved, err := filepath.EvalSymlinks(clean); err == nil {
```
danmux: This changes the hash for any project reached through a symlink, so those users silently get a fresh empty data dir and lose their existing sidecar/snapshot state and event log. No migration and nothing in the PR description. How common is that in practice — worth a one-time fallback that reads the old un-resolved dir if the new one is empty?

internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+// Event is a single status event written to the log.
+type Event struct {
+	Ts          time.Time `json:"ts"`
+	SidecarID   string    `json:"sidecar_id,omitempty"`
+	SidecarName string    `json:"sidecar_name,omitempty"`
+	Branch      string    `json:"branch,omitempty"`
+	Op          Op        `json:"op"`
+	Level       string    `json:"level"` // "step", "info", "warn", "done"
```
danmux: Comment is missing `"error"` now that `LevelError` exists.

internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,178 @@
...
+	Msg         string    `json:"msg"`
+}
+
+const logFile = "events.jsonl"
```
danmux, line 39: `events.jsonl` is append-only with no rotation or cap — every sync and validate grows it forever, and `Recent` parses the whole file on the first poll of every `chunk watch`. Truncate to the last N lines on `Open`, or roll the file past a size limit?

internal/tui/watch/model.go
```diff
@@ -0,0 +1,666 @@
...
+	case tea.WindowSizeMsg:
+		m.width = msg.Width
+		m.height = msg.Height
+		return m, nil
+
+	case tea.KeyPressMsg:
+		switch msg.Code {
+		case 'q', tea.KeyEscape:
+			return m, tea.Quit
+		case 's', tea.KeyDown:
```
danmux: Nav is bound to `s`/`w`, the footer at 463 advertises `↑/↓`, and the PR description says `j`/`k` — three different answers, and `j`/`k` aren't wired at all. `w`/`s` is an odd choice for a list; arrows plus vim keys?

internal/tui/watch/model.go
```diff
@@ -0,0 +1,666 @@
...
+
+// groupEvents partitions events into op-runs. A run ends when a "done" or
+// "error" event arrives, or when the op changes.
+func groupEvents(events []eventlog.Event) []eventGroup {
+	var groups []eventGroup
+	cur := make(eventGroup, 0, len(events))
+	for _, e := range events {
+		if len(cur) > 0 && e.Op != cur[0].Op {
+			groups = append(groups, cur)
+			cur = make(eventGroup, 0, len(events)-len(groups))
```
danmux: `len(events)-len(groups)` as a capacity is meaningless here (groups counted, events remaining — different units) and can't help. `make(eventGroup, 0)` like the branch below?

internal/tui/watch/model.go
```diff
@@ -0,0 +1,666 @@
...
+		snap := hasSnapshotFile(p.DataDir)
+		sidecars := loadSidecars(p.DataDir, p.ProjectRoot, snap, newHeadRefs[i])
+		allSidecars = append(allSidecars, sidecars...)
+
+		if p.Log == nil {
+			continue
+		}
+		offset := m.offsets[i]
+		if offset == 0 {
+			recent, _ := p.Log.Recent(recentEvents)
```
danmux: First poll reads and parses the whole log twice: `Recent(recentEvents)` for the events, then `TailFrom(0)` purely for the byte offset, throwing its events away. Either have `Recent` return the offset, or just call `TailFrom(0)` and trim to `recentEvents`.

internal/tui/watch/model.go
```diff
@@ -0,0 +1,666 @@
...
+	}
+	return true
+}
+
+// headRef returns the full HEAD SHA for the git repo at dir.
+func headRef(dir string) string {
+	if dir == "" {
+		return ""
+	}
+	out, err := exec.Command("git", "-C", dir, "rev-parse", "HEAD").Output()
```
danmux: Raw `exec.Command("git", ...)` here and again in `cmd/watch.go:93`, while `internal/gitutil` exists and `sidecar.CurrentBranch` is called on the line right above this in `loadData`. Also no context — a wedged git (stale `index.lock`, slow network mount) blocks the poll goroutine indefinitely. `gitutil` + `CommandContext`?

internal/cmd/watch.go
```diff
@@ -0,0 +1,99 @@
...
+func newWatchCmd() *cobra.Command {
+	var all bool
+
+	cmd := &cobra.Command{
+		Use:          "watch [dir...]",
+		Short:        "Live dashboard for active sidecars and recent activity",
+		SilenceUsage: true,
+		Args:         cobra.ArbitraryArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !term.IsTerminal(int(os.Stdout.Fd())) {
```
danmux: TTY check reads `os.Stdout` directly rather than going through `iostream.FromCmd(cmd)` — `internal/tui/tty.go` already exists for this. Also `cwd, _ := os.Getwd()` at 33 swallows the error; on failure `roots` starts with `""` and `filepath.Abs("")` will fail with an unhelpful message.

internal/cmd/watch.go
```diff
@@ -0,0 +1,99 @@
...
+				seen[abs] = true
+
+				dataDir, err := config.ProjectDataDir(abs)
+				if err != nil {
+					return fmt.Errorf("watch: data dir for %s: %w", abs, err)
+				}
+
+				// Register this project so future --all runs find it.
+				if err := os.MkdirAll(dataDir, 0o755); err == nil {
+					_ = os.WriteFile(filepath.Join(dataDir, "project-root"), []byte(abs), 0o644)
```
danmux, line 71: This breadcrumb is now written in two places (here and `SaveActiveTo`), and here it's written for *any* path handed to `watch`, even a non-git dir with no sidecar — so `--all` picks it up forever after. Restrict it to real project roots and keep the write in the sidecar package?

internal/cmd/watch.go
```diff
@@ -0,0 +1,99 @@
...
+
+				entries = append(entries, watch.ProjectEntry{
+					Log:         el,
+					DataDir:     dataDir,
+					ProjectRoot: abs,
+				})
+			}
+
+			m := watch.New(entries)
+			p := tea.NewProgram(m)
```
danmux: `tea.NewProgram(m)` without `tea.WithContext(cmd.Context())` — the program won't tear down on parent-context cancellation, and Ctrl-C only works because of the `'c'`+`ModCtrl` case in `Update`.

danmux (commented):
Two things that don't fit on a line:

**No tests.** `internal/eventlog` (178 lines) and `internal/tui/watch` (666 lines) ship with zero tests and a fully manual test plan. `Append`/`Recent`/`TailFrom` are filesystem-only, and `groupEvents`, `stripUUIDSuffix`, `truncate`, `padLine`, `ago`, `loadSidecars` are pure — all cheap to cover, and `groupEvents`/`stripUUIDSuffix` are the two most likely to be quietly wrong.

**Collision with #459.** That PR also adds `iostream.LevelError` plus a red stderr formatter, named `ui.ErrError` (+ a `ui.ErrRed` helper) rather than `ui.ErrFailure`. Whichever lands second conflicts on `status.go` and `format.go` — worth settling the name now?

Also: `docs/CLI.md` gets no `watch` entry, and AGENTS.md points there as the command reference.

internal/config/paths.go
```diff
@@ -46,15 +46,39 @@ func AppData() (string, error) {
...
+	if r, err := filepath.EvalSymlinks(clean); err == nil {
+		resolved = r
+	}
+	newDir := filepath.Join(base, fmt.Sprintf("%x", sha256.Sum256([]byte(resolved))))
+
+	// Only attempt migration when the symlink actually changed the path.
+	if resolved != clean {
+		oldDir := filepath.Join(base, fmt.Sprintf("%x", sha256.Sum256([]byte(clean))))
+		if _, statErr := os.Stat(oldDir); statErr == nil {
+			if entries, readErr := os.ReadDir(newDir); readErr != nil || len(entries) == 0 {
```
michael-webster: The condition treats a ReadDir error (e.g. permission denied on an existing directory) the same as an empty directory, so the rename is attempted when it should not be. On Linux/macOS rename(2) rejects a non-empty destination with ENOTEMPTY so no data loss occurs, but the intent is wrong and a future platform or edge case could behave differently.

Suggested fix:

```go
if readErr == nil && len(entries) == 0 {
    _ = os.Rename(oldDir, newDir)
}
```

Only migrate when the directory is confirmed absent or empty, not when it is unreadable.

michael-webster (approved):

PR #448, opened 2026-07-24
feat: attribute telemetry to CircleCI user ID when token is available

michael-webster:
## Summary

- `GetCurrentUser` now decodes the `/api/v2/me` response and returns a `CurrentUser{ID, Login}` struct instead of just an error
- `telemetry.Meta` gains a `CircleCIUserID` field; `Track` uses it as the Segment `UserId` when set, falling back to the anonymous `InstanceID` otherwise
- `setupTelemetry` calls a new `resolveCircleCIUserID` helper that best-effort fetches the user ID via the CircleCI client — any failure (no token, network error) silently falls back to the existing anonymous attribution

Mirrors the approach used in circleci-cli (`internal/telemetry/sender.go` + `internal/apiclient/me.go`).

## Test plan

- [ ] `go test -race ./internal/circleci/ ./internal/telemetry/ ./internal/cmd/` passes
- [ ] With a valid `CIRCLECI_TOKEN` set, verify `CHUNK_TELEMETRY_LOG=1 chunk version` logs an event with the CircleCI user ID as `UserId`
- [ ] Without a token, verify telemetry falls back to the anonymous instance ID

### DIFF
--- a/internal/circleci/circleci_test.go
+++ b/internal/circleci/circleci_test.go
@@ -41,7 +41,10 @@ func TestNewClient(t *testing.T) {
 		assert.NilError(t, err)
 
 		ctx := context.Background()
-		assert.NilError(t, c.GetCurrentUser(ctx))
+		u, err := c.GetCurrentUser(ctx)
+		assert.NilError(t, err)
+		assert.Equal(t, u.ID, "user-123")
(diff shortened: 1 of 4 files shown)


michael-webster:
Closing — reconsidering whether authenticated user attribution is the right approach for telemetry.

PR #450, opened 2026-07-27
validate: buffer command output and show timing on success

hanabel1:
## Summary

- Buffer stdout/stderr during command execution; only flush to stderr on failure
- Report elapsed time on success via the status callback (`install passed (123ms)`)
- Clean up pre-commit and Stop hook output — successful runs no longer stream raw command noise

## Events

- `LevelDone` status event fires with `"<name> passed (<elapsed>)"` after a successful command run

## Not collected

- Command output on success (suppressed by design)

## Opt-out

N/A — this affects all local `chunk validate` runs

## Implementation

`runCommand` in `internal/validate/validate.go` now writes both stdout and stderr into a `bytes.Buffer`. On failure the buffer is flushed to `streams.Err` so users still see full output. On success only the timing summary fires via `status`.

## Test plan

- [x] `TestRunAll` updated to assert on status messages instead of raw stream output
- [x] Failure path still copies buffer to stderr (existing error-path tests cover this)
- [x] `task test` passes with `-race`

### DIFF
--- a/internal/validate/validate.go
+++ b/internal/validate/validate.go
@@ -1,6 +1,7 @@
 package validate
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -205,11 +206,16 @@ func runCommand(ctx context.Context, workDir, name, command string, timeoutSec i
 
 	cmd := exec.CommandContext(ctx, "sh", "-c", command)
 	cmd.Dir = workDir
-	cmd.Stdout = streams.Out
-	cmd.Stderr = streams.Err
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
 
+	start := time.Now()
 	err := cmd.Run()
+	elapsed := time.Since(start).Round(time.Millisecond)
+
 	if err != nil {
+		_, _ = io.Copy(streams.Err, &buf)
 		if ctx.Err() == context.DeadlineExceeded {
 			return fmt.Errorf("%s command timed out after %ds", name, timeoutSec)
 		}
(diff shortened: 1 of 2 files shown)


internal/validate/validate.go
```diff
@@ -205,11 +206,16 @@ func runCommand(ctx context.Context, workDir, name, command string, timeoutSec i
 
 	cmd := exec.CommandContext(ctx, "sh", "-c", command)
 	cmd.Dir = workDir
-	cmd.Stdout = streams.Out
-	cmd.Stderr = streams.Err
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
```
michael-webster, line 210: are we intending to point the Stdout to the same buffer? Is the intent to make it so we throw away stdout?

michael-webster: oh i see you want all the output to go in one buffer, then flush.

internal/validate/validate_test.go
```diff
@@ -168,12 +168,12 @@ func TestRunAll(t *testing.T) {
...
 		}}
-		streams, out, _ := newStreams()
+		streams, _, _ := newStreams()
 		var statusBuf bytes.Buffer
 
 		assert.NilError(t, RunAll(context.Background(), ".", cfg, testStatus(&statusBuf), streams))
-		assert.Assert(t, strings.Contains(out.String(), "installed"), "got: %s", out.String())
-		assert.Assert(t, strings.Contains(out.String(), "tested"), "got: %s", out.String())
+		assert.Assert(t, strings.Contains(statusBuf.String(), "install passed"), "got: %s", statusBuf.String())
+		assert.Assert(t, strings.Contains(statusBuf.String(), "test passed"), "got: %s", statusBuf.String())
```
michael-webster, line 176: can we confirm that the errors are also displayed?

michael-webster (changes requested):

PR #451, opened 2026-07-27
Add parallel sidecars

schurchleycci:
## Summary

- Adds `--count N` to `chunk sidecar create` to spin up multiple sidecars in parallel and set them as the active group
- Updates `chunk sidecar sync` to deliver one bundle to all sidecars in the active group in parallel
- Updates `chunk validate` to automatically provision one sidecar per remote command when two or more remote commands are configured (reuses an existing group of the right size)
- Fixes three bugs discovered during implementation: nil check on `active` in `resolveSidecarID`, missing error path in sync fan-out, and a race in parallel create result collection

## Test plan

- [ ] `chunk sidecar create --count 3` creates three sidecars and sets them as the active group
- [ ] `chunk sidecar sync` delivers the bundle to all sidecars in the active group in parallel
- [ ] `chunk validate` with two or more remote commands creates matching sidecars and runs them in parallel
- [ ] Single-sidecar workflows are unaffected (no `--count`, no group — existing behaviour unchanged)
- [ ] Unit tests pass: `task test`
- [ ] Acceptance tests pass: `task acceptance-test`

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -47,7 +47,6 @@ func newSidecarCmd() *cobra.Command {
 	cmd.AddCommand(newSidecarCreateCmd())
 	cmd.AddCommand(newSidecarDeleteCmd())
 	cmd.AddCommand(newSidecarExecCmd())
-	cmd.AddCommand(newSidecarRunCmd())
 	cmd.AddCommand(newSidecarAddSSHKeyCmd())
 	cmd.AddCommand(newSidecarSSHCmd())
 	cmd.AddCommand(newSidecarSyncCmd())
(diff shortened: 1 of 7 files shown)


internal/cmd/sidecar.go
```diff
@@ -456,28 +571,67 @@ func newSidecarSSHCmd() *cobra.Command {
...
+				cwd = source
+			}
+
+			// Fan-out path: sync one bundle to multiple sidecars in parallel.
+			// Activated by --sidecar-ids flag or when the active state is a sidecar group.
+			ids, err := resolveSidecarGroupIDs(cmd.Context(), sidecarIDs)
+			if err != nil {
+				return err
+			}
+			if len(ids) > 1 {
```
z00b, line 605: `--sidecar-ids` with exactly one ID is silently ignored. When `resolveSidecarGroupIDs` returns a one-element slice, this branch is skipped and execution falls through to `resolveSidecarID`, which ignores the provided ID and resolves from `--sidecar-id` (empty) or the active sidecar instead.

Concrete failure: `chunk sidecar sync --sidecar-ids=sb-target` with a different active sidecar syncs to the wrong sidecar with no warning.

```suggestion
			if len(ids) == 1 {
				sidecarID = ids[0]
			} else if err := resolveSidecarID(cmd.Context(), &sidecarID); err != nil {
				return err
			}
```

internal/cmd/sidecar.go
```diff
@@ -282,7 +397,7 @@ func newSidecarDeleteCmd() *cobra.Command {
 			}
 			io.ErrPrintf("%s\n", ui.Success(fmt.Sprintf("Deleted sidecar %s", sidecarID)))
 
-			if active, lerr := sidecar.LoadActive(cmd.Context()); lerr == nil && active != nil && active.SidecarID == sidecarID {
+			if active, lerr := sidecar.LoadActive(cmd.Context()); lerr == nil && active != nil && active.ID() == sidecarID {
```
z00b: Deleting a non-primary sidecar from an active group leaves stale active state. This check clears the active file only when the deleted ID matches `active.ID()` (the first element). For an active group `[sb-1, sb-2]`, deleting `sb-2` leaves the group file intact — subsequent `sync` or `validate` calls fan-out to both IDs and fail on the deleted one.

Should check whether `sidecarID` appears anywhere in `active.SidecarIDs`, not just the primary slot. Same issue exists in the snapshot delete command (~line 951).

internal/cmd/validate.go
```diff
@@ -158,19 +160,32 @@ func validateNeedsSidecar(explicitRemote bool, cfg *config.ProjectConfig, hook *
...
 	}
 	return envVars, nil
 }
 
+// syncToSidecarsOpts synchronises the working tree to all sidecars in the group in parallel.
+func syncToSidecarsOpts(ctx context.Context, client *circleci.Client, opts *validateOpts, workDir string, statusFn iostream.StatusFunc) error {
+	authSock := os.Getenv(config.EnvSSHAuthSock)
+	err := sidecar.BundleSyncFanOut(ctx, client, opts.sidecarIDs, opts.identityFile, authSock, opts.workdir, workDir, statusFn)
+	if err != nil {
+		return &userError{msg: "Could not sync to sidecars.", err: err}
```
z00b, line 187: `syncToSidecarsOpts` swallows the `NoOriginRemoteError` hint. All errors from `BundleSyncFanOut` are wrapped with the generic "Could not sync to sidecars." message, but the single-sidecar sync path and `newSidecarSyncCmd`'s fan-out path both check for this error type and surface `msgNoOriginRemote` + `suggestionAddOrigin`. Group-mode validate in a repo without `origin` gives no actionable guidance.

```suggestion
	if _, ok := errors.AsType[*sidecar.NoOriginRemoteError](err); ok {
		return &userError{msg: msgNoOriginRemote, suggestion: suggestionAddOrigin, err: err}
	}
	return &userError{msg: "Could not sync to sidecars.", err: err}
```

danmux (commented):
## Critical

### `internal/cmd/sidecar.go:89` — nil pointer dereference in `resolveSidecarID`

The original guard was `if active == nil`. The replacement `if active.ID() == ""` panics when `LoadActive` returns `(nil, nil)` — which is the normal no-active-sidecar state. This is a regression that will panic every time a user runs a `chunk sidecar` subcommand with no active sidecar set. Restore the nil check before calling `ID()`, or confirm the `ID()` implementation is nil-receiver safe (the diff doesn't show its body).

## Required

### `internal/cmd/sidecar.go:246` — orphaned sidecars on partial creation failure

`createSidecars` spawns N goroutines. On partial failure (e.g. 2 of 3 succeed) it returns the successful sidecars alongside a joined error. The caller receives the error, discards the sidecars, and nothing gets saved. The successfully-created sidecars are left running and untracked — no reap path will clean them up. Either abort-and-delete already-created sidecars on any failure, or save the partial result before returning the error so the user can reap them.

### `internal/cmd/validate.go` — silently discarded error in `shouldUseSidecarGroup`

```go
remoteCfg, _ := splitByRemote(cfg)
```

If `splitByRemote` returns an error the function silently returns `false`, causing fallthrough to the single-sidecar path with no user feedback. A config parse error should surface rather than silently changing execution strategy.

danmux (approved):

PR #454, opened 2026-07-27
Add git pre-commit hook installation to chunk init

schurchleycci:
## Summary

- `chunk init` now writes `.git/hooks/pre-commit` (mode `0755`) containing `chunk validate`, alongside the existing agent hook files (`.claude/settings.json`, `.codex/hooks.json`)
- Uses `git rev-parse --git-common-dir` so the hook lands in the shared hooks directory even when running from a worktree
- Idempotent: no-op if the hook already calls `chunk validate`; appends if a hook exists but doesn't call it

## Motivation

The `.claude/settings.json` Stop hook runs after every agent turn, not just ones where files were edited. A git pre-commit hook covers the commit path for all clients (human, any agent, any git tool) at the git layer, without the per-turn overhead.

## Test plan

- [ ] `TestWriteGitHookNewFile` — new hook written with correct content and executable bit
- [ ] `TestWriteGitHookAlreadyUpToDate` — idempotent on second call
- [ ] `TestWriteGitHookAppendsToExisting` — appends to a pre-existing hook that doesn't call `chunk validate`
- [ ] Full test suite passes locally and on sidecar

### DIFF
--- a/internal/cmd/completion.go
+++ b/internal/cmd/completion.go
@@ -12,6 +12,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/closer"
 	"github.com/CircleCI-Public/chunk-cli/internal/config"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
+	"github.com/CircleCI-Public/chunk-cli/internal/tui"
 	"github.com/CircleCI-Public/chunk-cli/internal/ui"
 )
 
@@ -125,6 +126,27 @@ func installCompletion(streams iostream.Streams) (err error) {
 
 	streams.ErrPrintln(ui.Success("Completion installed."))
 	return nil
+}
+
+func maybeInstallCompletions(streams iostream.Streams) {
+	installed, err := completionInstalled()
+	if err != nil {
+		streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Skipping shell completions: %v", err)))
+		return
+	}
+	if installed {
+		return
+	}
+	yes, confirmErr := tui.Confirm("Install shell completions?", true)
+	if confirmErr != nil {
+		streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not confirm: %v", confirmErr)))
+		return
+	}
+	if yes {
+		if installErr := installCompletion(streams); installErr != nil {
+			streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not install completions: %v", installErr)))
+		}
+	}
 }
 
 func newCompletionInstallCmd() *cobra.Command {
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -383,32 +383,18 @@ func writeAllHookFiles(workDir string, commands []config.Command, streams iostre
 			return err
 		}
 	}
-	if err := writeGitHook(workDir, streams); err != nil {
-		return err
-	}
 	return nil
 }
 
 const gitHookContent = "#!/bin/sh\nchunk validate\n"
 
 // writeGitHook writes a pre-commit hook to the repository's git hooks
-// directory. It uses --git-common-dir so it targets the shared hooks directory
-// even when workDir is a worktree. If a pre-commit hook already exists and
-// calls chunk validate it is left as-is; if it exists but does not call chunk
-// validate, the call is appended.
-func writeGitHook(workDir string, streams iostream.Streams) error {
-	gitCmd := exec.Command("git", "rev-parse", "--git-common-dir")
-	gitCmd.Dir = workDir
-	out, err := gitCmd.Output()
-	if err != nil {
-		return &userError{msg: "Could not determine git directory.", err: fmt.Errorf("git rev-parse --git-common-dir: %w", err)}
-	}
-
-	gitCommonDir := strings.TrimSpace(string(out))
-	if !filepath.IsAbs(gitCommonDir) {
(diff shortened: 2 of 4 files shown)


internal/cmd/init.go
```diff
@@ -383,6 +383,80 @@ func writeAllHookFiles(workDir string, commands []config.Command, streams iostre
...
+		streams.ErrPrintln(ui.Success("Git pre-commit hook already up to date"))
+		return nil
+	}
+
+	content := string(existing)
+	if len(content) > 0 && content[len(content)-1] != '\n' {
+		content += "\n"
+	}
+	content += "chunk validate\n"
+	if err := os.WriteFile(path, []byte(content), 0o755); err != nil {
```
michael-webster: **Required — `os.WriteFile` silently overrides existing file permissions on the append path**

`os.WriteFile` always sets the mode to the supplied value, so any pre-existing permission bits (e.g. `0o700`, `0o744`) are silently replaced with `0o755`. In the new-file path that is intentional, but in the append case the user may have had deliberately restricted permissions on their hook.

Consider reading the existing mode before the write and preserving it:

```suggestion
info, err := os.Stat(path)
if err != nil {
	return &userError{msg: "Could not stat .git/hooks/pre-commit.", suggestion: suggestionCheckPerms, err: fmt.Errorf("stat pre-commit hook: %w", err)}
}
if err := os.WriteFile(path, []byte(content), info.Mode()); err != nil {
```

internal/cmd/init_test.go
```diff
@@ -336,6 +337,78 @@ func TestInstallCompletionSkipsWhenAlreadyInstalled(t *testing.T) {
...
+	err := writeGitHook(dir, streams)
+	assert.NilError(t, err)
+
+	data, err := os.ReadFile(filepath.Join(hooksDir, "pre-commit"))
+	assert.NilError(t, err)
+	content := string(data)
+	assert.Assert(t, strings.HasPrefix(content, existing))
+	assert.Assert(t, strings.Contains(content, "chunk validate"))
+
+	assert.Assert(t, bytes.Contains(errOut.Bytes(), []byte("Updated .git/hooks/pre-commit")))
```
michael-webster, line 413: **Required — append test does not verify the file is still executable**

`TestWriteGitHookNewFile` asserts `info.Mode()&0o111 != 0`, but this test omits that check. Since the append path calls `os.WriteFile` with a hardcoded mode, adding the same assertion here would catch any regression where the mode argument is accidentally dropped or changed.

internal/cmd/init.go
```diff
@@ -383,6 +383,80 @@ func writeAllHookFiles(workDir string, commands []config.Command, streams iostre
 			return err
 		}
 	}
+	if err := writeGitHook(workDir, streams); err != nil {
```
michael-webster: **Suggestion — `writeGitHook` is not independently skippable**

`writeGitHook` is unconditionally called inside `writeAllHookFiles`, which is gated by `--skip-hooks`. Users who want agent hook files (`.claude/settings.json`, `.codex/hooks.json`) but not a git pre-commit hook have no way to opt out. Is this intentional? If not, a `--skip-git-hook` flag (or handling it alongside `--skip-hooks`) may be worth adding.

internal/cmd/init.go
```diff
@@ -383,6 +383,80 @@ func writeAllHookFiles(workDir string, commands []config.Command, streams iostre
...
+// directory. It uses --git-common-dir so it targets the shared hooks directory
+// even when workDir is a worktree. If a pre-commit hook already exists and
+// calls chunk validate it is left as-is; if it exists but does not call chunk
+// validate, the call is appended.
+func writeGitHook(workDir string, streams iostream.Streams) error {
+	gitCmd := exec.Command("git", "rev-parse", "--git-common-dir")
+	gitCmd.Dir = workDir
+	out, err := gitCmd.Output()
+	if err != nil {
+		return &userError{msg: "Could not determine git directory.", err: fmt.Errorf("git rev-parse --git-common-dir: %w", err)}
```
michael-webster: **Suggestion — redundant `git rev-parse` invocation**

`newInitCmd` already calls `git rev-parse --git-dir` to assert we're inside a git repo before `writeAllHookFiles` is reached. `writeGitHook` then issues a second shell-out with `--git-common-dir`. No correctness issue, but the duplicate process spawn could be avoided by resolving `--git-common-dir` once in `newInitCmd` and threading the resolved path down to `writeGitHook`.

michael-webster (commented):
Two required fixes and two suggestions.

internal/cmd/completion.go
```diff
@@ -127,6 +128,27 @@ func installCompletion(streams iostream.Streams) (err error) {
 	return nil
 }
 
+func maybeInstallCompletions(streams iostream.Streams) {
```
schurchleycci, line 131: This is to get under the cyclomatic complexity threshold

internal/cmd/init_test.go
schurchleycci: Fixed!

internal/cmd/init.go
schurchleycci: This is fixed 👍

internal/cmd/init.go
schurchleycci: `--skip-git-hook` will work now

internal/cmd/init.go
schurchleycci: Done

michael-webster (approved):

PR #455, opened 2026-07-27
Wire chunk validate remote execution to the async HTTP exec API

schurchleycci:
## Summary

- Adds `Env map[string]string` to `ExecRequest` so env vars are forwarded in the HTTP request body
- Replaces `openSSHSession` with `newExecFn`, which builds a remote exec function using `client.Exec()` instead of an SSH session
- Removes `identityFile` from the exec path in `runValidate` and `runSplitCommands` (SSH is still used for the sync step)
- Replaces three SSH-specific unit tests with a single test verifying env vars reach the HTTP request body

## Notes

Depends on #452. Base branch should be updated to `main` once that merges.

## Test plan

- [ ] All existing tests pass (`go test -race ./...`)
- [ ] `TestOpenAPIExecPassesEnvVars` → `TestNewExecFnPassesEnvVars` verifies env vars in request body
- [ ] Manual: `chunk validate --remote` runs commands on sidecar without SSH session

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -281,7 +281,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return err
 	}
 
-	execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.identityFile, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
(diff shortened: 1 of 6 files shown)


internal/cmd/validate.go
```diff
@@ -499,7 +484,7 @@ func hostForwardEnv(token string) map[string]string {
 // When freshlyCreated is true, SSH failures are hard errors rather than
```
danmux: `SSH failures` — should this be updated to `exec failures` (or similar) now that this calls `newExecFn` over HTTP rather than SSH?

danmux (approved):

PR #457, opened 2026-07-28
Cache successful validate results in hook mode

schurchleycci:
## Summary

- Adds a generic `FileCache[T]` (`internal/filecache`) — one JSON file per entry, filename is `sha256(key)`, silent miss on corrupt entries. Each write sweeps entries older than 7 days plus any staged file an interrupted run left behind, so the directory stays bounded
- Adds `CachedResult` and `BuildCacheKey` to `internal/validate`, `gitutil.Fingerprint` for the working-tree half of the key, and `internal/hashutil` for the length-prefixed digest primitive both of those rest on
- Replaces `validate.HasGitChanges` with that same fingerprint, so the Stop hook's existing clean-tree skip and the new cache read one pass over the tree instead of disagreeing about what changed
- Wires the cache into `chunk validate`'s Stop hook path: on a successful run the outcome is stored, and subsequent hook invocations with identical inputs skip execution and print a one-line "skipped" message

## Motivation

In hook mode `chunk validate` fires on every Claude Code Stop turn. Re-running `go test ./...` (or an SSH exec to a sidecar) when nothing has changed since the last successful run is pure waste.

Invalidation is by content hash. The key is the command name plus a SHA-256 over:

- **`.chunk/config.json` whole** — commands and the environment block that decides what they run against, so a project that gitignores `.chunk/` still re-runs when either changes
- **the execution target** — a result validated against one sidecar is never reused for another
- **the HEAD SHA**
- **the porcelain status and the contents of every changed file** — the status entries record deletions, renames and mode changes; hashing contents on top is what catches a second edit to an already-dirty file, which `git status` output alone reports identically

Failures are never cached, so the agent always retries after applying a fix.

Entries also expire 7 days after they were written, swept on the next write. That is not the invalidation mechanism — a stale key simply never matches again — it just stops a content-addressed directory from growing one file per Stop turn forever.

When the tree cannot be fingerprinted at all (a dirty submodule, an unreadable changed file, more than 64 MiB of changed files) no cache is consulted, and hook mode prints the reason — otherwise a repo that gets no benefit from this has no way to find out why.

## Test plan

- [x] `go test -race ./...` and `task acceptance-test` pass
- [x] `TestValidateHookCacheMissAfterEdit` — second run on an untouched tree skips; editing a file puts the commands back on
- [x] `TestValidateHookFailureIsNotCached` — a failing command is never stored, so the next invocation re-runs rather than reporting a hit
- [x] Both verified by mutation: caching failures reddens the first, dropping the working-tree digest from the key reddens the second
- [x] `TestFileCache_Put_Sweeps*` — stale entries and orphaned staged files are collected; a staged file still being written, and any file the cache did not write, are left alone

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -791,8 +791,9 @@ func validateCacheDir(workDir string) (string, error) {
 	return filepath.Join(projectDir, "validate-cache"), nil
 }
 
-// hookResultCache returns a ResultCache and cache key for hook-mode runs,
-// or (nil, "") when caching does not apply (non-hook or inline cmd).
+// hookResultCache returns a ResultCache and cache key for hook-mode runs, or
+// (nil, "") when caching does not apply: non-hook runs, inline commands, or a
+// working tree whose state cannot be hashed reliably.
 func hookResultCache(hook *hookContext, inlineCmd, workDir, commandName string, cfg *config.ProjectConfig) (validate.ResultCache, string) {
 	if hook == nil || inlineCmd != "" {
 		return nil, ""
@@ -801,7 +802,10 @@ func hookResultCache(hook *hookContext, inlineCmd, workDir, commandName string, 
 	if err != nil {
 		return nil, ""
 	}
-	key := validate.BuildCacheKey(workDir, commandName, cfg.Commands)
+	key, ok := validate.BuildCacheKey(workDir, commandName, cfg.Commands)
+	if !ok {
+		return nil, ""
+	}
 	return filecache.FileCache[validate.CachedResult]{Dir: cacheDir}, key
 }
 
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -22,6 +22,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/eventlog"
 	"github.com/CircleCI-Public/chunk-cli/internal/filecache"
 	"github.com/CircleCI-Public/chunk-cli/internal/gitremote"
+	"github.com/CircleCI-Public/chunk-cli/internal/gitutil"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 	"github.com/CircleCI-Public/chunk-cli/internal/session"
 	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
@@ -140,9 +141,12 @@ func newValidateCmd() *cobra.Command {
 }
 
 // initHook applies hook-specific context, stream, and early-exit logic.
+// tree is the working-tree fingerprint for this run; the zero value means it
+// could not be established, and reports the tree as not clean so validation
+// still runs in ambiguous cases.
 // Returns updated ctx and streams, a skip flag (true = return nil immediately),
 // and a non-nil error when the hook should exit with a non-zero code.
-func initHook(ctx context.Context, hook *hookContext, workDir string, streams iostream.Streams) (context.Context, iostream.Streams, bool, error) {
+func initHook(ctx context.Context, hook *hookContext, workDir string, tree gitutil.Worktree, streams iostream.Streams) (context.Context, iostream.Streams, bool, error) {
 	if hook == nil {
 		return ctx, streams, false, nil
 	}
@@ -157,7 +161,7 @@ func initHook(ctx context.Context, hook *hookContext, workDir string, streams io
 		streams.ErrPrintln("chunk validate: hooks are disabled — skipping validation")
 		return ctx, streams, false, validate.NewHookExitError(1)
 	}
-	if !validate.HasGitChanges(workDir) {
+	if tree.Clean {
 		return ctx, streams, true, nil
 	}
 	return ctx, streams, false, nil
@@ -209,9 +213,20 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 	hook := detectHook(cmd.InOrStdin())
 	ctx := cmd.Context()
 
+	// The working-tree fingerprint answers both hook-only questions about the
+	// tree: whether there is anything to validate at all, and whether this exact
+	// tree already validated successfully. Computing it once keeps the two
+	// consistent and costs a single pass over the changed files. On failure it is
+	// the zero Worktree, which reads as "not clean" and is refused as a cache
+	// key, so both questions fall back to running the commands.
+	var tree gitutil.Worktree
+	if hook != nil {
+		tree, _ = gitutil.Fingerprint(workDir)
+	}
+
 	var skip bool
 	var hookErr error
-	ctx, streams, skip, hookErr = initHook(ctx, hook, workDir, streams)
+	ctx, streams, skip, hookErr = initHook(ctx, hook, workDir, tree, streams)
 	if hookErr != nil {
 		return hookErr
 	}
@@ -270,7 +285,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return errSilentExit
 	}
 
-	resultCache, cacheKey := hookResultCache(hook, opts.inlineCmd, workDir, name, cfg, execTarget(opts, cfg, activeSidecar))
+	resultCache, cacheKey := hookResultCache(hook, opts.inlineCmd, workDir, tree, name, cfg, execTarget(opts, cfg, activeSidecar))
 	if resultCache != nil {
 		if _, ok := resultCache.Get(cacheKey); ok {
 			streams.ErrPrintln("chunk validate: skipped (no changes since last successful run)")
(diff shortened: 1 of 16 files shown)


internal/filecache/filecache.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// FileCache is a generic JSON-on-disk cache keyed by arbitrary strings.
+// Each entry is stored as a single file whose name is the hex-encoded SHA-256
+// of the raw key string. Concurrent writes to the same key are safe: Put stages
+// the entry in its own temporary file and renames it into place, so a reader
+// sees either the previous entry or a complete new one, never a mixture. Two
+// writers racing on one key both produce valid entries and the last rename
+// wins.
+type FileCache[T any] struct {
```
michael-webster, line 43: i'm fine with a file based approach. though for our case, combining git's write-tree (for committed) and then a file based cache for uncommitted changes might be marginally faster depending on the diff. that's really just optimization though.

internal/validate/cache.go
```diff
@@ -0,0 +1,89 @@
+package validate
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
```
michael-webster: `ResultCache` has a single implementation (`filecache.FileCache[CachedResult]`). Per the team's no-single-impl-interface rule, could this be the concrete type at call sites instead? The integration test in `validate_test.go` already exercises the real `FileCache` end-to-end, so the interface adds no testability benefit.

internal/cmd/validate.go
```diff
@@ -283,6 +313,10 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 
 	execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.identityFile, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
 
+	if execErr == nil && resultCache != nil {
```
michael-webster, line 348: The `Put` error is silently discarded. A permission failure or full disk will cause caching to never work, with no user-visible signal — subsequent hook runs will keep re-executing all commands without explanation. Could this at least log to `streams.ErrPrintln` at a dim level so a broken cache directory surfaces?

internal/validate/cache.go
schurchleycci: Yup! Fixed this

internal/cmd/validate.go
schurchleycci: Good idea, added!

danmux:
needs rebase

danmux (commented):
Description mentions a `ResultCache` interface that isn't in the diff. Worth dropping from the body since there is only one implementation anyway.

internal/cmd/validate.go
```diff
@@ -291,6 +321,12 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 
 	execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
 
+	if execErr == nil && resultCache != nil {
```
danmux, line 348: This condition is the whole "failures are never cached" guarantee, and nothing tests it. `TestValidateHookCacheHitResetsAttempts` only runs `exit 0`.

If this ever gets refactored so a failure is stored, every later hook invocation on the same tree prints `skipped` and returns nil, so the agent stops with the build broken and no test goes red. Can we get a sibling test with `Run: "exit 1"` asserting the second run does not skip? A miss-path case (edit a file between runs, assert re-execution) would cover the other unchecked box in the test plan.

internal/filecache/filecache.go
```diff
@@ -0,0 +1,80 @@
...
+		return zero, false
+	}
+	return v, true
+}
+
+// Put writes v to the cache under key, creating Dir and any parent directories
+// as needed. The entry is staged in a temporary file alongside its final path
+// and renamed into place, which is atomic within a single directory. Entries
+// have mode 0600, the mode os.CreateTemp applies and the rename preserves.
+func (c FileCache[T]) Put(key string, v T) error {
```
danmux, line 74: Superseded entries are never removed, so this is roughly one file per Stop turn, per project, indefinitely. `CachedResult.CachedAt` is written and, per its own comment, read by nothing.

Also, `os.CreateTemp` orphans a `.tmp-*` file if the process dies between staging and rename, which is what happens when a Stop hook is interrupted. Those aren't collected either.

Since `CachedAt` is already there, would a sweep on `Put` (drop entries older than N days, or keep the newest few) be cheap enough to avoid the manual `rm -rf` that HOOKS.md currently prescribes?

internal/cmd/validate.go
```diff
@@ -261,6 +278,19 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
...
 
+	resultCache, cacheKey := hookResultCache(hook, opts.inlineCmd, workDir, tree, name, cfg, execTarget(opts, cfg, activeSidecar))
+	if resultCache != nil {
+		if _, ok := resultCache.Get(cacheKey); ok {
+			streams.ErrPrintln("chunk validate: skipped (no changes since last successful run)")
+			// A cache hit is a success, so it clears the failure counter just as
+			// WrapHookResult does for a real one. Leaving a stale count behind
+			// would bring the "ask the user for guidance" bail-out forward by a
+			// turn. resultCache is only non-nil in hook mode, so hook is set.
+			validate.ResetAttempts(hook.sessionID)
```
danmux: This is a second hook success path that has to stay in sync with `finishValidate` by hand. The comment is honest about it, but the next change to the success branch there (another counter, an event log write, a telemetry attribute) silently won't happen on a cache hit, and the symptom is drifting hook state rather than a failing test.

Would `return finishValidate(cmd, hook, nil, start, opts.sidecarID, cfg, statusFn, streams)` work here instead? Collapses the two paths and the duplicated reasoning goes with it.

internal/validate/cache.go
```diff
@@ -0,0 +1,83 @@
...
+// gitutil.Fingerprint could not establish the tree's state — or the commands
+// cannot be serialized. Callers must not read or write the cache in that case:
+// with no trustworthy git state the key would depend only on the config and
+// would therefore stay stable across code changes, turning every subsequent run
+// into a false cache hit.
+func BuildCacheKey(in CacheKeyInputs) (string, bool) {
+	if in.Worktree.Head == "" || in.Worktree.Digest == "" {
+		return "", false
+	}
+	cfgBytes, err := json.Marshal(in.Commands)
```
danmux: Only `Commands` is hashed. `execTarget` rescues `Validation.SidecarImage`, but `cfg.Environment` is in the key nowhere, and survives today only because `.chunk/config.json` shows up in `git status` and gets content-hashed (or moves HEAD when committed).

In a project that gitignores `.chunk/`, editing the `environment` block to add an env var or package the commands need changes neither the fingerprint nor this hash, so the next invocation reports a hit and never re-runs against the new environment. The gitignore caveat in HOOKS.md covers files a command reads, not chunk's own config driving the run.

Any reason not to marshal the fields that determine the run, or `cfg` wholesale, given it is already loaded?

internal/gitutil/fingerprint.go
```diff
@@ -0,0 +1,163 @@
...
+// digest budget costs a stat rather than a full pass over the file.
+func hashFile(h io.Writer, path string, remaining int64) (int64, bool) {
+	f, err := os.Open(path)
+	if err != nil {
+		return 0, errors.Is(err, os.ErrNotExist)
+	}
+	defer func() { _ = f.Close() }()
+
+	info, err := f.Stat()
+	if err != nil || !info.Mode().IsRegular() {
```
danmux: One unreadable file, one non-regular changed path (a dirty submodule), or one changed file over the budget fails the whole fingerprint, which makes `hookResultCache` return nil for that repo on every invocation.

So a repo with a submodule gets no benefit from this PR and no way to find out why: no `skipped` line ever, no diagnostic, and the HOOKS.md remedy doesn't apply because the cache was never consulted. A dim stderr note when `Fingerprint` fails in hook mode would make it diagnosable. That wants an error return rather than a bool to say which condition hit, which reads better at the call site too.

internal/gitutil/fingerprint.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return 0, false
+	}
+	_, _ = h.Write(fh.Sum(nil))
+	return n, true
+}
+
+// writePart mixes s into h length-prefixed, so that concatenations of different
+// parts cannot collide (["ab","c"] and ["a","bc"] hash differently).
+func writePart(h io.Writer, s string) {
```
danmux: Byte-identical to `writePart` in `internal/validate/cache.go`, doc comment included. It is the length-prefixing primitive both collision-avoidance arguments rest on, so if one copy is ever strengthened the other stays as-is, both remain internally consistent, and all tests keep passing while the two digests quietly diverge.

Separately: `validate.CacheKey` is exported but its only caller is `BuildCacheKey` in the same package.

danmux (approved):

PR #459, opened 2026-07-28
Improve chunk validate stop hook UX

hanabel1:
## Summary

Reworks `chunk validate` output in stop hook mode to be structured and timing-aware on both success and failure.

**Before:** every command printed its run string as it started, output streamed live regardless of outcome, no per-command timing, no indication of which commands were skipped.

**After:**
- Commands show in aligned columns with elapsed time: `✓ test  3.7s`, `✓ lint  0.4s`
- On **success**: full progress output shown, then `✓ chunk validate passed (9.4s)`
- On **failure**: full buffered output flushed to stderr, failed command shows `✗ test  9.6s`, remaining commands show `⚠ lint  skipped`
- Total elapsed shown at the end: `✗ done in 12.8s (failed)` or `done in 9.4s`
Screenshot 2026-07-28 at 3 03 17 PM
Screenshot 2026-07-28 at 3 41 43 PM

**Known limitation:** Claude Code shows a "Stop hook error: Failed with non-blocking status code: ✓ chunk validate passed (Xs)" prefix on success. This is a Claude Code behavior — any stderr output on exit 0 from a stop hook gets wrapped in that message. The success info is still surfaced; it's just noisier than ideal.

Also fixes a bug where a setup error (e.g. sync failure) would silently swallow the "using sidecar X" message, breaking `TestValidateHookMode_SessionIsolation`.

**Failure example:**
```
  ✗ test          9.6s
  ⚠ test-changed  skipped
  ⚠ lint          skipped
  ⚠ format        skipped
  ✗ done in 12.8s (failed)
```

## Implementation

- Add `LevelError` to `iostream.Status` and `ErrError`/`ErrRed` formatting helpers
- Add `FormatDuration` to `internal/validate` for consistent ms/s/m output
- Hook mode buffers all output into `hookBuf`; flushed to stderr on both success and failure
- Wrap `setupRemote` + `loadSidecarEnvVars` + `runValidate` in a closure so all error paths reach the `io.Copy` flush
- Local commands buffer stdout/stderr and flush only on failure
- Exit 0 on hook success to avoid double "non-blocking" message

## Test plan

- [ ] `task test` passes
- [ ] Stop hook with passing validate shows full output + `✓ chunk validate passed`
- [ ] Stop hook with failing validate shows full output, `✗ done in Xs (failed)`, exits 2 to re-signal Claude

### DIFF
--- a/acceptance/validate_test.go
+++ b/acceptance/validate_test.go
@@ -173,10 +173,10 @@ func TestValidateRunLocal(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	assert.Assert(t, strings.Contains(combined, "echo installed"),
-		"expected install command output, got: %s", combined)
-	assert.Assert(t, strings.Contains(combined, "echo tested"),
-		"expected test command output, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "install"),
+		"expected install status in output, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "test"),
+		"expected test status in output, got: %s", combined)
 }
 
 func TestValidateRunLocalFailure(t *testing.T) {
@@ -244,9 +244,9 @@ func TestValidateRunNamed(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	assert.Assert(t, strings.Contains(combined, "echo tested"),
-		"expected test command in output, got: %s", combined)
-	// Should not run install
+	assert.Assert(t, strings.Contains(combined, "test"),
+		"expected test status in output, got: %s", combined)
+	// Should not run install — its echo output would appear on failure, not on success
 	assert.Assert(t, !strings.Contains(combined, "echo installed"),
 		"should not run install when running named test command, got: %s", combined)
 }
@@ -280,8 +280,9 @@ func TestValidateInlineCmd(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	assert.Assert(t, strings.Contains(combined, "inline-output"),
-		"expected inline command output, got: %s", combined)
+	// Command output is buffered and discarded on success; check for the status line instead.
+	assert.Assert(t, strings.Contains(combined, "custom"),
+		"expected custom command status in output, got: %s", combined)
 }
 
 func TestValidateInlineCmdDryRun(t *testing.T) {
--- a/acceptance/validate_test.go
+++ b/acceptance/validate_test.go
@@ -559,8 +559,8 @@ func TestValidateHookMode_SuccessLine(t *testing.T) {
 		hookStdin(t, "test-session-success-line", false))
 
 	assert.Equal(t, result.ExitCode, 0, "expected exit 0 for passing hook; stderr: %s", result.Stderr)
-	assert.Assert(t, strings.Contains(result.Stderr, "chunk validate passed"),
-		"expected 'chunk validate passed' in stderr; got: %s", result.Stderr)
+	assert.Assert(t, strings.Contains(result.Stdout, "chunk validate passed"),
+		"expected 'chunk validate passed' in stdout; got stdout: %s stderr: %s", result.Stdout, result.Stderr)
 }
 
 // TestValidateHookMode_SetupErrorFlushedToStderr verifies that when setup fails
--- a/acceptance/validate_test.go
+++ b/acceptance/validate_test.go
@@ -173,10 +173,10 @@ func TestValidateRunLocal(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	assert.Assert(t, strings.Contains(combined, "install"),
-		"expected install status in output, got: %s", combined)
-	assert.Assert(t, strings.Contains(combined, "test"),
-		"expected test status in output, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "installed"),
+		"expected install command output in result, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "tested"),
+		"expected test command output in result, got: %s", combined)
 }
 
 func TestValidateRunLocalFailure(t *testing.T) {
@@ -244,11 +244,10 @@ func TestValidateRunNamed(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	assert.Assert(t, strings.Contains(combined, "test"),
-		"expected test status in output, got: %s", combined)
-	// Should not run install — its echo output would appear on failure, not on success
-	assert.Assert(t, !strings.Contains(combined, "echo installed"),
-		"should not run install when running named test command, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "tested"),
+		"expected test command output in result, got: %s", combined)
+	assert.Assert(t, !strings.Contains(combined, "installed"),
+		"install command must not run when only 'test' is requested, got: %s", combined)
 }
 
 func TestValidateRunNamedNotConfiguredNonTTY(t *testing.T) {
@@ -280,9 +279,8 @@ func TestValidateInlineCmd(t *testing.T) {
 
 	assert.Equal(t, result.ExitCode, 0, "stderr: %s", result.Stderr)
 	combined := result.Stdout + result.Stderr
-	// Command output is buffered and discarded on success; check for the status line instead.
-	assert.Assert(t, strings.Contains(combined, "custom"),
-		"expected custom command status in output, got: %s", combined)
+	assert.Assert(t, strings.Contains(combined, "inline-output"),
+		"expected command output in result, got: %s", combined)
 }
 
 func TestValidateInlineCmdDryRun(t *testing.T) {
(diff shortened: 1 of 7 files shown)


internal/cmd/validate.go
```diff
@@ -268,27 +275,53 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return err
 	}
 
-	freshlyCreated, err := setupRemote(ctx, circleCIClient, opts, image, cfg, activeSidecar, statusFn, workDir, streams)
-	if err != nil {
-		return err
+	// In hook context, buffer all detailed output. On success we discard it
+	// and show only a brief summary; on failure we flush the full detail.
```
danmux: This comment no longer matches the code — line 319 flushes `hookBuf` unconditionally, on success as well as failure. Nothing is discarded any more.

internal/cmd/validate.go
```diff
@@ -268,27 +275,53 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
...
+	var hookBuf *bytes.Buffer
+	if hook != nil {
+		hookBuf = new(bytes.Buffer)
+		streams = iostream.Streams{Out: hookBuf, Err: hookBuf}
```
danmux: Given the buffer is now always flushed at 319, what does buffering still buy? The cost is that in hook mode the agent sees nothing at all until the whole run finishes — for a multi-minute remote validate that's a long silence, and it undoes the live streaming restored in `runCommand`. Could this just write straight through and only append the summary at the end?

internal/cmd/validate.go
```diff
@@ -268,27 +275,53 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
...
+		// Only load env vars and resolve secrets when a sidecar is actually
+		// being used — avoids parsing .env.local or hitting secrets APIs on
+		// purely local runs.
+		envVars, err := loadSidecarEnvVars(ctx, circleCIClient, opts, workDir, statusFn)
+		if err != nil {
+			return err
+		}
+		return runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, freshlyCreated, opts.workdir, allRemote, envVars, cfg, statusFn, streams)
+	}()
+	if hook == nil {
```
danmux: `hook == nil` suppresses the total in hook mode, but the failure example in the PR description shows `✗ done in 12.8s (failed)` — so on a hook failure there's now no total elapsed at all, only the per-command times. Intended?

internal/cmd/validate.go
```diff
@@ -268,27 +275,53 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
...
 	if hook != nil {
 		maxAttempts := cfg.StopHookMaxAttempts
 		if maxAttempts <= 0 {
 			maxAttempts = validate.DefaultMaxAttempts
 		}
-		return validate.WrapHookResult(hook.sessionID, execErr, maxAttempts, streams.Err)
+		hookErr := validate.WrapHookResult(hook.sessionID, execErr, maxAttempts, streams.Err)
+		_, _ = io.Copy(cmd.ErrOrStderr(), hookBuf)
+		if hookErr == nil && execErr == nil {
+			_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "%s\n", ui.ErrSuccess(fmt.Sprintf("chunk validate passed (%s)", validate.FormatDuration(time.Since(start)))))
```
danmux: The success line goes to stderr, which is exactly what triggers the "Failed with non-blocking status code" wrapper you call out as a known limitation. Since you already `return nil` here, would writing this one line to `cmd.OutOrStdout()` avoid the wrapper entirely?

internal/validate/validate.go
```diff
@@ -74,14 +85,25 @@ func RunAll(ctx context.Context, workDir string, cfg *config.ProjectConfig, stat
...
+		if err := runCommand(ctx, workDir, c.Name, c.Run, c.Timeout, maxWidth, status, streams); err != nil {
 			for j := i + 1; j < len(cfg.Commands); j++ {
-				status(iostream.LevelWarn, fmt.Sprintf("%s: skipped (%s failed)", cfg.Commands[j].Name, c.Name))
+				status(iostream.LevelWarn, fmt.Sprintf("%-*s  skipped", maxWidth, cfg.Commands[j].Name))
 			}
 			return err
 		}
 	}
+	if len(cfg.Commands) > 1 {
+		status(iostream.LevelDone, fmt.Sprintf("%-*s  %s", maxWidth, fmt.Sprintf("%d passed", len(cfg.Commands)), FormatDuration(time.Since(start))))
```
danmux: Padding `"%d passed"` into the command-name column via a nested `Sprintf` reads odd — the string isn't a command name so the alignment doesn't buy anything. Also this aggregate only appears for `len(cfg.Commands) > 1` and `RunRemote` has no equivalent, so remote runs never get a summary line. Deliberate asymmetry?

internal/validate/validate.go
```diff
@@ -118,12 +140,25 @@ func RunRemote(ctx context.Context, execFn func(ctx context.Context, script stri
 		}
 		commands = []config.Command{*c}
 	}
+
+	maxWidth := 0
```
danmux: Third copy of this max-name-length loop (also at 88). A small `nameWidth(commands []config.Command) int` would drop all three.

More substantive: `cmd/validate.go` splits commands into `localCfg`/`remoteCfg` and calls `RunAll`/`RunRemote` separately, so each subset computes its own width — the local and remote blocks won't align with each other in a mixed run, which is the case the column was added for.

internal/validate/validate.go
```diff
@@ -118,12 +140,25 @@ func RunRemote(ctx context.Context, execFn func(ctx context.Context, script stri
...
+
+	for i, c := range commands {
 		run := expandCommand(workDir, c.Run)
 		script := "cd " + shellEscape(dest) + " && " + run
-		status(iostream.LevelInfo, fmt.Sprintf("Running %s (remote): %s", c.Name, c.Run))
+		start := time.Now()
 		stdout, stderr, exitCode, err := execFn(ctx, script)
+		elapsed := time.Since(start)
 		if err != nil {
+			status(iostream.LevelError, fmt.Sprintf("%-*s  exec error: %v", maxWidth, c.Name, err))
```
danmux: The exec error is reported twice: once as a status line here, once via the wrapped error returned at 162 which `cmd/` prints. Drop the `%v` from the status line and keep the returned error?

internal/validate/validate.go
```diff
@@ -133,17 +168,26 @@ func RunRemote(ctx context.Context, execFn func(ctx context.Context, script stri
 			_, _ = fmt.Fprint(streams.Err, stderr)
 		}
 		if exitCode != 0 {
+			if stdout != "" || stderr != "" {
+				status(iostream.LevelInfo, c.Name+":")
```
danmux: This `c.Name+":"` header lands *after* stdout/stderr have already been written at 165–170, so it labels nothing — it reads as a header for whatever comes next. Same at 202. Move it above the two `Fprint` calls, or drop it?

internal/validate/validate.go
```diff
@@ -133,17 +168,26 @@ func RunRemote(ctx context.Context, execFn func(ctx context.Context, script stri
...
+			}
+			status(iostream.LevelError, fmt.Sprintf("%-*s  %s", maxWidth, c.Name, FormatDuration(elapsed)))
+			for j := i + 1; j < len(commands); j++ {
+				status(iostream.LevelWarn, fmt.Sprintf("%-*s  skipped", maxWidth, commands[j].Name))
```
danmux: Third copy of the "mark the rest skipped" loop (99, 160, 176). Worth a `skipRemaining(status, commands[i+1:], width)` helper?

internal/validate/validate.go
```diff
@@ -154,8 +198,13 @@ func RunRemoteInline(ctx context.Context, execFn func(ctx context.Context, scrip
 		_, _ = fmt.Fprint(streams.Err, stderr)
 	}
 	if exitCode != 0 {
+		if stdout != "" || stderr != "" {
+			status(iostream.LevelInfo, name+":")
+		}
+		status(iostream.LevelError, fmt.Sprintf("%-*s  %s", len(name), name, FormatDuration(elapsed)))
 		return fmt.Errorf("remote %s failed with exit code %d", name, exitCode)
 	}
+	status(iostream.LevelDone, fmt.Sprintf("%-*s  %s", len(name), name, FormatDuration(elapsed)))
```
danmux: `%-*s` with width `len(name)` is a no-op — it can never pad. Same at 204, and `RunInline`/`RunNamed` pass `len(name)`/`len(c.Name)` as `nameWidth` at 70/79 for the same non-effect. Pass `0` and let the single-command paths print unpadded?

internal/validate/validate.go
```diff
@@ -208,17 +256,23 @@ func runCommand(ctx context.Context, workDir, name, command string, timeoutSec i
 	cmd.Stdout = streams.Out
 	cmd.Stderr = streams.Err
 
+	start := time.Now()
 	err := cmd.Run()
+	elapsed := time.Since(start)
+
 	if err != nil {
 		if ctx.Err() == context.DeadlineExceeded {
+			status(iostream.LevelError, fmt.Sprintf("%-*s  timed out after %ds", nameWidth, name, timeoutSec))
```
danmux: Timeout branch reports `timed out after %ds` but not the elapsed time, while every sibling branch reports a duration. Minor inconsistency in the column.

acceptance/validate_test.go
```diff
@@ -548,6 +547,59 @@ func writeSidecarState(t *testing.T, e *testenv.TestEnv, projectRoot, sessionID,
...
+	env.Extra["CIRCLECI_ORG_ID"] = "org-aaa"
+
+	result := binary.RunCLIWithStdin(t, []string{
+		"validate", "--identity-file", identityFile,
+	}, env, workDir, hookStdin(t, "test-session-setup-err", false))
+
+	assert.Assert(t, result.ExitCode != 0, "expected failure; stderr: %s", result.Stderr)
+	// The hook output (including the setup error) must be flushed to stderr,
+	// not silently swallowed when hookBuf is non-empty on error exit.
+	assert.Assert(t, strings.Contains(result.Stderr, "chunk validate"),
```
danmux: This assertion is vacuous for what the test is named for: `statusFn(LevelStep, "chunk validate")` fires unconditionally at `validate.go:287`, so `Stderr` contains `"chunk validate"` even if the setup error is swallowed by `hookBuf`. Assert on the actual sync/SSH failure text instead — otherwise the test can't fail for the regression it guards.

internal/validate/validate_test.go
```diff
@@ -280,9 +315,11 @@ func TestRunRemote(t *testing.T) {
 			{Name: "test", Run: "echo test"},
 		}}
 		streams, out, _ := newStreams()
+		var statusBuf bytes.Buffer
 
-		assert.NilError(t, RunRemote(context.Background(), execFn, cfg, "", "/workspace", t.TempDir(), func(iostream.Level, string) {}, streams))
+		assert.NilError(t, RunRemote(context.Background(), execFn, cfg, "", "/workspace", t.TempDir(), testStatus(&statusBuf), streams))
 		assert.Assert(t, strings.Contains(out.String(), "remote output"), "got: %s", out.String())
+		assert.Assert(t, strings.Contains(statusBuf.String(), "install"), "got: %s", statusBuf.String())
```
danmux: `Contains(statusBuf, "install")` matches `[done] install`, `[error] install`, and `[warn] install  skipped` alike. Same at 447 and 514. Since `testStatus` now tags levels, `"[done] install"` would actually pin the outcome — which is the point of the tagging change.

danmux (commented):
Heads-up on a collision: #446 also adds `iostream.LevelError` and a red stderr formatter, but names it `ui.ErrFailure` (inline `wrapErr("31", ...)`) where this one adds `ui.ErrRed` + `ui.ErrError`. Whichever merges second will conflict on both files — worth agreeing on one name before either lands?

Rest of the comments are inline.

internal/validate/validate.go
```diff
@@ -13,6 +13,7 @@ import (
 
 	"github.com/CircleCI-Public/chunk-cli/internal/config"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
+	"github.com/CircleCI-Public/chunk-cli/internal/ui"
```
michael-webster: Other business packages (internal/sidecar, internal/buildprompt) never import internal/ui — they pass plain strings to status() and let the cmd/ layer apply formatting via newStatusFunc. This is the only non-cmd package importing internal/ui, which sets a precedent and risks an import cycle if internal/ui ever needs to reference internal/validate.

Suggested fix: replace ui.FormatDuration with a small private helper in validate.go:

```go
func formatElapsed(d time.Duration) string {
    if d < time.Second {
        return fmt.Sprintf("%dms", d.Milliseconds())
    }
    if d < time.Minute {
        return fmt.Sprintf("%.1fs", d.Seconds())
    }
    return d.Round(time.Second).String()
}
```

Same logic, no cross-layer import. The internal/ui import can then be dropped entirely.

michael-webster (approved):
left one comment to fix before merging.

PR #462, opened 2026-07-29
Reduce org ID friction: add chunk org list and auto-detect org ID in init

schurchleycci:
## Summary

Sidecar commands need a CircleCI org ID, but there was no way to discover or capture it through the CLI — users either knew it already or got blocked. This PR closes that gap.

`chunk init` now auto-detects the org ID after writing the project config. Single-org users get it picked silently; multi-org users see a picker. The ID is saved to `.chunk/config.json` so sidecar commands just work from that point on.

For cases where init isn't an option, `chunk org list` provides a discovery command. The `orgPicker` error message now points to it, and the sidecar skill uses it instead of stopping to ask the user.

## Test plan

- [ ] `chunk init` while authenticated to one org — org ID auto-selected and written to `.chunk/config.json`
- [ ] `chunk init` while authenticated to multiple orgs — picker appears, selected ID written
- [ ] `chunk init` while not authenticated — completes without error, no org ID written
- [ ] `chunk init` in a non-interactive environment (no TTY) — completes without error or spurious warning
- [ ] `chunk init --skip-org-id` — org ID detection step skipped
- [ ] `chunk org list` — orgs printed in table form; `--json` returns JSON
- [ ] `chunk sidecar create` with single-org account and no orgID in config — org auto-selected, no picker

### DIFF
--- a/internal/cmd/init.go
+++ b/internal/cmd/init.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -12,6 +13,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/CircleCI-Public/chunk-cli/internal/anthropic"
+	"github.com/CircleCI-Public/chunk-cli/internal/authprompt"
 	"github.com/CircleCI-Public/chunk-cli/internal/config"
 	"github.com/CircleCI-Public/chunk-cli/internal/gitremote"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
@@ -427,8 +429,31 @@ func printInitSummary(commands []config.Command, streams iostream.Streams) {
 	}
 }
 
+// detectOrgID attempts to resolve a CircleCI org ID without prompting for auth.
+// Auto-selects when the user belongs to exactly one org; shows a picker when
+// interactive with multiple orgs. Skipped gracefully on auth failure, no TTY,
+// or cancellation — none of these are fatal for chunk init.
+func detectOrgID(ctx context.Context, rc config.ResolvedConfig, streams iostream.Streams, cfg *config.ProjectConfig) {
+	client, err := authprompt.ResolveCircleCIClient(rc, nil)
+	if err != nil {
+		if !errors.Is(err, authprompt.ErrNeedsAuth) {
+			streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not detect org ID: %v", err)))
+		}
+		return
+	}
+	orgID, err := orgPicker(ctx, client)()
+	if err != nil {
+		if !errors.Is(err, tui.ErrNoTTY) && !errors.Is(err, tui.ErrCancelled) {
+			streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not detect org ID: %v", err)))
+		}
+		return
+	}
+	cfg.OrgID = orgID
+	streams.ErrPrintf("Org ID: %s\n", ui.Bold(orgID))
+}
+
 func newInitCmd() *cobra.Command {
-	var force, skipHooks, skipGitHook, skipValidate, skipCompletions, skipSkills, skipTestSuites bool
+	var force, skipHooks, skipGitHook, skipValidate, skipCompletions, skipSkills, skipTestSuites, skipOrgID bool
 	var projectDir string
 
 	cmd := &cobra.Command{
@@ -495,9 +520,13 @@ hook config files.`,
 				streams.ErrPrintf("Detected repository: %s\n", ui.Bold(fmt.Sprintf("%s/%s", org, repo)))
 			}
 
+			rc, rcErr := config.Resolve("", "", insecureStorage)
+			if rcErr != nil {
+				streams.ErrPrintf("%s\n", ui.Warning(fmt.Sprintf("Could not load config: %v", rcErr)))
+			}
+
 			// Step 2: Validate command detection
 			if !skipValidate {
-				rc, _ := config.Resolve("", "", insecureStorage)
 				claude, _ := anthropic.New(anthropic.Config{APIKey: rc.AnthropicAPIKey, BaseURL: rc.AnthropicBaseURL})
 				commands, detectErr := validate.DetectCommands(ctx, claude, workDir)
 				if detectErr != nil {
@@ -517,6 +546,11 @@ hook config files.`,
 				}
 			}
 
+			// Step 3: CircleCI org ID
+			if !skipOrgID && cfg.OrgID == "" {
+				detectOrgID(ctx, rc, streams, cfg)
+			}
+
 			// Save config
 			if err := config.SaveProjectConfig(workDir, cfg); err != nil {
 				return &userError{
@@ -527,7 +561,7 @@ hook config files.`,
 			}
 			streams.ErrPrintln(ui.Success("Wrote .chunk/config.json"))
 
-			// Step 3: Write hook config files for supported agents.
+			// Step 4: Write hook config files for supported agents.
 			if !skipHooks {
 				if err := writeAllHookFiles(workDir, cfg.Commands, streams); err != nil {
 					return err
@@ -539,12 +573,12 @@ hook config files.`,
 				}
 			}
 
-			// Step 4: Shell completions
+			// Step 5: Shell completions
 			if !skipCompletions {
 				maybeInstallCompletions(streams)
 			}
 
-			// Step 5: CircleCI Smarter Testing test-suites.yml
+			// Step 6: CircleCI Smarter Testing test-suites.yml
 			if skipTestSuites {
 				printTestSuitesHint(workDir, streams)
 			} else {
@@ -553,7 +587,7 @@ hook config files.`,
 				}
 			}
 
-			// Step 6: Agent skills
+			// Step 7: Agent skills
 			if !skipSkills {
 				installSkillsStep(workDir, streams)
 			}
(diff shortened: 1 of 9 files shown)


hanabel1 (approved):
i think this could use some automated tests tho for (detectOrgID and orgPicker)

schurchleycci:
> i think this could use some automated tests tho for (detectOrgID and orgPicker)

Good call, I'll add these before merging

internal/cmd/init.go
```diff
@@ -450,8 +452,28 @@ func writeGitHook(gitCommonDir string, streams iostream.Streams) error {
 	return nil
 }
 
+// detectOrgID attempts to resolve a CircleCI org ID without prompting for auth.
+// Auto-selects when the user belongs to exactly one org; shows a picker when
+// interactive with multiple orgs. Skipped gracefully on auth failure, no TTY,
+// or cancellation — none of these are fatal for chunk init.
+func detectOrgID(ctx context.Context, rc config.ResolvedConfig, streams iostream.Streams, cfg *config.ProjectConfig) {
+	client, err := authprompt.ResolveCircleCIClient(rc, nil)
```
danmux, line 437: `ResolveCircleCIClient` returns `ErrNeedsAuth` for missing creds, but a genuine failure out of `circleci.NewClient` (bad base URL, etc.) gets swallowed here too. Gate the silent return on `errors.Is(err, authprompt.ErrNeedsAuth)` and warn otherwise?

internal/cmd/init.go
```diff
@@ -540,6 +562,16 @@ hook config files.`,
 				}
 			}
 
+			// Step 3: CircleCI org ID
```
danmux, line 549: Step numbering collides — there's already a `// Step 3: Write hook config files` further down.

internal/cmd/init.go
```diff
@@ -540,6 +562,16 @@ hook config files.`,
...
+			if !skipOrgID && cfg.OrgID == "" {
+				rc, err := config.Resolve("", "", insecureStorage)
```
danmux: Step 2 does `rc, _ := config.Resolve(...)` and ignores the error; the identical failure here prints a warning. Resolve once above both steps and handle the error consistently?

internal/cmd/org.go
```diff
@@ -15,11 +16,64 @@ func newOrgCmd() *cobra.Command {
...
 	cmd.AddCommand(newOrgCreateCmd())
+	cmd.AddCommand(newOrgListCmd())
+	return cmd
+}
+
+func newOrgListCmd() *cobra.Command {
+	var jsonOut bool
+
+	cmd := &cobra.Command{
+		Use:   "list",
```
danmux, line 31: No coverage for `org list` — table output, `--json`, and the empty case are all untested. `org_test.go` only exercises create.

internal/cmd/org.go
```diff
@@ -15,11 +16,64 @@ func newOrgCmd() *cobra.Command {
...
+		Short: "List CircleCI organizations",
+		Long:  "List CircleCI organizations the authenticated user belongs to.\n\nUseful for finding your org ID to pass as --org-id or store with 'chunk config set orgID <id>'.",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			insecureStorage, _ := cmd.Flags().GetBool("insecure-storage")
+			rc, _ := config.Resolve("", "", insecureStorage)
+			io := iostream.FromCmd(cmd)
+
+			client, err := ensureCircleCIClient(cmd.Context(), cmd, rc, io, tui.PromptHidden)
+			if err != nil {
+				return &userError{
```
danmux: This throws away the `userError` `ensureCircleCIClient` already built (code `auth.circleci_token_required`, `ExitAuthError`) and relabels oauth/network failures as auth failures. Returning `err` unwrapped keeps the exit code — matters more now the skill branches on "error means not authenticated".

internal/cmd/org.go
```diff
@@ -15,11 +16,64 @@ func newOrgCmd() *cobra.Command {
...
+					msg:        "Could not list organizations.",
+					suggestion: "Check your network connection.",
+					err:        fmt.Errorf("list collaborations: %w", err),
+				}
+			}
+
+			if jsonOut {
+				enc := json.NewEncoder(io.Out)
```
danmux: Every other JSON path goes through `iostream.PrintJSON` (sidecar.go:180, config.go:72) — any reason to hand-roll the encoder here? Also a nil `collabs` encodes as `null` rather than `[]`, which is awkward for anything parsing it.

internal/cmd/sidecar_test.go
```diff
@@ -55,6 +59,95 @@ func TestSnapshotCreateNameTooLong(t *testing.T) {
...
+		called = true
+		return "picker-org", nil
+	})
+	assert.NilError(t, err)
+	assert.Equal(t, got, "from-flag")
+	assert.Assert(t, !called, "pickOrg should not be called when flag is set")
+}
+
+func TestResolveOrgID_ConfigOrgID(t *testing.T) {
+	t.Setenv(config.EnvCircleCIOrgID, "env-org")
```
danmux, line 74: Name says ConfigOrgID but this sets the env var — same path as `TestResolveOrgID_FallsBackToPickOrg`'s sibling. The `.chunk/config.json` branch of `config.ResolveOrgID` is still uncovered.

skills/chunk-sidecar/SKILL.md
```diff
@@ -41,11 +42,10 @@ Do **not** run `echo $CIRCLE_TOKEN`, `env`, `printenv`, or any other command tha
...
-   3. If **both** are unset, **stop** — do not call `chunk sidecar create`, `chunk sidecar list`, or any other `chunk sidecar` command. Ask the user for their org ID **exactly once** using this message:
-
-      > No CircleCI orgID found in `.chunk/config.json`. Sidecar cannot select an org in a non-interactive session. Please provide your CircleCI org ID. To persist it for future sessions, I will run: `chunk config set orgID <id>`
-
-      After the user provides the ID, run `chunk config set orgID <id>`, then continue to Step 2.
+   3. If **both** are unset, run `chunk org list` to discover available orgs:
+      - **Single org returned** — use that ID automatically: `chunk config set orgID <id>`, then continue to Step 2.
+      - **Multiple orgs returned** — show the list to the user and ask which org to use **exactly once**. After they reply, run `chunk config set orgID <id>`, then continue to Step 2.
+      - **Error or empty list** — the user is not authenticated or has no orgs. Stop and ask them to run `chunk auth set circleci` first.
```
danmux: Empty list exits 0 with the warning on stderr, so the agent sees success with no rows rather than an error. Point this step at `chunk org list --json` so there's something deterministic to parse?

internal/cmd/init.go
```diff
@@ -450,8 +452,28 @@ func writeGitHook(gitCommonDir string, streams iostream.Streams) error {
...
+// detectOrgID attempts to resolve a CircleCI org ID without prompting for auth.
+// Auto-selects when the user belongs to exactly one org; shows a picker when
+// interactive with multiple orgs. Skipped gracefully on auth failure, no TTY,
+// or cancellation — none of these are fatal for chunk init.
+func detectOrgID(ctx context.Context, rc config.ResolvedConfig, streams iostream.Streams, cfg *config.ProjectConfig) {
+	client, err := authprompt.ResolveCircleCIClient(rc, nil)
+	if err != nil {
+		return
+	}
+	orgID, err := orgPicker(ctx, client)()
```
danmux, line 444: Why go via the collaborations picker at all here? Step 1 has already detected the VCS org/repo from the git remote, and `GET /api/v3/orgs?filter[slug]=<provider>/<org>` resolves a slug straight to the org UUID (public-api-service `v3/api.go:118`). `circleci-cli` does exactly this in `internal/apiclient/org.go:54` (`ResolveOrgID`), as does `circleci/mcp` (`client/pas/org.go` `GetOrgBySlug`).

That would be deterministic rather than a guess: no picker, no TTY dependency, correct for multi-org users, and it picks the org that actually owns *this* repo instead of the first collaboration in the list. The empty-list case maps to "not on CircleCI" and can stay a soft skip.

`internal/circleci/projects.go:59` `GetProjectBySlug` also already returns `org_id` and is currently unused — either route gets there without a picker.

danmux: On the plumbing cost: v3 responses are JSON:API-enveloped (`{"data": [...], "page": {...}}`), and `chunk-cli`'s only v3 usage today is the sidecar surface, which isn't. `circleci-cli` already has the generic for it — `internal/apiclient/client.go:154`:

```go
type v3List[T any] struct {
	Data []T `json:"data"`
	Page struct {
		Next *string `json:"next"`
		Prev *string `json:"prev"`
	} `json:"page"`
}
```

plus `v3Entity[T]` for single-resource responses (`client.go:150`), and `filterParam(key, val)` → `filter[key]=val` (`client.go:179`). `ResolveOrgID` is then three lines on top of that.

Worth copying the same two generics + `filterParam` into `internal/circleci` rather than hand-unwrapping `data` at the call site — every future v3 endpoint needs them, and `internal/httpcl` already has `QueryParam` (`request.go:66`) so there's nothing else to build.

schurchleycci: > Why go via the collaborations picker at all here? Step 1 has already detected the VCS org/repo from the git remote, and `GET /api/v3/orgs?filter[slug]=<provider>/<org>` resolves a slug straight to the org UUID (public-api-service `v3/api.go:118`). `circleci-cli` does exactly this in `internal/apiclient/org.go:54` (`ResolveOrgID`), as does `circleci/mcp` (`client/pas/org.go` `GetOrgBySlug`).
> 
> That would be deterministic rather than a guess: no picker, no TTY dependency, correct for multi-org users, and it picks the org that actually owns _this_ repo instead of the first collaboration in the list. The empty-list case maps to "not on CircleCI" and can stay a soft skip.
> 
> `internal/circleci/projects.go:59` `GetProjectBySlug` also already returns `org_id` and is currently unused — either route gets there without a picker.

Does that work for standalone orgs or just classic ones?

PR #469, opened 2026-07-31
Be helpful when chunk validate isn't configured

schurchleycci:
## Summary

- When `chunk validate` is run with no commands configured, the error now points users toward `chunk init` and explains it installs the `chunk-sidecar` skill so their AI coding agent can pick up from there
- Suppressed the redundant detail line that was repeating the error message

## Test plan

- [ ] `TestValidateNoConfigShowsSkillHint` — asserts the suggestion mentions both `chunk init` and `chunk-sidecar`
- [ ] Run `chunk validate` in a project without `.chunk/config.json` and confirm the output reads cleanly

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -233,7 +233,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		}
 		return &userError{
 			msg:        "No validate commands configured.",
-			suggestion: "Run 'chunk init' first.",
+			suggestion: suggestionValidateNotConfigured,
 			errMsg:     "no validate commands configured",
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/cmd/validate.go
```diff
@@ -699,11 +700,15 @@ func sidecarAutoName(ctx context.Context, workDir string) string {
 	return base + "-validate"
 }
 
+const suggestionValidateNotConfigured = "Run 'chunk init' to detect and configure validation commands.\n" +
+	"This also installs the chunk-sidecar skill so your AI coding agent can help you set up remote validation on a sidecar."
```
michael-webster: Let's be explicit and include the skill name to invoke

michael-webster (approved):
approved mod wording change

PR #472, opened 2026-08-03
Track chunk sidecar setup success

hanabel1:
- Adds a chunk_sidecar_setup telemetry event to chunk sidecar setup with setup-specific context: detected stack, number of setup steps, whether the environment was cached, and outcome
- The generic chunk_command_invocation event (emitted for every command) only records success/failure and duration — this event lets us slice success rate by stack and see whether cached vs. fresh environment detection correlates with failures

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -19,6 +19,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/gitutil"
 	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
 	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
+	"github.com/CircleCI-Public/chunk-cli/internal/telemetry"
 	"github.com/CircleCI-Public/chunk-cli/internal/tui"
 	"github.com/CircleCI-Public/chunk-cli/internal/ui"
 )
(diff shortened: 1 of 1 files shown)


schurchleycci:
A few minor things:

1. setup_steps counts defined steps, not completed steps — real issue, low priority

The name is genuinely ambiguous, but whether it matters depends on what the metric is for. Looking at the code, setup_steps is clearly measuring configuration complexity (how many setup steps does this project have?), not execution progress. It's paired with stack and env_cached — all "about the project," not "about this run." For that use case the current value is correct. The rename to setup_step_total would be a legitimate clarification, but it's a documentation/naming issue rather than a bug.

2. Environment detection failures are invisible — real issue, worth fixing

This one is real and non-obvious. If DetectEnvironment fails at line 963–970, the function returns early and the chunk_sidecar_setup telemetry call at line 1047 is never reached. The event simply doesn't fire. There's no comment flagging this as intentional. If the metric is tracking the success rate of chunk sidecar setup runs, it's silently undercounting failures. Even if intentional (e.g. "we only want to track the setup steps phase"), it should be documented.

3. Missing comment on test step exclusion — real issue, easy fix

The filter at line 1043 has no explanation, but the exact same filter at line 1200 has: // test step is for Dockerfile CMD only, not for SSH execution. The comment already exists in the project — it just needs to be added to the counting loop too. This is a concrete gap that will confuse a future reader.

Summary: Items 2 and 3 are genuine gaps worth addressing. Item 1 is real but a judgment call on naming preference rather than a correctness issue.

schurchleycci (approved):

PR #475, opened 2026-08-05
completion: write static file instead of spawning subprocess on every shell open

michael-webster:
## Summary

- `chunk completion install` now generates the completion script once and writes it to `~/.config/chunk/completion.{zsh,bash}`, sourcing that static file from the rc file instead of `source <(chunk completion zsh)`
- `chunk upgrade` regenerates the static file after a successful upgrade so completions stay in sync with new commands/flags
- `chunk completion uninstall` removes both the rc file entry and the static file; handles the old subprocess-style source line gracefully for existing installs

## Motivation

`source <(chunk completion zsh)` was running `chunk completion zsh` as a subprocess on every new shell session, generating ~400 telemetry events per day from shell startups rather than intentional user invocations.

## Test plan

- [ ] `task acceptance-test` — all completion acceptance tests pass with updated assertions for static file source line
- [ ] `task test` — unit tests in `internal/cmd` pass
- [ ] Manual: `chunk completion install` writes `~/.config/chunk/completion.zsh` and sources it from `.zshrc`
- [ ] Manual: open a new shell — no `chunk` subprocess spawned
- [ ] Manual: `chunk upgrade` followed by checking `~/.config/chunk/completion.zsh` is updated

### DIFF
--- a/acceptance/completion_test.go
+++ b/acceptance/completion_test.go
@@ -43,44 +43,34 @@ func TestCompletionInstallBash(t *testing.T) {
 	env := testenv.NewTestEnv(t)
 	env.Extra["SHELL"] = "/bin/bash"
 
-	bashrc := filepath.Join(env.HomeDir, ".bashrc")
-	err := os.WriteFile(bashrc, []byte("# bashrc\n"), 0o644)
-	assert.NilError(t, err)
-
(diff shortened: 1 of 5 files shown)


michael-webster:
Closing in favor of a different approach — will package completion scripts to land in the right place during installation, following the circleci-cli pattern.

PR #481, opened 2026-08-06
Add project-scope skill installation

michael-webster:
## Summary

- Introduces a `Scope` enum (`ScopeUser`, `ScopeProject`) to the `internal/skills` package — all callers now pass a scope explicitly
- **User scope** (default): existing behavior unchanged — installs into `~/.claude/skills` and `~/.agents/skills`, skipping agents whose config dirs don't exist
- **Project scope**: installs into `.claude/skills` and `.agents/skills` relative to the current working directory, creating directories as needed (no pre-existing agent config required)
- Adds `--scope user|project` flag to both `skill install` and `skill list` (default: `user`)

## Reproduction commands

```bash
# User scope (existing behavior, unchanged)
chunk skill install
chunk skill install --scope user

# Project scope — installs into ./.claude/skills and ./.agents/skills
chunk skill install --scope project

# List with project scope (agents always shown as available)
chunk skill list --scope project

# Invalid scope returns a clear error
chunk skill install --scope global
# → error: invalid scope: must be 'user' or 'project'
```

## Test plan

- [x] All existing acceptance tests pass (user scope unchanged)
- [x] `TestSkillsInstallProjectScope` — installs to project dir, leaves user home untouched
- [x] `TestSkillsInstallProjectScopeUpToDate` — idempotent on second run
- [x] `TestSkillsInstallProjectScopeNotIsolatedFromUserScope` — user and project installs coexist
- [x] `TestSkillsListProjectScope` — list shows missing before install, current after
- [x] `TestSkillsInstallInvalidScope` — unknown scope returns non-zero exit + clear error
- [x] Unit tests for `ScopeProject` in `internal/skills/skills_test.go`

### DIFF
--- a/acceptance/skills_test.go
+++ b/acceptance/skills_test.go
@@ -286,3 +286,106 @@ func TestSkillsListMixedStates(t *testing.T) {
 	assert.Assert(t, strings.Contains(combined, "missing"),
 		"expected 'missing' for deleted skill, got: %s", combined)
 }
+
+func TestSkillsInstallProjectScope(t *testing.T) {
+	env := testenv.NewTestEnv(t)
+	projectDir := t.TempDir()
+
+	// Project scope requires no pre-existing agent dirs.
+	result := binary.RunCLI(t, []string{"skill", "install", "--scope", "project"}, env, projectDir)
+
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
+	combined := result.Stdout + result.Stderr
+	assert.Assert(t, strings.Contains(combined, "claude:"),
+		"expected per-agent output for claude, got: %s", combined)
+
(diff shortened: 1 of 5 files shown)


schurchleycci:
Approved mod some Claude findings:

1. install.go:172 — Write failures in installForAgent are silently swallowed via continue, returning an empty result that the cmd layer prints as "all skills up to date" with exit 0. With --scope project, the old ConfigDir existence guard no longer gates execution, making this more reachable.
2. install.go:205 — The identical os.Stat+os.IsNotExist probe wrapped in SkipIfAbsent is duplicated in Status; a shared agentAvailable(agent) bool helper would remove the maintenance hazard.
3. acceptance/skills_test.go:289 — TestSkillsInstallProjectScope only asserts .claude/skills/ was written; .agents/skills/ is never checked. A broken Codex project-scope install would pass acceptance tests undetected.

schurchleycci (approved):

PR #484, opened 2026-08-07
Fix PreToolUse hook matcher so pre-commit validation actually fires

michael-webster:
## Summary

- **Root cause**: `"Bash(git commit*)"` as the `matcher` on a `PreToolUse` hook group was silently broken. Claude Code evaluates `matcher` against the tool name using exact string comparison or JS regex — not command content. That string is treated as a regex requiring characters after `"Bash"`, so it never matched the tool name `"Bash"` and the pre-commit hook **never fired**.
- **Fix**: set `matcher: "Bash"` on the hook group (exact tool-name match) and move the git-commit filter to `if: "Bash(git commit*)"` on each individual hook entry, which is where command-content filtering belongs per the hook spec.
- **Migration**: `mergeHooks` now recognises the legacy `"Bash(git commit*)"` group matcher in existing `settings.json` files and replaces it in-place, so re-running `chunk init` upgrades existing projects without leaving a duplicate group.

## Test plan

- [ ] `go test ./internal/settings/ -race` — all existing + new tests pass
- [ ] `go test ./... -race` — full suite green
- [ ] `go tool golangci-lint run ./internal/settings/` — 0 issues
- [ ] New `TestBuildHookMatcherIsToolName` asserts generated matcher is `"Bash"` and entries carry `if: "Bash(git commit*)"`
- [ ] New `TestMergeHooksMigratesLegacyMatcher` asserts existing files with old matcher are upgraded to new format with no duplicate groups


michael-webster:
## Verification steps

Build and install the branch locally first (symlinks `dist/chunk` into `~/.local/bin`):

```bash
cd /path/to/chunk-cli
task dev-install
```

---

### 1. Verify the generated settings.json shape

Create a temp project with a `Taskfile.yml` so `chunk init` detects commands and generates hooks:

```bash
TESTDIR=$(mktemp -d) && cd "$TESTDIR"

git init
git remote add origin git@github.com:example/myrepo.git

cat > Taskfile.yml << 'EOF'
version: "3"
tasks:
  test:
    cmds:
      - echo PRE_COMMIT_RAN
  lint:
    cmds:
      - echo LINT_RAN
EOF

chunk init --skip-git-hook --skip-completions --skip-skills --force
```

**Assert with jq:**
```bash
jq '.hooks.PreToolUse[0].matcher' .claude/settings.json          # → "Bash"
jq '.hooks.PreToolUse[0].hooks[0].if' .claude/settings.json      # → "Bash(git commit*)"
```

---

### 2. Verify migration: old settings.json is upgraded without a duplicate group

```bash
cat > .claude/settings.json << 'EOF'
{
  "$schema": "https://json.schemastore.org/claude-code-settings.json",
  "_comment": "Generated by chunk init.",
  "permissions": {"allow": ["Bash(chunk:*)"]},
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash(git commit*)",
        "hooks": [
          {"type": "command", "command": "cd ${CLAUDE_PROJECT_DIR:-.} && task test", "timeout": 300}
        ]
      }
    ]
  }
}
EOF

chunk init --skip-git-hook --skip-completions --skip-skills --force

jq '.hooks.PreToolUse | length' .claude/settings.json            # → 1  (no duplicate)
jq '.hooks.PreToolUse[0].matcher' .claude/settings.json          # → "Bash"
jq '.hooks.PreToolUse[0].hooks[0].if' .claude/settings.json      # → "Bash(git commit*)"
```

---

### 3. Verify the hook fires on a commit and not on unrelated Bash calls

```bash
claude   # open Claude Code in $TESTDIR
```

Inside the Claude Code session:

**Should fire — expect `PRE_COMMIT_RAN` in hook output:**
```
Bash(git commit --allow-empty -m "test: verify hook")
```

**Should NOT fire — no hook output:**
```
Bash(ls -la)
Bash(echo hello)
```

schurchleycci (approved):

PR #494, opened 2026-08-11
update mutation testing skill

jesseworld22:
Adds chunk validate variants for running mutation-testing variants across parallel ephemeral sidecars, with a non-racing SyncEphemeral path and an updated testing-gaps skill.

### DIFF
--- a/acceptance/validate_variants_test.go
+++ b/acceptance/validate_variants_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/testing/binary"
 	testenv "github.com/CircleCI-Public/chunk-cli/internal/testing/env"
 	"github.com/CircleCI-Public/chunk-cli/internal/testing/fakes"
+	"github.com/CircleCI-Public/chunk-cli/internal/testing/gitrepo"
 	"github.com/CircleCI-Public/chunk-cli/internal/testing/recorder"
 	"github.com/CircleCI-Public/chunk-cli/internal/variants"
 )
@@ -158,7 +159,9 @@ func TestValidateVariantsNamedCommand(t *testing.T) {
 	env := testenv.NewTestEnv(t)
 	env.CircleCIURL = srv.URL
 
-	workDir := env.HomeDir
+	// A git repo with an origin remote, so the remote workspace resolves to the
+	// same <sidecarHome>/<repo> the rest of the CLI syncs into.
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
 	writeChunkConfig(t, workDir, nil)
 	path := writeVariantsFile(t, workDir, []variants.Variant{
 		{ID: "MUT-001"},
@@ -207,7 +210,7 @@ func TestValidateVariantsCreatesAndDeletesSidecars(t *testing.T) {
 	env := testenv.NewTestEnv(t)
 	env.CircleCIURL = srv.URL
 
-	workDir := env.HomeDir
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
 	writeChunkConfig(t, workDir, nil)
 	path := writeVariantsFile(t, workDir, []variants.Variant{
 		{ID: "MUT-001"},
@@ -244,7 +247,7 @@ func TestValidateVariantsImageFromConfig(t *testing.T) {
 	env := testenv.NewTestEnv(t)
 	env.CircleCIURL = srv.URL
 
-	workDir := env.HomeDir
+	workDir := gitrepo.SetupGitRepo(t, "my-org", "my-repo")
 	writeChunkConfig(t, workDir, map[string]interface{}{
 		"validation": map[string]interface{}{
 			"sidecarImage": "snap-from-config",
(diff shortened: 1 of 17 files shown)


acceptance/validate_variants_test.go
```diff
@@ -0,0 +1,271 @@
...
+		// No --image flag; should use config value.
+	}, env, workDir)
+
+	reqs := cci.Recorder.AllRequests()
+	creates := filterVariantRequests(reqs, "POST", "/api/v3/sidecar/instances")
+	assert.Assert(t, len(creates) >= 1, "expected at least 1 create request")
+
+	var body map[string]interface{}
+	assert.NilError(t, json.Unmarshal(creates[0].Body, &body))
+	assert.Equal(t, body["image"], "snap-from-config",
```
danmux: Is this one currently failing? `CreateSidecar` sends the V3 envelope (`{"data":{"attributes":{"name":..,"image":..}}}`), so `body["image"]` on the top-level map is nil. Ran it at this SHA:

```
--- FAIL: TestValidateVariantsImageFromConfig (0.08s)
    validate_variants_test.go:269: assertion failed: <nil> (body["image"] <nil>) != snap-from-config (string)
```

The other nine `TestValidateVariants*` pass, as does `./internal/variants/...` with `-race`. Would asserting via `cci.Sidecars[0].Image` be simpler, the way `internal/cmd/sidecar_test.go:42` does it?

internal/cmd/validate_variants.go
```diff
@@ -0,0 +1,182 @@
...
+			// sidecars the reaper will never collect, because variant sidecars
+			// are deliberately absent from the active-sidecar file.
+			statusFn := newStatusFunc(streams)
+			if swept := variants.SweepOrphans(ctx, client, resolvedOrgID, statusFn); swept > 0 {
+				statusFn(iostream.LevelInfo, fmt.Sprintf("swept %d orphaned variant sidecar(s) from a previous run", swept))
+			}
+
+			cmdStrings := make([]string, len(cmds))
+			for i, c := range cmds {
+				cmdStrings[i] = c.Run
```
danmux: Raw `c.Run` skips template expansion, doesn't it? The existing remote path runs `expandCommand(workDir, c.Run)` in `validate.RunRemote` to substitute `{{CHANGED_PACKAGES}}`, and this repo's own config has `"run": "task test -- {{CHANGED_PACKAGES}}"`. Since `--name` bypasses the `remote` filter, `--name test-changed` would ship the literal `{{CHANGED_PACKAGES}}` to the sidecar, exit non-zero for a shell reason, and `runVariant` sets `Killed = true`. Every mutant then reads as caught.

Would reusing `validate.RunRemote` beat re-implementing the exec loop here? It already handles expansion, escaping, per-command reporting and the `timeout` field (which is also dropped on this path, so a runaway mutant holds a billed sidecar and a parallel slot indefinitely).

internal/cmd/validate_variants.go
```diff
@@ -0,0 +1,182 @@
...
+func resolveVariantsWorkspace(ctx context.Context, workdirFlag, projectDir string) string {
+	if workdirFlag != "" {
+		return workdirFlag
+	}
+	if active, err := sidecar.LoadActive(ctx); err == nil && active != nil && active.Workspace != "" {
+		return active.Workspace
+	}
+	_, repo, err := gitremote.DetectOrgAndRepo(projectDir)
+	if err == nil && repo != "" {
+		return "./workspace/" + repo
```
danmux: Should this default match `sidecar.ResolveWorkspace`? Everywhere else the remote workspace is `<sidecarHome>/<repo>`, and `chunk sidecar env build` provisions dependencies in that directory before the snapshot is taken. With no active sidecar file (fresh clone, agent worktree, CI) this returns `./workspace/<repo>`, so the sync clones into a directory the snapshot never prepared, the build fails environmentally, and `Killed = true` again. That also undercuts the `validation.sidecarImage` requirement SKILL.md now insists on.

`sidecar.ResolveWorkspace(ctx, workdir, repo)` already implements the same three-step priority with the right default.

danmux (commented):
Three things worth a look before merge. Two of them share a failure mode: any non-zero exit sets `Killed = true`, so an environmental failure gets reported as a caught mutant, and the skill reports a clean bill of health. That's the one direction this tool must not fail in.

internal/variants/variants.go
```diff
@@ -0,0 +1,204 @@
...
+// SweepOrphans deletes sidecars in the org left behind by an earlier variants
+// run. It is a backstop, not the primary cleanup: Run deletes its own sidecars
+// and the signal handler covers interrupts, but a crash or a lost network
+// connection can still strand one, and nothing else will ever collect it.
+//
+// Only names carrying namePrefix are touched. Deletion failures are reported
+// and skipped rather than aborting the sweep — a stale entry the API refuses to
+// delete must not block the run the caller actually asked for.
+func SweepOrphans(ctx context.Context, client *circleci.Client, orgID string, status iostream.StatusFunc) int {
+	existing, err := client.ListSidecars(ctx, orgID, false)
```
danmux, line 373: Can this sweep tell an orphan from a live sidecar? It deletes every `variant-*` sidecar the caller owns, with no age or ownership guard beyond `all=false`. Two `chunk validate variants` runs by the same user, in two worktrees or two repos, is a normal shape for an agent-driven skill, and the second run's pre-flight sweep would delete the first run's in-flight sidecars. Those variants then die mid-test and get reported as `error`, so it degrades to inconclusive rather than a false pass, but it is still a run silently destroyed by an unrelated one.

`sidecar.Reap` handles the same hazard by sparing anything touched inside `StaleAfter`, on the reasoning that "a concurrent session on another branch legitimately owns its own sidecar". Could the name carry a run ID or a start timestamp so the sweep only takes names it can prove are stale?

Related: `TestRunDeletesEverySidecarOnCancel` may not be guarding the leak it describes. The context is cancelled before `Run` is called, so `sidecar.Create` fails on the dead context before any request is sent. Instrumented at this SHA:

```
creates=0 deletes=0 total_requests=0
result MUT-001 err="cancelled before start"
result MUT-002 err="create sidecar: ... giving up after 1 attempt(s): context canceled"
result MUT-003 err="cancelled before start"
```

So `assert.Equal(t, creates, deletes)` is comparing `0` with `0`, and would still pass with the deferred `DeleteSidecar` removed entirely. Cancelling from inside the fake's create handler on the first POST, then asserting `creates >= 1 && creates == deletes`, would actually pin it.

acceptance/validate_variants_test.go
jesseworld22: fixed:
- parses nested envelope and asserts on `body.Data.Attributes.Image`
- didn't use `cci.Sidecars[0]` - run deletes its own sidecar 
- added a check that the sidecar name starts with `variant-` + moved test into a real git repo

internal/cmd/validate_variants.go
jesseworld22: fixed:

- commands now go through `validate.ExpandCommand(workDir, c.Run)` before being sent
- timeout is now honored per command with a `--timeout flag` as the fallback
- a timeout is recorded as an error (rather than a kill)
- `Killed = true` also runs errors
- didn't reuse `validate.RunRemote` 
- `RunRemote` doesn't handle `timeout` either - only the local `runCommand` path does

internal/cmd/validate_variants.go
jesseworld22: initially, my function had its own private copy of the "where does code go on the remote machine?" logic - and its fallback pointed at the wrong folder.


fixed:

1. deleted the duplicate logic and called the share function.
2. it now returns an error instead of guessing.
3. kept the flag as an escape hatch
4. a missing git remote isn't treated as fatal on its own

internal/variants/variants.go
jesseworld22: fixed:

- names now carry a timestamp
- sweep only deletes names older than `orphanAfter` 
- command timeouts now bound a variant's lifetime well under 2h

the test:
- correct: it was comparing 0 with 0
- now it cancels from inside the create handler on the first POST, then asserts creates >= 1 && creates == deletes
- creates >= 1 is the load-bearing half - without it the test can rot back to 0-vs-0
- asserts cancelled variants aren't reported as `Killed` and carry an error

.github/agents/chunk-cli-debugger.agent.md
```diff
@@ -0,0 +1,68 @@
+---
```
danmux: did you mean to commit this?

jesseworld22: good catch - dropped the commit

michael-webster (approved):

PR #497, opened 2026-08-12
preserve unknown config keys in ‘chunk config set’

jesseworld22:
`chunk config set` rewrote `.chunk/config.json` through a typed struct and silently dropped every field it didn't model -including keys hand-added to `commands[].` config writes now merge onto the existing file, preserving unrecognized keys verbatim while still letting the struct delete the ones it owns.


internal/config/config.go
```diff
@@ -13,22 +13,66 @@ import (
...
+// Callers must therefore load before saving, or they will drop keys the file
+// already had. A missing or unparseable file falls back to a plain marshal; a
+// caller that must not overwrite an unparseable config checks for that first,
+// via LoadProjectConfigForUpdate.
+func marshalPreserving(path string, cfg any) ([]byte, error) {
+	data, err := marshalCompact(cfg)
+	if err != nil {
+		return nil, err
+	}
+	if existing, readErr := os.ReadFile(path); readErr == nil {
```
hanabel1: tldr: save reads the file, merges new values in and then writes it back. if the file is broken json at read, it skips the merge and overwrites with only the fields the code knows about. the project config path has a guard for it and the user config path doesn't. the fix is to either add the same guard for the user config, or make the silent swallowing of the error into an actual error that gets returned. 

_Claude comment:_ The project config path is protected: callers use LoadProjectConfigForUpdate, which refuses to proceed on a malformed file. But Save(UserConfig) has no equivalent guard. If the user config becomes unparseable between the Load() call and the marshalPreserving re-read (or if it's already malformed when a user-config config set runs), Merge returns ErrInvalidJSON, the error is swallowed, and the file is overwritten with only struct-modeled keys — silently dropping every unknown field. This directly contradicts the PR's stated invariant.

Suggested fix — surface the error or add a user-config guard:

```
if existing, readErr := os.ReadFile(path); readErr == nil {
    if merged, mergeErr := jsonmerge.Merge(data, existing, cfg); mergeErr == nil {
        data = merged
    } else if !errors.Is(mergeErr, jsonmerge.ErrInvalidJSON) {
        return nil, fmt.Errorf("merge existing config: %w", mergeErr)
    }
    // ErrInvalidJSON: file is corrupt; fall back to plain marshal (callers
    // must guard against this via LoadProjectConfigForUpdate / equivalent).
}


```Alternatively, add a LoadUserConfigForUpdate that mirrors the project config guard and call it before Save in the user-config branch of newConfigSetCmd.

Comment (jesseworld22) [reply internal/config/config.go]: from claude:  

Fixed in 0194007: Merge error is no longer swallowed, so config. Save refuses an unparseable user config and leaves the file untouched instead of flattening it to the modeled keys. 

Guarded at the write rather than adding a LoadUserConfigForUpdate, since Load already has that contract and a wrapper would still miss the file breaking between the load and the re-read. The one exception is SaveProjectConfig, which still replaces a broken file via a new marshalOverwriting because chunk init -- force depends on that; both behaviors are now tested and documented in ARCHITECTURE.md.

File: internal/jsonmerge/jsonmerge.go [hunk 2]
```diff
@@ -0,0 +1,313 @@
...
+	// Keys the type does not model are the user's, not chunk's to rewrite.
+	for _, m := range existingMembers {
+		if _, modeled := fields[m.key]; !modeled {
+			out = append(out, m)
+		}
+	}
+	return encodeObject(out)
+}
+
+func mergeArray(typed, existing json.RawMessage, elem *node) json.RawMessage {
```
hanabel1: mergeArray pairs elements by "name". An existing element without that key is never stored in byName, so its unknown keys are silently dropped with no warning. collectUnknown does the opposite — it recurses into every element regardless of name — so UnknownKeys reports keys pre-save that Merge won't actually preserve.

Real-world impact is low since unnamed commands are non-functional, but the two functions should agree on the behavior. Easiest fix: have collectUnknown skip unnamed elements to match what Merge does.

jesseworld22: fixed:

1. `collectUnknown` now skips array elements without a `name` -> agrees with `mergeArray`.

 2.  added `TestUnknownKeysSkipsUnnamedArrayElements` which asserts both halves.
      - the unnamed element's unknown key isn't reported.
      - `Merge` confirms the unnamed `fileExt` is dropped.

acceptance/config_test.go
```diff
@@ -330,6 +330,93 @@ func TestConfigSetOrgID(t *testing.T) {
...
+    {"name": "format", "run": "task fmt", "fileExt": "go", "always": true},
+    {"name": "lint", "run": "task lint", "fileExt": "go", "limit": 5},
+    {"name": "test", "run": "task test", "fileExt": "go", "limit": 5},
+    {"name": "build", "run": "task build", "fileExt": "go", "always": false}
+  ],
+  "myTool": {"nested": [1, 2]}
+}`
+	assert.NilError(t, os.WriteFile(configPath, []byte(original), 0o644))
+
+	result := binary.RunCLI(t, []string{"config", "set", "validation.sidecarImage", "snap-1"}, env, workDir)
```
michael-webster, line 386: you can use the same pattern to set an unknown config key instead of manually inserting it into the json like you are doing on line 349.

`binary.RunCLI(t, []string{"config", "set", "madeUpConfig", "madeup value"}, env, workDir)`

acceptance/config_test.go
```diff
@@ -330,6 +330,93 @@ func TestConfigSetOrgID(t *testing.T) {
...
+	assert.Equal(t, result.ExitCode, 0, "stdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+
+	data, err := os.ReadFile(configPath)
+	assert.NilError(t, err)
+
+	var cfg map[string]json.RawMessage
+	assert.NilError(t, json.Unmarshal(data, &cfg))
+	_, kept := cfg["myTool"]
+	assert.Assert(t, kept, "unknown top-level key was dropped: %s", data)
```
michael-webster: i don't think this assert is testing what you want here, i'd expect the value to still be present when you read back the config.

michael-webster (changes requested):

acceptance/config_test.go
jesseworld22: tried that first - config set only writes modeled keys and exits 1 with Unknown config key: `madeUpConfig` - so unknown keys only reach the file by hand-editing or another tool writing it.

 added `TestConfigSetRejectsUnknownKey` in e76937f to pin the rejection.

acceptance/config_test.go
jesseworld22: fixed in e76937f: it now compares `myTool` against what was written, and asserts each command`'s `fileExt` is still `"go"` rather than merely present.

internal/cmd/validate.go
```diff
@@ -73,15 +75,38 @@ func detectHook(r io.Reader) *hookContext {
 
 func runValidateList(workDir string, jsonOut bool, streams iostream.Streams, statusFn iostream.StatusFunc) error {
 	cfg, err := config.LoadProjectConfig(workDir)
+	unreadable := err != nil && errors.Is(err, config.ErrParseProjectConfig)
 	if err != nil {
+		// A config that does not parse has commands we cannot see, so say so
```
michael-webster: test

internal/cmd/config.go
```diff
@@ -153,13 +165,14 @@ func newConfigSetCmd() *cobra.Command {
...
 			}
 
 			if !config.ValidConfigKeys[key] {
 				return &userError{
 					msg:    fmt.Sprintf("Unknown config key: %q.", key),
-					detail: "Supported keys: model, useSSHIdentityFile, telemetry, orgID, validation.sidecarImage.",
+					detail: fmt.Sprintf("Supported keys: %s, %s.", configKeyList(config.ValidConfigKeys), configKeyList(config.ValidProjectConfigKeys)),
 					errMsg: fmt.Sprintf("unknown config key %q", key),
 				}
 			}
```
michael-webster, line 184: test-178

internal/cmd/config.go
```diff
@@ -207,7 +220,58 @@ func newConfigSetCmd() *cobra.Command {
...
```
michael-webster, line 228: test-220

internal/cmd/config.go
```diff
@@ -124,8 +126,18 @@ func newConfigSetCmd() *cobra.Command {
 		Use:   "set <key> <value>",
 		Short: "Set a config value",
 		Long:  "Set a config value. Use 'chunk auth set <provider>' to store credentials with validation.\n\nUser keys: model, useSSHIdentityFile, telemetry\nProject keys: orgID, validation.sidecarImage",
```
michael-webster: test-128

internal/cmd/config.go
michael-webster, line 184: The project-config path a few lines above uses `malformedProjectConfigError(err)` to surface a parse failure with a fix hint. The user-config path here returns the raw error, so a truncated `~/.config/chunk/config.json` produces "An unknown error occurred" with no actionable message. `errors.Is(err, config.ErrParseUserConfig)` + `malformedProjectConfigError(err)` would match the pattern established above.

internal/cmd/validate.go
michael-webster: When `jsonOut` is false this `ErrPrintf` fires here, and then the `if unreadable` block below fires too — the user sees a raw error line followed by the user-friendly warning and fix suggestion for the same condition. The `--json` path doesn't hit the `if unreadable` block so it's internally consistent, but the text path emits three lines for one error. Dropping this `ErrPrintf` call and letting `statusFn` carry both paths would de-duplicate.

internal/cmd/config.go
michael-webster: Error messages for unknown keys are now generated dynamically via `configKeyList(config.ValidConfigKeys)` and stay accurate automatically. This `Long` help text keeps a hand-written parallel copy — adding a key to the maps will silently diverge the help output. Could use the same `configKeyList` calls to generate this string, or at minimum a test that asserts the keys listed here match the map contents.

acceptance/init_test.go
```diff
@@ -185,6 +185,100 @@ func TestInitForcePreservesSkippedSections(t *testing.T) {
...
+	env.AnthropicKey = ""
+
+	result := binary.RunCLI(t, []string{
+		"init", "--skip-hooks", "--skip-validate",
+	}, env, workDir)
+
+	assert.Assert(t, result.ExitCode != 0,
+		"expected non-zero exit for malformed config\nstdout: %s\nstderr: %s", result.Stdout, result.Stderr)
+	combined := result.Stdout + result.Stderr
+	assert.Assert(t, strings.Contains(combined, ".chunk/config.json"),
```
michael-webster: Several new tests in this file and `sidecar_gaps_test.go` use `assert.Assert(t, strings.Contains(...))` rather than `assert.Check(t, cmp.Contains(...))` used by nearby tests. Two differences: `assert.Assert` short-circuits on first failure (masking later assertions), and `cmp.Contains` formats a diff showing the actual string on failure. Acceptance tests are slow to iterate, so the diagnostic quality matters.

internal/config/project.go
```diff
@@ -125,8 +223,36 @@ func (c *ProjectConfig) FindCommand(name string) *Command {
 	return nil
 }
 
-// SaveProjectConfig writes the config back to .chunk/config.json.
+// SaveProjectConfig writes the config back to .chunk/config.json. Keys the file
+// had that ProjectConfig does not model ride along in the Extra fields and are
+// written back out, so callers must load before saving: a config built from
+// scratch has no unknown keys to carry and would drop the ones on disk.
+//
+// An existing file that does not parse is refused with ErrParseProjectConfig and
```
michael-webster, line 231: The parse-guard here reads the on-disk file into a throwaway probe, then overwrites with the caller-supplied `cfg`. If any caller ever passes a freshly constructed `ProjectConfig` (empty `Extra`), the guard passes and all hand-added keys are silently dropped — defeating the PR's purpose. Every current caller loads first, but the signature doesn't enforce that.

A structural fix: replace `SaveProjectConfig` with a function that owns the roundtrip itself, accepting a mutation fn instead of a complete struct:

```go
func UpdateProjectConfig(workDir string, fn func(*ProjectConfig)) error {
    cfg, err := LoadProjectConfigForUpdate(workDir)
    if err != nil {
        return err
    }
    fn(cfg)
    return writeProjectConfigAtomic(workDir, cfg)
}
```

Callers change from load-then-pass to:

```go
config.UpdateProjectConfig(workDir, func(cfg *config.ProjectConfig) {
    cfg.OrgID = value
})
```

Unknown keys are always preserved because the function owns the roundtrip — a caller structurally cannot skip the load. `LoadProjectConfigForUpdate` becomes an internal detail.

jesseworld22:
1. closed the two remaining paths that silently dropped unknown config keys - chunk init replacing a config it couldn't read, and sidecar env/setup flattening the environment block.

2. moved the validate --list reporting above the --json branch instead of deleting the extra print, because deleting it would have made a broken config completely silent in JSON mode.

internal/cmd/sidecar.go
```diff
@@ -11,6 +10,8 @@ import (
 	"regexp"
 
 	petname "github.com/dustinkirkland/golang-petname"
+	json "github.com/go-json-experiment/json"
```
michael-webster, line 13: are we not able to just use jsonv2 directly?

internal/config/config.go
```diff
@@ -1,32 +1,87 @@
...
+		return false, fmt.Errorf("write %s: %w", name, err)
+	}
+	// Durability before the rename, so a crash cannot leave the renamed file
+	// pointing at unflushed contents.
+	if err := f.Sync(); err != nil {
+		_ = f.Close()
+		return false, fmt.Errorf("sync %s: %w", name, err)
+	}
+	// CreateTemp makes the file 0o600; widen it only where the caller asks.
+	if err := f.Chmod(perm); err != nil {
```
michael-webster, line 65: what's this for?

acceptance/config_test.go
```diff
@@ -330,6 +363,128 @@ func TestConfigSetOrgID(t *testing.T) {
 	assert.Equal(t, cfg["orgID"], "org-xyz")
 }
 
+// The reported bug, end to end: writing one key must not strip keys chunk does
+// not model, on commands or at the top level.
+func TestConfigSetPreservesUnknownProjectKeys(t *testing.T) {
```
michael-webster, line 368: nit (optional): this test verifies preservation by checking individual substrings, which takes ~60 lines and leaves gaps (e.g. a key that survives but with a mangled value would pass). A whole-document JSON comparison would be more compact and more precise — unmarshal the output into a `map[string]json.RawMessage`, compare against the known-good document using `assert.DeepEqual`, and the entire preservation story fits in ~15 lines. Something like:

```go
var got, want map[string]json.RawMessage
assert.NilError(t, json.Unmarshal(data, &got))
assert.NilError(t, json.Unmarshal([]byte(expected), &want))
assert.DeepEqual(t, got, want)
```

Not a blocker — the current approach is correct and thorough. Just worth considering if the test gets touched again.

PR #499, opened 2026-08-12
Resume empty output streams instead of failing on them

schurchleycci:
## Summary

- Fix `streamCommandOutput` to treat an empty SSE stream (200 OK, no frames, no error) as a resume trigger rather than a stall, so commands that produce no output while running don't get abandoned prematurely
- Update the stall counter to increment only on genuine connectivity errors (`err != nil`), and skip the backoff delay when `err == nil || stalls == 0`
- Add `EmptyStreamsBeforeExit` to the fake CircleCI server and a corresponding test (`TestExecKeepsResumingOnEmptyStreams`) to cover this case

## Test plan

- [ ] `task test` passes with `-race`
- [ ] `TestExecKeepsResumingOnEmptyStreams` exercises the empty-stream path against the fake, confirming many consecutive empty responses don't exhaust the stall budget
- [ ] Existing resume tests (`TestExecKeepsResumingWhileServerStaysAlive`, `TestExecGivesUpAfterTooManyStalls`) still pass unchanged

### DIFF
--- a/internal/circleci/client.go
+++ b/internal/circleci/client.go
@@ -309,13 +309,13 @@ func (c *Client) streamCommandOutput(ctx context.Context, commandID string, onOu
 		}
 
 		// A connection that delivered any frame proves the far end is alive and
-		// talking, so it resets the stall count even when the command produced no
-		// new output — a command can legitimately be silent for minutes while it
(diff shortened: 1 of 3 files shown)


danmux (approved):
no clue

PR #500, opened 2026-08-12
merge the stop hook into an existing settings.json

jesseworld22:
## description

`Merge` handled `permissions` and `PreToolUse` but never called `mergeStopHooks.` Just `MergeCodex` did.

So `chunk init` on a repo that **already had** a `.claude/settings.json` merged its commit hooks and dropped the `Stop` hook on the floor. The repo ended up looking fully configured while nothing ran `chunk validate` at session end. Every `Stop` assertion in the suite was on the `MergeCodex` path, which is how this survived.

## changes since last review

The fix started as one line in `Merge` calling the existing `mergeStopHooks`. Self-review found that landing it that way traded one silent-data-loss bug for another, so `mergeHooks` now owns hook **entries** rather than hook **groups**. That is a bigger diff than "one line," and it changes two behaviors beyond the original bug:

- **PreToolUse no longer deletes a user's own Bash hooks.** `CommitMatcher` is the bare tool name `Bash`, not a command glob, so any team with a non-commit `Bash` PreToolUse hook shares that group with chunk. Replacing the group wholesale — which is what the old code did, and what the one-line fix would have kept doing for PreToolUse — dropped their entry. Chunk's entries are identified by `CommitIfFilter` instead, and a group still on the legacy matcher is treated as chunk's in full since its entries predate the `if` tag.
- **Duplicate `chunk validate` entries now collapse.** The old `mergeStopHooks` replaced the whole group, which incidentally deduped it. Replacing a single entry in place does not, so stale duplicates would have survived every subsequent merge and run validation twice per session. Every owned entry is stripped before the generated one is reinserted, so re-running `init` converges.

Ownership is now symmetric across both hook types and both the Claude and Codex paths share it, which is what let `MergeCodex` diverge from `Merge` in the first place.

## tests

`internal/settings/merge_test.go` covers the merge itself: adding `Stop` to an existing file, replacing chunk's entry in place, leaving a user's own `Stop` group alone, preserving a user entry that shares chunk's group, the same for a user entry in the `Bash` PreToolUse group, duplicate collapse across groups and matchers, and idempotency on each shared-group path.

On the acceptance test asked for in review — one is added (`TestInitWritesFiringStopHook`, plus a real Stop-hook assertion on the `settings.example.json` fallback), but it is worth being explicit that **it does not catch this bug**, and I verified that by reintroducing the defect: the acceptance test still passes and only `internal/cmd` `TestWriteSettingsExistingMergeApplied` fails.

The reason is structural. `writeSettings` only calls `Merge` when a `.claude/settings.json` already exists, and that path is gated behind `tui.Confirm`, which returns `ErrNoTTY` under a test harness and falls back to writing `settings.example.json` from the generated bytes. So a fresh-repo `init` never reaches `Merge`, and the one acceptance-reachable existing-settings path never reaches it either. Covering the merge from an acceptance test needs either a PTY dependency (none in the tree today — `kr/pty` is only an indirect `go.sum` hash) or a non-interactive apply flag. Happy to add either if that is the bar; flagging it rather than implying the acceptance test guards the fix.

The actual regression guard is at the `cmd` layer, where the write happens. That test previously asserted only `merged["hooks"] != nil`, which is exactly why the bug shipped: `hooks` was non-nil from the merged commit hooks while `Stop` was absent. It now asserts per hook type.

## scope change since first review

This PR originally also added `chunk hook repair`, Stop-hook reporting in `chunk validate` and `chunk hook status`, and a `validation.requireStopHook` opt-out. Per review, that general-purpose repair tooling is out of scope - the ticket was to find the affected repos and fix them directly. All of it has been dropped; what's left is the root-cause bug fix.

## affected repos (fixed directly)

What I did: Swept all 594 non-archived, non-fork repos in `circleci` and `CircleCI-Public` via the GitHub GraphQL API for `.chunk/config.json`, then checked each for a firing Stop hook. **9 repos have a chunk config; all 9 are sidecar-configured; 5 had no firing Stop hook.** PRs are open on each:

| Repo | Gap | PR |
|---|---|---|
| `CircleCI-Public/circleci-cli` | had settings.json, no `Stop` block — this bug | CircleCI-Public/circleci-cli#1739 |
| `circleci/opentelemetry-collector-deploy` | no `.claude/settings.json` | circleci/opentelemetry-collector-deploy#773 |
| `circleci/public-api-service` | no `.claude/settings.json` | circleci/public-api-service#1236 |
| `circleci/factory-bot` | no `.claude/settings.json` | circleci/factory-bot#170 |
| `circleci/sandbox-provisioner` | no `.claude/settings.json` | circleci/sandbox-provisioner#368 |

Already healthy: `soc-integrations`, `notifications`, `llmops-service`, `chunk-cli`.

Note that only `circleci-cli` is explained by this merge bug — the other four never had a `.claude/settings.json` written at all, which is a separate gap.

### DIFF
--- a/internal/settings/merge.go
+++ b/internal/settings/merge.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"sort"
 
 	udiff "github.com/aymanbagabas/go-udiff"
@@ -24,6 +25,11 @@ const CommitIfFilter = "Bash(git commit*)"
 // of chunk init. Recognised during merge so existing settings can be migrated
 // to the current format without leaving a stale duplicate group behind.
 const legacyCommitMatcher = "Bash(git commit*)"
+
+// StopCommand is the Stop hook command that chunk manages. Merge identifies
+// chunk's own Stop entry by this exact string, so it must stay in sync with the
+// command written by Build and BuildCodex.
+const StopCommand = "chunk validate"
 
 // MergeResult holds the computed merge without performing any I/O.
 type MergeResult struct {
@@ -179,28 +185,28 @@ func mergeHooks(merged, generated map[string]interface{}) {
 	}
 
 	// Replace existing group with same matcher (or legacy matcher), or append.
-	replaced := false
-	for i, g := range mergedPreToolUse {
-		group, isMap := g.(map[string]interface{})
-		if !isMap {
-			continue
-		}
-		matcher, _ := group["matcher"].(string)
-		if matcher == CommitMatcher || matcher == legacyCommitMatcher {
-			mergedPreToolUse[i] = chunkGroup
-			replaced = true
-			break
-		}
-	}
-	if !replaced {
-		mergedPreToolUse = append(mergedPreToolUse, chunkGroup)
-	}
-
-	mergedHooks["PreToolUse"] = mergedPreToolUse
+	mergedHooks["PreToolUse"] = replaceOrAppend(mergedPreToolUse, isChunkCommitGroup, chunkGroup)
 }
 
-// mergeStopHooks replaces the chunk-managed Stop hook group (identified by the
-// "chunk validate" command) within Stop, preserving all other Stop groups.
+// isChunkCommitGroup reports whether a PreToolUse group is the chunk-managed
+// commit group, accepting the legacy matcher so older settings migrate in place.
+func isChunkCommitGroup(g interface{}) bool {
+	group, ok := g.(map[string]interface{})
+	if !ok {
+		return false
+	}
+	matcher, _ := group["matcher"].(string)
+	return matcher == CommitMatcher || matcher == legacyCommitMatcher
+}
+
+// mergeStopHooks installs chunk's Stop hook entry into hooks.Stop, preserving
+// every entry chunk does not own.
+//
+// Chunk owns a single entry (identified by StopCommand), not a whole group. A
+// user may have added their own entries to that same group, so the entry is
+// replaced in place and its siblings are left alone — replacing the enclosing
+// group would silently delete them. Only when no chunk entry exists anywhere is
+// chunk's own group appended.
 func mergeStopHooks(merged, generated map[string]interface{}) {
 	genHooks, ok := generated["hooks"].(map[string]interface{})
 	if !ok {
@@ -211,15 +217,19 @@ func mergeStopHooks(merged, generated map[string]interface{}) {
 		return
 	}
 
-	// Find the chunk-managed group in generated Stop hooks.
-	var chunkGroup interface{}
+	// Find chunk's group in the generated Stop hooks, and the entry within it.
+	var chunkGroup, chunkEntry interface{}
 	for _, g := range genStop {
-		if isChunkStopGroup(g) {
-			chunkGroup = g
+		_, entries, isGroup := stopGroupEntries(g)
+		if !isGroup {
+			continue
+		}
+		if i := slices.IndexFunc(entries, isChunkStopEntry); i >= 0 {
+			chunkGroup, chunkEntry = g, entries[i]
 			break
 		}
 	}
-	if chunkGroup == nil {
+	if chunkEntry == nil {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


internal/settings/merge.go
schurchleycci, line 1: A few issues from Claude:

  ● 270 [correctness]    isChunkStopGroup returns true for any group containing a 'chunk validate' entry, even if the group also has user-owned entries; mergeStopHooks then replaces the entire group, silently deleting those entries.
  ● 270 [correctness]    The 'chunk validate' command string is a bare literal with no named constant, unlike CommitMatcher and CommitIfFilter which are declared constants.
  ● 182 [simplification] The replace-or-append loop is copy-pasted verbatim in both mergeHooks (lines 182–199) and mergeStopHooks (lines 239–251), differing only in the slice variable and the match predicate.

internal/settings/merge_test.go
schurchleycci, line 1: And a few more:

 ● 282 [test-coverage]  TestMergeReplacesStopHook never asserts result.Changed, but the production caller in init.go gates the file write on result.Changed.
  ● 335 [test-coverage]  TestMergeNoChangeWhenAlreadyMerged has no Stop hook in its fixture, leaving Merge() Stop-hook idempotency completely untested.
  ● 310 [test-coverage]  TestMergePreservesUserStopHooks asserts only that two Stop groups are present, not that the chunk group actually received the generated content.

internal/settings/merge.go
jesseworld22: fixed:

- group replacement: chunk owns an entry rather than a group. `isChunkStopGroup` removed; `mergeStopHooks` swaps chunk's entry in place, user siblings untouched. new test + verified with a real chunk init.
- bare literal: added `StopCommand` (`merge.go:32`), used by the matcher and `Build/BuildCodex.` Left `init.go:373` (the git hook script).
- duplicated loop: extracted generic `replaceOrAppend[T]` (`merge.go:289`), used by both call sites.

internal/settings/merge_test.go
jesseworld22: fixed:

- `TestMergeReplacesStopHook`:  now asserts `result.Changed`, plus entry count and command so a stacked duplicate can't pass either.
- `TestMergeNoChangeWhenAlreadyMerged`: added a Stop hook to the fixture. Also added `TestMergeNoChangeWhenChunkStopEntryHasUserSibling`. the single-entry fixture wouldn't catch a reorder from the new in-place swap.
 - `TestMergePreservesUserStopHooks`: now asserts content of both groups (user's command/timeout preserved, chunk's group carries the generated values).

michael-webster (commented):
Can you add an acceptance test for this that confirms the behavior.

internal/settings/merge.go
```diff
@@ -73,6 +79,11 @@ func Merge(existing, generated []byte) (*MergeResult, error) {
 	// Merge hooks.PreToolUse — replace the chunk-managed hook group by matcher.
 	mergeHooks(merged, generatedMap)
 
+	// Merge hooks.Stop — replace the chunk-managed group by command. Without
+	// this a repo that already had a settings.json keeps its commit hooks but
+	// never gets the Stop hook, so validation stops running at session end.
+	mergeStopHooks(merged, generatedMap)
```
michael-webster: i don't think you need a new function to merge the stop hooks, merge hooks just wasn't handling the case probably. i'd extend merge hooks to also include the stop hook merging as well.

jesseworld22: from claude:

done in adbf765 - `mergeStopHooks` is gone and `mergeHooks` handles both hook types, so `Merge` and `MergeCodex` make one call instead of two.

kept the difference in how they merge, since it is not incidental: chunk owns the whole PreToolUse group (matched by matcher) but only a single Stop *entry* (matched by command), because a user may have their own entries in that same group and replacing the group would delete them. That is now documented on the one function instead of split across two, and the duplicated "create hooks if absent" block collapses into a `hooksMap` helper.

no behavior change — the existing tests pin both halves independently: six exercise generated settings carrying only PreToolUse, six only Stop, three both.

jesseworld22:
claude summary on acceptance test:

added `TestInitWritesFiringStopHook` in acceptance/init_test.go, + a real Stop-hook assertion on the settings.example.json fallback (that test previously only checked the file existed).

being straight about this though: the acceptance test does not catch this bug. checked by reintroducing the defect - it still passes, only internal/cmd `TestWriteSettingsExistingMergeApplied` fails.

it's structural. writeSettings only calls Merge when a .claude/settings.json already exists, and that path is gated behind tui.Confirm → ErrNoTTY → falls back to writing settings.example.json. so fresh-repo init never reaches Merge, and the one acceptance-reachable existing-settings path doesn't either. getting there needs a PTY dep (none in the tree) or a --yes flag. both easy, both new surface area on a PR you already asked me to shrink — tell me which and I'll add it.

the real guard is the cmd-layer test. it previously asserted only merged["hooks"] != nil, which is why this shipped: hooks was non-nil from the merged commit hooks while Stop was missing. now asserts per hook type.

jesseworld22:
added `TestInitWritesFiringStopHook` in acceptance/init_test.go, + a real Stop-hook assertion on the settings.example.json fallback (that test previously only checked the file existed).

the acceptance test does not catch this bug.

checked by reintroducing the defect - acceptance test still passes, and only internal/cmd `TestWriteSettingsExistingMergeApplied` fails.

schurchleycci (approved):

PR #501, opened 2026-08-13
sidecar: incremental working-tree sync to preserve golangci-lint caches

michael-webster:
## Summary

- **Pure no-op path**: when HEAD commit and working-tree patch are identical to last sync, all SSH operations are skipped. Lint drops from ~57 s to ~1.6 s in the steady-state dev loop.
- **Delta sync path**: when only the working-tree changed (no new commits), reverse the previously applied patch then apply the new one — only files in the patches are touched, committed `.go` files are left alone and lint caches stay valid.
- **Full reset+clean+apply**: unchanged fallback for new commits, fresh sidecars, or delta failures.

## State additions

- `LastSyncedPatchHash` in `sidecar.json` — SHA-256 hex digest of the last applied patch for quick equality checks.
- `sidecar[.session[-hash]].diff` alongside `sidecar.json` — full patch content for the delta reverse step, written `0o600` (patch may contain sensitive code).
- Patch file helpers now take an explicit `repoRoot` derived from the `BundleSync` `cwd` arg so branch lookup is always correct regardless of process working directory.

## Test plan

- [ ] `go test -race ./internal/sidecar/...` passes
- [ ] `TestBundleSync_NoOpSkipsBundle` — verifies no reset/apply sent on second identical sync
- [ ] `TestBundleSync_DeltaSkipsResetOnPatchChange` — verifies delta path skips reset when only working-tree changed
- [ ] `TestBundleSync_NoPatchFileUsesFullSync` — verifies fallback to reset when patch file is missing
- [ ] Full suite: `task test`

### DIFF
--- a/internal/sidecar/active.go
+++ b/internal/sidecar/active.go
@@ -26,6 +26,11 @@ type ActiveSidecar struct {
 	OrgID         string `json:"org_id,omitempty"`
 	Workspace     string `json:"workspace,omitempty"`
 	LastSyncedRef string `json:"last_synced_ref,omitempty"`
+	// LastSyncedPatchHash is a SHA-256 hex digest of the working-tree patch
+	// last applied to the sidecar. BundleSync uses it to skip the
+	// reset+clean+apply cycle when neither commits nor working-tree have
+	// changed since the previous sync.
+	LastSyncedPatchHash string `json:"last_synced_patch_hash,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


michael-webster:
Confirmed this skips full syncs with incremental changes by reapplying patches instead of the full bundle.

schurchleycci (approved):

PR #504, opened 2026-08-17
Don't write any files that are not needed for the current repo

hanabel1:
https://linear.app/circleci/issue/FACT-414/dont-write-any-files-that-are-not-needed-for-the-current-repo-eg-codex

## Summary

- Migrates .claude/settings.json from the old "matcher": "Bash(git commit*)" group format to "matcher": "Bash" with per-entry "if" filters, matching how Claude Code actually evaluates hook matchers
- Adds a File Conventions section to AGENTS.md telling agents not to create config files for other AI tools (CODEX.md, GEMINI.md, etc.) unless the project uses them

## Why

The legacy matcher used a glob on the tool name, which Claude Code evaluates differently from the per-entry if filter. The current format is the correct way to filter by command content within a tool match.

The AGENTS.md guidance replaces a hook-based approach that was too blunt — a hook can't know whether the project uses Codex; behavioral guidance in context can.

## Test plan

- [ ] task test passes
- [ ] Run chunk init in a test repo and confirm .claude/settings.json is generated with the new format


.claude/settings.json
```diff
@@ -9,24 +9,37 @@
...
           {
             "type": "command",
+            "if": "Bash(git commit*)",
             "command": "cd ${CLAUDE_PROJECT_DIR:-.} && chunk validate test",
             "timeout": 300
           }
         ]
+      },
+      {
+        "matcher": "Write",
```
michael-webster: Hmm, what's the reasoning on this change? I don't think we want to stop claude from writing to these locations. We probably don't want to write Codex specific configuration if the user isn't using codex but I don't think we want to block the agents from writing to files.

michael-webster (changes requested):
I'm not sure on the intent on this. I think we don't want to create a bunch of dot folders for agents the user isn't using but I don't think we want to block agents from writing to files.

.claude/settings.json
hanabel1: Ah okay yeah I'll remove that

PR #505, opened 2026-08-17
Add validate --mark-remote to route commands to the sidecar

danmux:
## Summary

Stacked on #497 - review and merge that one first. This branch's diff is the last commit only.

- `chunk validate --mark-remote [name]` sets `remote: true` on one command, or on every configured command when no name is given, then exits without running anything.
- Until now the only thing that set the flag was `chunk sidecar setup`, and only for `install` and gate-role commands. Anything else - or a sidecar the `chunk-sidecar-setup` skill brought up through its manual path, which marks nothing - left hand-editing `.chunk/config.json` as the only option.
- Commands already marked report no change. An unknown name errors and does not touch the file. A config that does not parse is refused via `LoadProjectConfigForUpdate` rather than replaced.
- The write goes through `SaveProjectConfig`, so the unknown keys #497 preserves - including ones hand-added to a command - survive being marked. There's a test for exactly that.
- `chunk-sidecar-setup` Stage 6 now tells the agent what to do when nothing ran on the sidecar, instead of leaving it stuck.

Unmarking is still a hand edit. A one-way flag is a bit of a trap, so shout if you want `--unmark-remote` in here too.

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -67,6 +67,7 @@ chunk
 │   --cmd <command>                 # Run an inline command
 │   --save                          # Save --cmd to config
 │   --remote                        # Run on the active sidecar
+│   --mark-remote                   # Mark [name] (or all commands) remote in config, then exit
 │   --sidecar-id <id>               # Remote execution in specific sidecar
 │   --org-id <id>                   # Organization ID (used when creating a new sidecar)
 │   --identity-file <path>          # SSH identity file for sidecar
@@ -171,6 +172,12 @@ chunk
   to disable.
 - `config set` user keys: `model`, `telemetry`. Project keys (`.chunk/config.json`):
   `orgID`, `validation.sidecarImage`. Credentials use `chunk auth set`, not `config set`.
+- `validate --mark-remote` sets `remote: true` on commands in `.chunk/config.json`
+  and exits without running anything. With a `[name]` it marks that one command,
+  without it every configured command. Commands already marked are reported as no
+  change. `chunk sidecar setup` marks install and gate commands automatically, so
+  `--mark-remote` is for the rest: a sidecar set up by hand, or a command whose
+  role does not qualify. Unmarking is still a hand edit of the config.
 - Telemetry is anonymous and opt-out. It's disabled by the
   `CHUNK_NO_TELEMETRY` / `NO_ANALYTICS` / `DO_NOT_TRACK` / `CI` environment
   variables (first match wins, in that order), or `chunk config set telemetry false`.
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -173,11 +173,20 @@ chunk
 - `config set` user keys: `model`, `telemetry`. Project keys (`.chunk/config.json`):
   `orgID`, `validation.sidecarImage`. Credentials use `chunk auth set`, not `config set`.
 - `validate --mark-remote` sets `remote: true` on commands in `.chunk/config.json`
-  and exits without running anything. With a `[name]` it marks that one command,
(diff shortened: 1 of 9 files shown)


schurchleycci:
Some findings:

[P1] Bare --mark-remote marks autofix commands, and the docs tell agents to run that form — internal/config/project.go:171, skills/chunk-sidecar-setup/SKILL.md:209, docs/GETTING_STARTED.md:208

MarkCommandRemote("") sets remote: true on every command, including role: autofix. Sidecar setup already refuses that via commandEligibleForSidecarRemote. Init still emits a format autofix command for Go/Taskfile repos.

The skills and getting-started docs lead with:

chunk validate --mark-remote           # mark every configured command
and then say to leave autofix local. Agents will copy the command, not the prose. Formatters then rewrite files on the sidecar and those edits never come back to the working tree — the failure sidecar setup was written to avoid.

Skip RoleAutofix on the unnamed path (named --mark-remote format can still override), and change the recommended invocation so “mark the rest” does not mean “mark format too.” The current TestMarkCommandRemote fixture already has format remote, so it would not catch this.

[P2] After setup writes sidecarImage, per-command routing is ignored — internal/cmd/validate.go:364, skills/chunk-sidecar-setup/SKILL.md:230, docs/GETTING_STARTED.md:213

New copy says chunk validate runs marked commands on the sidecar and the rest locally. That is only true when validation.sidecarImage is unset. This branch still does:


validate.go
Ln 364–366
if cfg.HasSidecarImage() {
	allRemote = true
}
Stage 8 of the same setup skill writes that key. From then on every command, including an unmarked formatter, runs remote. “Leave autofix local” cannot survive the flow the skill itself describes. Either stop forcing allRemote from the snapshot image, or stop documenting per-command routing as the post-setup default.

[P3] --list cannot show what Stage 6 tells agents to inspect — internal/validate/validate.go:61, skills/chunk-sidecar-setup/SKILL.md:209

Stage 6 says check what is marked with chunk validate --list, then mark the rest. List prints name: run only — no remote, no role. --list --json would include remote, but that is not what the skill runs. Agents following that step cannot tell which commands are already remote or which are autofix.

danmux:
All three land. Fixed in 83035bb.

**[P1]** Right, and the fixture point is the sharp bit - `format` was already `Remote: true`, so the assertion passed either way. Bare `--mark-remote` now skips `role: autofix` and prints what it left alone:

```
✓ Marked remote: test
  chunk validate               now runs these on the sidecar
  left local (autofix rewrites files): format
  mark one by name to override
```

`--mark-remote format` still marks it, for the caller who means it. Fixture reworked so `format` starts unmarked and `lint` carries the already-remote case, and there's a CLI test on the bare path.

**[P2]** Correct, and it predates this branch - `HasSidecarImage() → allRemote` has been on main since d66a51b (28 Jul). So I've taken the second of your two options and documented it rather than changing the routing, since that's a behaviour call beyond this PR. `docs/CLI.md`, `GETTING_STARTED.md`, and both skills now say per-command routing only applies while `validation.sidecarImage` is unset, and the setup skill says it at Stage 8 where it writes the key, not somewhere the agent has already scrolled past.

Worth flagging the underlying tension though: `MarkRemoteCommandsForSidecarSetup` carefully excludes autofix, then Stage 8 of the same flow makes that exclusion moot. Either `allRemote` should respect `role: autofix`, or per-command marking is dead weight for snapshot projects. Happy to take that on separately if you have a view on which way it should go.

**[P3]** Fixed - `--list` now tags each command:

```
test [remote, gate]: task test
format [local, autofix]: task fmt
bare [local]: echo hi
```

schurchleycci (approved):

PR #506, opened 2026-08-18
Add a code audit findings document

danmux:
## Summary

A point-in-time review of the repo for correctness, duplication and adherence to our own
conventions in `AGENTS.md` and `ARCHITECTURE.md`. Docs only, no code touched.

62 findings in `docs/CODE_AUDIT.md`, grouped and ordered by severity, each with the defect,
why it matters and the fix. Audited at `6bb97f9`; every `file:line` was re-derived against
that commit.

### The ones worth reading first

- **`internal/github` stacks its own retry on top of `httpcl`'s**, so a GitHub 5xx costs up to
  12 HTTP attempts and ~27s per GraphQL query. One field fixes it.
- **`main.go` reads `userError` through type assertions, not `errors.As`**, so wrapping one
  discards message, detail, suggestion, code and exit code. Confirmed by running it, not just
  reading it.
- **`gitutil` runs git against the process cwd** with no dir or ctx, so
  `--project /elsewhere` diffs the wrong repo on the `--checkout` sync path. `GeneratePatch`
  also mutates the user's index with a swallowed cleanup error.
- **Stack detection exists twice and the copies disagree**: `chunk init` writes `pytest` where
  the sidecar builds `uv run pytest`. Most likely of these to get reported as a bug by someone
  who has no idea two detectors exist.
- **The auth provider matrix is hand-written 18 times** across 3 files. `internal/cmd/auth.go`
  is 727 lines and almost all of it is this.
- **`envbuilder` has 31 stack-conditional branches across 3 parallel switches**, which is the
  root cause of both `nolint:gocyclo`. The doc proposes splitting by stack rather than by phase.

### Also in there

- What was checked and found **clean**, so none of it gets re-checked. Layering holds, output
  discipline holds, all 12 test skips are justified, `gitutil` is not over-exported.
- Why several of these got through: `dupl` is not enabled, `gocyclo` has no `min-complexity`,
  and `funlen` is excluded for tests but never enabled at all.
- A suggested order of work, cheapest and highest value first.

### Notes for review

- Deliberately **not** added to the documentation map in `AGENTS.md`. It is a point-in-time
  report that will go stale, not a standing reference like `ARCHITECTURE.md` or `CLI.md`.
- One finding (`validate --list` swallowing a config load error) is already addressed by #497
  and is marked as such so it does not get fixed twice.
- Findings marked `[swept]` came from a broad automated sweep and were not individually
  re-verified by hand. Everything else was confirmed by reading the code.

### DIFF
--- a/docs/CODE_AUDIT.md
+++ b/docs/CODE_AUDIT.md
@@ -1,0 +1,668 @@
+# Code audit
+
+A point-in-time review of `chunk-cli` for correctness, duplication and adherence to the
+conventions in [AGENTS.md](../AGENTS.md) and [ARCHITECTURE.md](ARCHITECTURE.md).
+
+**Audited at:** `6bb97f9` (main, after #505 merged). Every `file:line` below was re-derived
+against that commit. Line numbers drift — treat them as a starting point, not a contract.
+
+**Nothing in this audit has been fixed.** It is a work list.
+
+## How to read this
+
+Findings are grouped by kind and ordered by severity within each group. Each one gives the
+defect, why it matters, and the fix. Where a finding is a symptom of a deeper one it says so,
+so the same thing does not get fixed twice.
+
+Items marked **[swept]** came from a broad automated sweep and were not individually
+re-verified by hand. Everything else was confirmed by reading the code. Two findings were
+confirmed by running code rather than reading it, and say so explicitly.
+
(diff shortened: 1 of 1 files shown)


danmux:
Closing - superseded by ongoing work.

PR #508, opened 2026-08-18
Add remote vs local label to validate output and watch TUI

schurchleycci:
## Summary

- Validate output now includes `(local)` or `(remote)` on each individual command result line (e.g. `✓ lint  5.6s (remote)`), so users can tell at a glance which runner produced each result when commands are split across executors
- The validate summary line now shows `n/m passed  Xs` instead of `done in Xs`, making it immediately clear how many commands succeeded alongside the elapsed time
- The watch TUI now shows a local runner entry per project alongside remote sidecars — it displays pass/fail state and activity feed for local runs using the same event log infrastructure
- Event log recording is now wired for all runs (not just sidecar runs), writing events to the project data dir so both the TUI and local runs share the same log path

demo

demo-watch

## Test plan

- [ ] Run `chunk validate` locally — each command result shows `(local)` (e.g. `✓ lint  5.6s (local)`)
- [ ] Run `chunk validate` with a sidecar — each command result shows `(remote)`
- [ ] Summary line reads `n/m passed  Xs` (not `done in Xs`)
- [ ] Open `chunk watch` with no sidecars configured — local runner entry appears after first run, showing pass/fail
- [ ] Open `chunk watch` with sidecars — local and remote entries both appear; header sidecar count excludes the local entry
- [x] Existing tests pass: `task test`

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -405,7 +405,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 			// A hit is a success, so it finishes the same way a real successful run
 			// does rather than repeating that bookkeeping here: clearing the failure
 			// counter, and whatever else the success branch grows later.
-			return finishValidate(cmd, hook, nil, start, opts.sidecarID, cfg, wrapEventLogStatusFn(statusFn, "", nil, workDir, hook), streams)
+			return finishValidate(cmd, hook, nil, start, cfg, wrapEventLogStatusFn(statusFn, "", nil, workDir, hook), streams)
 		}
 	}
 
@@ -463,7 +463,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 			streams.ErrPrintf("  %s\n", ui.ErrDim(fmt.Sprintf("chunk validate: cache write failed: %v", err)))
 		}
 	}
-	return finishValidate(cmd, hook, execErr, start, opts.sidecarID, cfg, statusFn, streams)
+	return finishValidate(cmd, hook, execErr, start, cfg, statusFn, streams)
 }
 
 // loadEnvVarsWithRetry loads sidecar env vars, and if the sidecar is unusable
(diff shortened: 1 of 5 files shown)


internal/cmd/validate.go
```diff
@@ -303,7 +322,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 			// A hit is a success, so it finishes the same way a real successful run
 			// does rather than repeating that bookkeeping here: clearing the failure
 			// counter, and whatever else the success branch grows later.
-			return finishValidate(cmd, hook, nil, start, opts.sidecarID, cfg, statusFn, streams)
+			return finishValidate(cmd, hook, nil, start, cfg, &commandCounter{}, wrapEventLogStatusFn(statusFn, "", nil, workDir, hook), streams)
```
danmux: Cache-hit path passes a zeroed `&commandCounter{}` (passed=0, total=0) into `finishValidate`, which unconditionally formats `"%d/%d passed"`. A warm-cache run now prints `"0/0 passed  0.1s"` instead of anything meaningful.

The old code gated the sidecar status lines on `sidecarID != ""` — that gate was removed when `sidecarID` was dropped from the signature, and the cache-hit path has no equivalent guard.

Fix: populate `counter.total` from `cfg.Commands` (same as the non-cache path a few lines below), or skip the summary line entirely for cache hits.

internal/cmd/validate.go
```diff
@@ -48,6 +48,25 @@ func newStatusFunc(streams iostream.Streams) iostream.StatusFunc {
...
+// runner. It detects command results by the "(local)" or "(remote)" suffix that
+// RunRemote and runCommand append to every per-command status line.
+type commandCounter struct {
+	passed int
+	total  int
+}
+
+func (c *commandCounter) wrap(fn iostream.StatusFunc) iostream.StatusFunc {
+	return func(level iostream.Level, msg string) {
+		if strings.HasSuffix(msg, "(local)") || strings.HasSuffix(msg, "(remote)") {
```
danmux: `commandCounter.wrap` detects per-command outcomes by suffix-matching on status message strings produced in `internal/validate/validate.go`. Any rename of `"(local)"` or `"(remote)"` in that package silently zeros `counter.passed` for every run, producing `"0/N passed"` on full success — no compile error, no test failure.

The counter belongs in the `validate` package alongside the logic it observes, or `runValidate` should return a structured result (pass count, total) so the cmd layer doesn't have to parse message strings across the package boundary.

schurchleycci: Removed commandCounter entirely from cmd/validate.go. Instead, all Run* functions in the validate package now return (Result, error) where Result holds Passed and Total counts. The counts are computed right where the pass/fail is determined — no more suffix-matching on "(local)"/"(remote)" strings across a package boundary. runValidate and runSplitCommands propagate the Result up to finishValidate.

internal/cmd/validate.go
schurchleycci: Fixed by computing n := len(cfg.Commands) and passing validate.Result{Passed: n, Total: n} to finishValidate on the cache-hit path

danmux (approved):

PR #516, opened 2026-08-19
Add chunk prune command using the server-side prune endpoint

schurchleycci:
## Summary

- Adds a top-level `chunk prune` command that deletes the current user's sidecars via `POST /api/v3/sidecar/instances/prune`
- Uses the server-side prune endpoint (sandbox-provisioner#379, public-api-service#1286) rather than a list-then-delete loop — single round-trip, no need to enumerate IDs
- The API prunes caller-owned sidecars created more than 1 hour ago by default
- `--before <duration>` overrides the cutoff (e.g. `--before 24h`) by sending `scope.to` in the request body
- `--org-id` falls back to config/env/picker like other commands
- Reports the total deleted count; clears local active-sidecar state when any deletions occurred

Closes FACT-424.

prune

## Test plan

- [x] `chunk prune --help` shows correct usage and `--before` flag
- [x] `chunk prune` with sidecars present deletes them and clears active sidecar state
- [x] `chunk prune --before 24h` sends `scope.to` set to 24h ago
- [x] `chunk prune` with nothing to delete prints "No sidecars deleted"
- [x] `chunk sidecar delete` still works as before (single sidecar)


demo/prune.tape
```diff
@@ -0,0 +1,43 @@
+Output demo/prune.gif
```
danmux: accidentaly committed ?

danmux (approved):

danmux:
Are we dropping 'sidecar' ? 

It is a bit odd that it is 'chunk sidecar list' but 'chunk prune' ?

demo/prune.tape
schurchleycci: Ah good catch!

PR #517, opened 2026-08-19
[FACT-395] Extract chunk watch polling into a background daemon

michael-webster:
## What

Extracts the file I/O out of the watch TUI into a background daemon process.

Before this PR, `chunk watch` polled sidecar state files, event logs, and git
refs directly inside the BubbleTea model's `loadData` tick. This PR moves that
work into a separate `watchd` daemon process that runs in the background and
serves state over a Unix socket at `~/.chunk/watchd/watchd.sock`.

## How it works

`chunk watch` calls `watchd.EnsureRunning` before starting the TUI. If the
daemon isn't running, it launches `chunk watch _daemon` as a detached child
process (stdout/stderr go to `~/.chunk/watchd/watchd.log`) and waits up to 5s
for the PID file and socket to appear. If the daemon fails to start, `watch`
prints a warning but continues — the TUI just shows empty state until it can
connect.

The daemon (`watchd.RunDaemon`) calls `sidecar.AllProjectRoots` on a 5-second
tick to discover registered projects, then for each project reads sidecar JSON
files, tails the event log incrementally, and reads the current branch and HEAD
ref from git. It stores the processed state in memory and serves it over the
Unix socket. Connections are handled concurrently; reads use `sync.RWMutex`.

The TUI's `loadData` is now a single socket call: `watchd.FetchSnapshot(roots)`
returns a `Snapshot` with one `ProjectSnapshot` per root. `buildDataMsg` maps
that directly into the `dataMsg` the model already consumed. `ProjectEntry` now
only holds `ProjectRoot` — no event log handle, no data dir, no offset.

## Commits

1. **Add watch daemon infrastructure** — the `watchd` package minus the client: types, IPC wire format, Unix socket listener, poll loop, load helpers (`LoadSidecars`, `AnnotateActivity`, `SortByActivity`, `FilterSidecars`, `CapEvents`), and platform launch helpers.

2. **Add watch daemon client** — `FetchSnapshot`, `EnsureRunning`, `ping`, and `launchDaemon`. Separated so the infrastructure is reviewable on its own before the client that uses it.

3. **TUI: delegate polling to the watch daemon** — strips ~600 lines of file I/O from `model.go`, replaces `loadData` with the socket call, removes tests for the deleted helpers, adds `TestBuildDataMsg_*` for the new mapping layer.

## Test plan

- [ ] `go test ./internal/tui/watch/... ./internal/watchd/... -race`
- [ ] `task build && dist/chunk watch` — TUI launches, daemon starts, sidecars appear
- [ ] Kill daemon mid-session (`kill $(cat ~/.chunk/watchd/watchd.pid)`) — TUI shows empty state, recovers on next poll when daemon restarts


internal/tui/watch/client.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// loadFromDaemon is the default Model.loadFn. It fetches a snapshot from the
+// watch daemon and converts it to a dataMsg. If the daemon is unreachable it
+// is relaunched; the next tick will reconnect.
+func loadFromDaemon(m Model) tea.Msg {
+	msg, err := fetchFromDaemon(m)
+	if err != nil {
+		_ = watchd.EnsureRunning([]string{"watch", "_daemon"})
+		return dataMsg{}
```
schurchleycci: A fetch failure returns an empty dataMsg, and Update's dataMsg case (model.go:231-236) assigns all six fields unconditionally — so projects, sidecars, events, branches, and headRefs all get nil'd. One transient socket hiccup blanks the whole dashboard instead of holding the last good frame.

The _ = on EnsureRunning compounds it: there's also no signal about why the screen went empty. Could we keep prior state on failure and surface the reason — e.g. return an errMsg that sets a lastErr for the status line, and leave the data fields untouched?

internal/watchd/daemon.go
```diff
@@ -0,0 +1,215 @@
...
+	_ = os.Remove(sockPath)
+
+	ln, err := net.Listen("unix", sockPath)
+	if err != nil {
+		return fmt.Errorf("listen on %s: %w", sockPath, err)
+	}
+	defer func() { _ = ln.Close() }()
+	defer func() { _ = os.Remove(sockPath) }()
+
+	d := &daemon{projects: make(map[string]*projectState)}
```
schurchleycci, line 62: This only checks that some process holds the PID in the pidfile — it never checks that the socket answers. If the PID is live but the socket isn't usable (PID reuse after an unclean exit, or a permissions problem), EnsureRunning returns nil, the relaunch from the TUI is a no-op, and we retry forever against a daemon that will never respond.

ping already exists right above — could this be if running && ping(sockPath) { return nil }, with SocketPath() hoisted up? That makes "running" mean "actually serving".

schurchleycci (approved):

internal/tui/watch/client.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/CircleCI-Public/chunk-cli/internal/watchd"
+)
+
+// loadFromDaemon is the default Model.loadFn. It fetches a snapshot from the
+// watch daemon and converts it to a dataMsg. If the daemon is unreachable it
+// is relaunched; the next tick will reconnect.
+func loadFromDaemon(m Model) tea.Msg {
+	msg, err := fetchFromDaemon(m)
+	if err != nil {
+		_ = watchd.EnsureRunning([]string{"watch", "_daemon"})
```
danmux: EnsureRunning is called from a tea.Cmd goroutine on every poll tick. When the daemon is down, multiple ticks are in-flight concurrently — each goroutine races through the IsRunning check and calls launchDaemon independently, potentially spawning multiple child processes. Could the restart responsibility live entirely in the watch.go startup path, with the tick just returning dataMsg{} silently?

internal/watchd/daemon.go
```diff
@@ -0,0 +1,215 @@
...
+func (d *daemon) poll() {
+	roots, _ := sidecar.AllProjectRoots()
+
+	d.mu.RLock()
+	work := make([]*projectState, 0, len(roots))
+	newProjects := make(map[string]*projectState)
+	for _, root := range roots {
+		ps, ok := d.projects[root]
+		if !ok {
+			ps = d.initProject(root)
```
danmux: initProject does os.MkdirAll, os.OpenFile, and os.Stat inside the d.mu.RLock() block. Any writer waiting for the write lock in updateProject is stalled behind these file opens. Could initProject be called after releasing the read lock, with results merged under the write lock?

internal/watchd/daemon.go
```diff
@@ -0,0 +1,215 @@
...
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	filter := make(map[string]bool, len(roots))
+	for _, r := range roots {
+		filter[r] = true
+	}
+
+	var projects []ProjectSnapshot
+	for _, ps := range d.projects {
```
danmux, line 200: d.projects is a map — iteration order is random. In watchAll mode (roots empty) the snapshot returns projects in a different order every tick, so the project rows in the TUI shuffle position between polls. Worth sorting by root path before iterating?

internal/cmd/watch.go
```diff
@@ -29,6 +30,10 @@ func newWatchCmd() *cobra.Command {
 				return fmt.Errorf("watch requires a TTY")
 			}
 
+			if err := watchd.EnsureRunning([]string{"watch", "_daemon"}); err != nil {
+				_, _ = fmt.Fprintf(cmd.ErrOrStderr(), "chunk watch: daemon unavailable, running without background updates: %v\n", err)
```
danmux: Rest of the cmd layer uses iostream.FromCmd(cmd) for output — could this use io.ErrPrintln(...) for consistency and testability?

PR #518, opened 2026-08-19
Select a sidecar snapshot when a repo has no image configured

jesseworld22:
Problem: a repo with no <validation.sidecarImage> got the bare default image - no toolchain, so every command failed on the environment instead of the code.

Fix: pick from the org's own snapshots instead, matching on a snapshot's name and tag.

Ranking, strongest first:
1. Rep name - git remote, then <vcs.repo>, then the directory name
2. Detected stack
3. Tiebreak: the org's own snapshot over an equivalent system one

Matching is token-based, not substring: <go> matches <go-base> but not <mongo-api>.  The wrong prepared environment fails in ways that look like repo's own, so a loose match is worse than no match.

Fallback: no match means the default image and a printed reason. selection never fails a run - an unreachable snapshot API is warned about and treated as no match.

Main Changes: split the ranking out of the prose paragraph so it's scannable, promoted "Problem / Fix / Fallback" to labels, and moved the "why token-based" justification to follow the rule it explains rather than interrupting it.

### DIFF
--- a/acceptance/sidecar_snapshot_select_test.go
+++ b/acceptance/sidecar_snapshot_select_test.go
@@ -1,0 +1,119 @@
+package acceptance
+
+import (
+	"encoding/json"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/testing/binary"
+	testenv "github.com/CircleCI-Public/chunk-cli/internal/testing/env"
+	"github.com/CircleCI-Public/chunk-cli/internal/testing/fakes"
+	"github.com/CircleCI-Public/chunk-cli/internal/testing/gitrepo"
+)
+
+// writeRemoteCommandConfig writes a .chunk/config.json with one remote command
+// and no validation.sidecarImage — the "repo with no sidecar image" case.
+func writeRemoteCommandConfig(t *testing.T, workDir string) {
+	t.Helper()
+	chunkDir := filepath.Join(workDir, ".chunk")
+	assert.NilError(t, os.MkdirAll(chunkDir, 0o755))
+	cfg := map[string]any{
+		"commands": []map[string]any{
+			{"name": "test", "run": "echo test-output", "remote": true},
+		},
+	}
+	data, err := json.Marshal(cfg)
+	assert.NilError(t, err)
+	assert.NilError(t, os.WriteFile(filepath.Join(chunkDir, "config.json"), data, 0o644))
+}
+
+// createdSidecarImage returns the image attribute of the single create-sidecar
+// request recorded by cci.
+func createdSidecarImage(t *testing.T, cci *fakes.FakeCircleCI) any {
+	t.Helper()
+	createReqs := filterByPath(cci.Recorder.AllRequests(), "/api/v3/sidecar/instances")
+	assert.Equal(t, len(createReqs), 1, "expected exactly 1 create-sidecar request")
+
+	var body map[string]any
+	assert.NilError(t, json.Unmarshal(createReqs[0].Body, &body))
+	envelope, ok := body["data"].(map[string]any)
+	assert.Assert(t, ok, "expected data envelope in request body")
+	attrs, ok := envelope["attributes"].(map[string]any)
+	assert.Assert(t, ok, "expected attributes in data envelope")
+	return attrs["image"]
+}
+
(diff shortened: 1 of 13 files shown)


internal/sidecar/snapshotselect.go
```diff
@@ -0,0 +1,188 @@
...
+	"go":         {"golang"},
+	"javascript": {"js", "node", "nodejs"},
+	"typescript": {"ts", "node", "nodejs"},
+	"python":     {"py"},
+	"ruby":       {"rb"},
+	"java":       {"jvm", "openjdk", "jdk"},
+	"scala":      {"jvm", "openjdk", "jdk", "sbt"},
+	"dotnet":     {"net", "csharp", "dot-net"},
+	"cpp":        {"c++", "cplusplus"},
+	"php":        {"php"},
```
michael-webster: This entry is self-referential: the stack name itself is always implicitly matched, so mapping `"php"` to `{"php"}` adds nothing. Can be removed.

```suggestion
```

internal/sidecar/snapshotselect.go
```diff
@@ -0,0 +1,188 @@
...
+		case normalizeName(s.Name) == repo:
+			score += scoreRepoExact
+			reason = fmt.Sprintf("named for repo %s", c.Repo)
+		case containsAll(tokens, strings.Split(repo, "-")):
+			score += scoreRepoToken
+			reason = fmt.Sprintf("mentions repo %s", c.Repo)
+		}
+	}
+
+	if stack := strings.ToLower(c.Stack); stack != "" && stack != "unknown" {
```
michael-webster: Dead guard: `snapshotCriteria` in `snapshotpick.go` already normalizes `"unknown"` to `""` before this code runs, so the `stack != "unknown"` check is never exercised. The doc comment on `SnapshotCriteria.Stack` can also be simplified to just say callers pass `""` when detection failed.

```suggestion
	if stack := strings.ToLower(c.Stack); stack !=  {
```

internal/sidecar/snapshotselect.go
```diff
@@ -0,0 +1,188 @@
...
+	if score > 0 && !s.IsSystem {
+		score += scoreOwned
+	}
+	return score, reason
+}
+
+// containsAll reports whether every part is present in tokens. Empty parts are
+// ignored so a trailing separator in a repo name does not fail the match.
+func containsAll(tokens map[string]bool, parts []string) bool {
+	found := false
```
michael-webster: The `found` bool can be removed in favour of direct early returns, which is clearer:

```suggestion
	for _, p := range parts {
		if p !=  && !tokens[p] {
			return false
		}
	}
	return true
```

internal/sidecar/snapshotselect.go
```diff
@@ -0,0 +1,188 @@
...
+// "Chunk CLI" and "chunk-cli" compare equal.
+func normalizeName(s string) string {
+	var b strings.Builder
+	for _, r := range strings.ToLower(s) {
+		switch {
+		case r >= 'a' && r <= 'z', r >= '0' && r <= '9':
+			b.WriteRune(r)
+		default:
+			b.WriteRune('-')
+		}
```
michael-webster, line 94: `normalizeName` converts `+` to `-` (via the catch-all branch), but `isSeparator` explicitly preserves `+` so that `"c++"` survives as a single token. This means `normalizeName("c++")` → `"c"`, causing spurious exact-match hits and silently breaking the token-path match for C++ repos.

Suggest treating `+` the same way as `isSeparator` does — keep it as a literal character rather than collapsing it to a dash.

acceptance/sidecar_snapshot_select_test.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+	}
+	data, err := json.Marshal(cfg)
+	assert.NilError(t, err)
+	assert.NilError(t, os.WriteFile(filepath.Join(chunkDir, "config.json"), data, 0o644))
+}
+
+// createdSidecarImage returns the image attribute of the single create-sidecar
+// request recorded by cci.
+func createdSidecarImage(t *testing.T, cci *fakes.FakeCircleCI) any {
```
michael-webster: Returning `any` forces callers to write `image == nil || image == ""` to handle the JSON-unmarshaled value. Returning `string` directly would simplify every callsite:

```suggestion
func createdSidecarImage(t *testing.T, cci *fakes.FakeCircleCI) string {
```

...and update the body to assert/convert the type before returning.

michael-webster (commented):
Code review findings from automated analysis. See inline comments for details.

michael-webster (approved):

PR #520, opened 2026-08-19
Stop two claude sessions from sharing the same sidecar

hanabel1:
## Summary
- If you have two Claude sessions open in the same repo, they were both grabbing the same sidecar which is noisy
- Sidecar state files now record which session owns them. A session will only reuse a sidecar that's already its own or one that nobody owns (created from a plain terminals)
- Parallel sessions each get their own sidecar. 
- Chunk watch has groupings by directory not sidecars which is kind of confusing but I'll address that in another PR. Stay tuned lol.

https://github.com/user-attachments/assets/b2f397d8-f108-4558-8b4f-5271041e0ae5

## Implementation 

-  ActiveSidecar now has a session_id field. Everytime it saves it stamps the current session as owner.
-  when a session closes, its sidecar stays reserved until reap cleans it up. 
-  session id comes from CLAUDE_CODE_SESSION ID which is set automatically by claude code or CHUNK_SESSION_ID ( a manual override, wired into the command context on start up. 
- updated the skill too

## Test plan

- [ ] `task test` passes
- [ ] `task lint` passes
- [ ] Open two Claude sessions in the same repo — each should show a different sidecar ID in `chunk sidecar current`
- [ ] Close one session; the remaining session keeps its own sidecar


internal/cmd/validate.go
```diff
@@ -943,12 +943,15 @@ func resolveOrCreateSidecarID(ctx context.Context, client *circleci.Client, side
...
-	// This prevents accumulation of one sidecar per Claude Code session.
-	if existing, err := sidecar.LoadAnyActive(ctx); err == nil && existing != nil {
-		if saveErr := sidecar.SaveActive(ctx, *existing); saveErr != nil {
-			streams.ErrPrintf("warning: could not promote active sidecar: %v\n", saveErr)
-		}
+	// Fall back to a sidecar this project already has before creating a new one,
+	// which is what stops one sidecar accumulating per agent session. Adoption
+	// skips any sidecar another session is still using, so two sessions sharing
+	// this working tree get one sidecar each.
+	existing, adoptErr := sidecar.AdoptIdleActive(ctx)
```
michael-webster: hmmm do we want to try to reuse sidecars like this? they should be cheap to spin up per session no?

hanabel1: yeah you're right. I dropped it. Sessions now just create their own sidecar without trying to reuse idle ones.

danmux (commented):
Three things:

**`docs/CLI.md:225` + `acceptance/parallel_sessions_test.go:112`** (High) - Doc and test contradict each other. `CLI.md` says a session adopts an unowned (no-session) sidecar. `TestSessionCreatesOwnSidecarAlways` asserts the opposite with message `"a session must create its own sidecar, not adopt an unowned one"`. The `sidecar setup` → agent-session handover workflow described in the docs would silently create a second sidecar instead of reusing the prepared one. Which is correct?

**`internal/sidecar/reap_test.go:89`** - Comment references `AdoptIdleActive` which doesn't exist anywhere in the codebase after `LoadAnyActive` was removed. Stale name will send anyone following the code in circles.

**`internal/cmd/sidecar.go:81` + `internal/cmd/validate.go:1007`** - `shortSessionID` helper defined in `sidecar.go` but the same 8-char truncation is inlined again in `sidecarAutoName` in `validate.go`. Both are in `internal/cmd`. Call the helper.

danmux (approved):

PR #525, opened 2026-08-24
Notify users when a newer version of chunk is available

schurchleycci:
## Summary

- Adds a cached background version check (24h TTL, stored in `~/.local/state/chunk/update-check.json`) that runs alongside every command via a goroutine in `PersistentPreRunE`
- After any command completes, if a newer release was found, prints a one-line notice to stderr pointing to `chunk upgrade` or `brew upgrade chunk` depending on install method
- The `chunk watch` TUI also checks on startup and shows the update notice right-aligned in amber in the footer once the goroutine returns
- Skips in CI environments and dev builds; uses GitHub's releases API (same endpoint as `chunk upgrade`)

## Test plan

- [ ] Run any `chunk` command when on the latest version — no notice appears
- [ ] Manually write a stale cache entry with an older `latest_version` to force a re-fetch, confirm notice appears after the command
- [ ] Run `chunk watch`, confirm notice appears in footer bottom-right when update is available
- [ ] Confirm notice says `brew upgrade chunk` when binary is under `/opt/homebrew/bin/`
- [ ] Confirm no notice appears when `CI=true`

update-demo

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -202,7 +202,8 @@ in `config.Resolve` and makes clients testable.
 | `CLAUDE_WORKING_DIR` | validate | Active worktree directory (Stop hook context) |
 | `CHUNK_HOOKS_DISABLED` | validate, hook | Disable Stop-hook validation when set (any non-empty value) |
 | `XDG_CONFIG_HOME` | config | User config directory (default: `~/.config`) |
-| `XDG_DATA_HOME` | sidecar, validate, upgrade | Per-project state directory, including the hook-mode validate result cache and the 24 h update-check cache (default: `~/.local/share`) |
+| `XDG_DATA_HOME` | sidecar, validate | Per-project state directory, including the hook-mode validate result cache (default: `~/.local/share`) |
+| `XDG_STATE_HOME` | upgrade | User state directory for the 24 h update-check cache (default: `~/.local/state`) |
 | `CHUNK_NO_TELEMETRY` | telemetry | Disable anonymous usage telemetry (any non-empty value) |
 | `NO_ANALYTICS` | telemetry | Disable anonymous usage telemetry (any non-empty value) |
 | `DO_NOT_TRACK` | telemetry | Disable anonymous usage telemetry (any non-empty value) |
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -194,7 +194,7 @@ in `config.Resolve` and makes clients testable.
 | `ANTHROPIC_API_KEY` | anthropic, config, validate | Anthropic authentication |
 | `ANTHROPIC_BASE_URL` | anthropic, validate | API endpoint override |
 | `GITHUB_TOKEN` | github | GitHub authentication |
-| `GITHUB_API_URL` | github | GitHub API endpoint override |
+| `GITHUB_API_URL` | github, upgrade | GitHub API endpoint override (also used by the update check) |
 | `CIRCLE_TOKEN` / `CIRCLECI_TOKEN` | circleci | CircleCI authentication |
 | `CIRCLECI_ORG_ID` | sidecar | CircleCI organization ID (overrides `orgID` in `.chunk/config.json`) |
 | `CIRCLECI_BASE_URL` | circleci | CircleCI endpoint override |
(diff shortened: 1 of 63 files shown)


internal/upgrade/check.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if cached, ok := readCache(cacheFile); ok && time.Since(cached.CheckedAt) < checkCacheTTL {
+		return newerTag(cached.LatestVersion, current)
+	}
+
+	// Claim the 24 h window before the request, not after. Short-lived
+	// commands kick this off in a goroutine and exit without waiting for it,
+	// so a cache written only on success would leave every later invocation
+	// refetching and burning through the rate limit.
+	_ = writeCache(cacheFile, updateCache{CheckedAt: time.Now()})
```
danmux: Does the notice ever land for someone who mostly runs short commands? The claim is written here, but the goroutine that got us here is killed at process exit and `printUpdateNotice` only does a non-blocking read. `chunk config show` or `chunk auth status` is gone in ~10 ms, well before a GitHub round trip finishes, so the fetch dies with the claim already on disk: the cache holds `latest_version: ""` for 24 h, and the next day does the same thing.

Would a short claim work better here, e.g. `CheckedAt: time.Now().Add(-checkCacheTTL + retryWindow)` so an aborted fetch is retried in a few minutes rather than tomorrow, with the full 24 h only written once a real answer comes back?

internal/upgrade/check.go
```diff
@@ -0,0 +1,124 @@
...
+
+func writeCache(path string, c updateCache) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		return err
+	}
+	data, err := json.Marshal(c)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(path, data, 0o600)
```
danmux: `os.WriteFile` truncates before it writes, and this is called from a goroutine that can be killed mid-write at process exit, and from two `chunk` processes at once (parallel agent sessions doing this is a live concern, see #520). A reader landing in that window gets an empty or partial file, `readCache` returns `ok == false`, and the next invocation refetches — the same rate-limit hole the claim on line 72 exists to close.

Write to a temp file in the same dir and `os.Rename` over it?

docs/ARCHITECTURE.md
```diff
@@ -194,19 +194,19 @@ in `config.Resolve` and makes clients testable.
...
+| `GITHUB_API_URL` | github, upgrade | GitHub API endpoint override (also used by the update check) |
 | `CIRCLE_TOKEN` / `CIRCLECI_TOKEN` | circleci | CircleCI authentication |
 | `CIRCLECI_ORG_ID` | sidecar | CircleCI organization ID (overrides `orgID` in `.chunk/config.json`) |
 | `CIRCLECI_BASE_URL` | circleci | CircleCI endpoint override |
 | `CLAUDE_PROJECT_DIR` | settings | IDE-provided project directory used by generated `PreToolUse` hooks |
 | `CLAUDE_WORKING_DIR` | validate | Active worktree directory (Stop hook context) |
 | `CHUNK_HOOKS_DISABLED` | validate, hook | Disable Stop-hook validation when set (any non-empty value) |
 | `XDG_CONFIG_HOME` | config | User config directory (default: `~/.config`) |
-| `XDG_DATA_HOME` | sidecar, validate | Per-project state directory, including the hook-mode validate result cache (default: `~/.local/share`) |
+| `XDG_DATA_HOME` | sidecar, validate, upgrade | Per-project state directory, including the hook-mode validate result cache and the 24 h update-check cache (default: `~/.local/share`) |
```
danmux: `Check` resolves the cache dir with `config.AppState()`, which reads `XDG_STATE_HOME` (default `~/.local/state`) — not `XDG_DATA_HOME`. As written this row points anyone isolating chunk state at the wrong variable, and `XDG_STATE_HOME` is absent from the table entirely even though the code reads it. Worth a separate `XDG_STATE_HOME` row for the update-check cache and dropping `upgrade` from this one?

danmux (commented):
## Critical

### `internal/upgrade/check.go` — cache pre-write poisons update checks for 24 hours on fast commands

`CheckForUpdate` writes an empty cache entry (timestamp, `LatestVersion: ""`) *before* the HTTP request, then fires the request in a goroutine. For fast commands — the vast majority of CLI invocations — the process exits before the goroutine's HTTP call completes. The empty pre-write record stays on disk with a fresh 24h timestamp.

On the next invocation: `readCache` finds a fresh entry, evaluates `newerTag("", current)` → returns `""` → no notice shown. Same outcome on any network error: HTTP fails, empty cache stays, 24h silence.

The feature is effectively dead for the first 24 hours after any fast command and on any network hiccup.

Fix: only write the cache entry on a successful HTTP response. The concurrent-invocation guard can use a short TTL (e.g. 5 min) on the pre-write if you need it, or skip it entirely — duplicate concurrent checks are rare and harmless.

## Required

### `internal/upgrade/check.go` — `testing.Testing()` called in production package

`Check()` imports and calls `testing.Testing()` to skip during test runs. This couples production logic to test execution state and adds a test dependency to the production binary. `CheckForUpdate` already accepts injected parameters making it easy to test without calling `Check()` directly. Replace the `testing.Testing()` guard with an env var (or just rely on the existing `current == "dev"` guard in `CheckForUpdate`).

danmux (approved):

schurchleycci:
> ## Critical
> ### `internal/upgrade/check.go` — cache pre-write poisons update checks for 24 hours on fast commands
> `CheckForUpdate` writes an empty cache entry (timestamp, `LatestVersion: ""`) _before_ the HTTP request, then fires the request in a goroutine. For fast commands — the vast majority of CLI invocations — the process exits before the goroutine's HTTP call completes. The empty pre-write record stays on disk with a fresh 24h timestamp.
> 
> On the next invocation: `readCache` finds a fresh entry, evaluates `newerTag("", current)` → returns `""` → no notice shown. Same outcome on any network error: HTTP fails, empty cache stays, 24h silence.
> 
> The feature is effectively dead for the first 24 hours after any fast command and on any network hiccup.
> 
> Fix: only write the cache entry on a successful HTTP response. The concurrent-invocation guard can use a short TTL (e.g. 5 min) on the pre-write if you need it, or skip it entirely — duplicate concurrent checks are rare and harmless.
> 
> ## Required
> ### `internal/upgrade/check.go` — `testing.Testing()` called in production package
> `Check()` imports and calls `testing.Testing()` to skip during test runs. This couples production logic to test execution state and adds a test dependency to the production binary. `CheckForUpdate` already accepts injected parameters making it easy to test without calling `Check()` directly. Replace the `testing.Testing()` guard with an env var (or just rely on the existing `current == "dev"` guard in `CheckForUpdate`).

Both addressed! Good catches.

PR #528, opened 2026-08-25
Fail instead of running locally when no sidecar is found

jesseworld22:
## Summary
- `chunk validate` was silently passing: when sidecar creation failed, it warned once, ran `remote`-marked commands locally instead, and exited `0` — so the Stop hook saw success and said nothing.
- The real fix suggestion (e.g. `chunk org list`) was getting swallowed by `%v` formatting, leaving users with a bare `create sidecar: not authorized` instead.
- `resolveSidecar` and `runSplitCommands` no longer degrade to local runs on failure — they return errors. This covers all three creation failures and the two post-creation cases (unreachable sidecar, no workspace), so `remote` commands are never silently run locally (`grep -c "locally instead"` → `0`).
- New errors are more actionable: `cannotCreateSidecar` names the org and where its ID came from (flag/env/config/auto-pick); `sidecarSyncError` checks for missing SSH keys and surfaces the `ssh-keygen` fix instead of a bare "not found."
- "No active sidecar" is now a proper status line instead of a raw first-line banner in the hook's `Stop hook error:` output.
- Both new errors list the commands that got skipped. `.gitignore` also now excludes `test-reports/`.

## Notes

- **Sharp edge:** if a repo has both remote and local commands and the sidecar is unavailable, nothing runs now — not even the local ones. Intentional, since a half-run reporting a pass is the same false green in a smaller package.
- `chunk init` does not install a `Stop` hook; repos get it from a checked-in `.claude/settings.json` (ours came from #272). Exposure was narrower than "every repo that ran `init`."
- The missing-workspace path is covered by unit tests on the error helper, not an end-to-end repro — that needs a live sidecar with no workspace built.

## Test Plan

- [ ] `task test` passes (1366 tests, 2 env-gated skips)
- [ ] Against a fake API refusing `POST /api/v3/sidecar/instances`, `chunk validate` with a `remote`-marked command fails non-zero and runs nothing locally — verified for all three creation failures
- [ ] Same three cases on released 0.7.156 run locally and exit `0`, confirming the regression
- [ ] Control: `orgID` set correctly and creation refused still fails closed, and the error names `Project config (.chunk/config.json)` as the source
- [ ] Happy path: sidecar is still created and commands still run remotely
- [ ] `chunk validate --identity-file /nonexistent` surfaces `SSH key not found` with the `ssh-keygen` suggestion
- [ ] Full `chunk validate` against a real sidecar: 4/4 passed, all remote

Fixes FACT-426.

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -91,6 +91,18 @@ func resolveOrgID(orgID, workDir string, pickOrg func() (string, error)) (string
 	return pickOrg()
 }
 
+// orgSource names where resolveOrgID would have taken the org ID from, for use
+// in error messages. It mirrors resolveOrgID's precedence and must be kept in
+// step with it. An empty result means the picker supplied the value, which for
+// a single-collaboration account happens without prompting.
+func orgSource(orgID, workDir string) string {
+	if orgID != "" {
+		return "--org-id"
+	}
+	_, source := config.ResolveOrgID(workDir)
+	return source
+}
+
 func orgPicker(ctx context.Context, client *circleci.Client) func() (string, error) {
 	return func() (string, error) {
 		collabs, err := client.ListCollaborations(ctx)
@@ -121,10 +133,13 @@ func orgPicker(ctx context.Context, client *circleci.Client) func() (string, err
 		idx, err := tui.SelectFromList("Select an organization:", labels)
 		if err != nil {
 			if errors.Is(err, tui.ErrNoTTY) {
+				// hideDetail: the wrapped error is "no interactive terminal
+				// available", which the message already says in full.
 				return "", &userError{
 					msg:        "No interactive terminal available to select an organization.",
 					suggestion: "Run 'chunk org list' to find your org ID, then set it with 'chunk config set orgID <id>' or pass --org-id.",
 					err:        err,
+					hideDetail: true,
 				}
 			}
 			return "", &userError{msg: "Could not select an organization.", suggestion: "Pass --org-id instead.", err: err}
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ test-fetch-on-stale-sidecar.log
 # toolchain installers downloaded into the repo root
 *.pkg
 *.tar.gz
+
+# Test and lint reports written by Taskfile (RESULTS_DIR).
+test-reports/
--- a/internal/cmd/usererr.go
+++ b/internal/cmd/usererr.go
@@ -74,6 +74,47 @@ func cannotCreateSidecar(orgID, source string, err error) error {
 		wrap(err)
 }
 
+// unreachableSidecar and missingWorkspace phrase the two ways a sidecar that
+// already exists can fail to run the commands routed to it. Both return an
+// error rather than letting those commands run locally: a command is marked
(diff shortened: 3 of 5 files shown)


michael-webster (commented):
### \`orgSource\` duplicates \`resolveOrgID\` precedence and will silently drift

**\`internal/cmd/sidecar.go:103\`** | **Medium**

\`orgSource\` re-implements \`resolveOrgID\`'s lookup order independently — both check \`if orgID != ""\` first, then both call \`config.ResolveOrgID(workDir)\`. The code comment even flags the hazard: "must be kept in step with it." The concrete failure scenario: if a future change adds a new org ID source between the flag check and \`config.ResolveOrgID\`, \`resolveOrgID\` would use it but \`orgSource\` would not. The \`cannotCreateSidecar\` error would then attribute the rejection to the wrong source, sending the user to fix the wrong config.

Since \`resolveOrgID\` already calls \`config.ResolveOrgID\` internally, the source is available at resolution time — no need to re-derive it. Consider returning source alongside the ID:

\`\`\`go
func resolveOrgID(orgID, workDir string, pickOrg func() (string, error)) (id, source string, err error) {
    if orgID != "" {
        return orgID, "--org-id", nil
    }
    if v, src := config.ResolveOrgID(workDir); v != "" {
        return v, src, nil
    }
    id, err = pickOrg()
    return id, "", err
}
\`\`\`

This lets \`orgSource\` and its parallel flag check be deleted, with \`cannotCreateSidecar\` receiving the source from the single authoritative call site.

internal/cmd/sidecar.go
```diff
@@ -91,6 +91,18 @@ func resolveOrgID(orgID, workDir string, pickOrg func() (string, error)) (string
 	return pickOrg()
 }
 
+// orgSource names where resolveOrgID would have taken the org ID from, for use
+// in error messages. It mirrors resolveOrgID's precedence and must be kept in
+// step with it. An empty result means the picker supplied the value, which for
+// a single-collaboration account happens without prompting.
+func orgSource(orgID, workDir string) string {
```
michael-webster, line 98: non-blocking, but do we already have this defined somewhere?

michael-webster (approved):

danmux (commented):
## Required

### `internal/cmd/usererr.go` — `unreachableSidecar` drops the transient-start hint

The old inline error included `"The sidecar may still be starting. Try again in a moment."` The new function replaces that with a generic suggestion to create a replacement. A user who runs `chunk sidecar create` then immediately runs `chunk validate` will hit this path because the sidecar is still initialising — telling them to provision another sidecar is wrong and burns quota. The specific transient hint should be preserved alongside the generic suggestion.

### `internal/cmd/usererr.go:48` — `cannotCreateSidecar` silently returns `nil` for all non-auth errors

The function only returns a non-nil error for `ErrNotAuthorized`; every other error (network, rate limit, server error) returns `nil`. The name reads as a general sidecar-creation-failure constructor, so the next person wiring up a call site won't realise the nil-for-non-auth contract exists. `notAuthorized` is narrowly named and clearly signals its scope; this should be `notAuthorizedToCreateSidecar` or carry a doc comment making the contract explicit.

danmux (approved):

PR #530, opened 2026-08-25
[FACT-447] Load .env.local for local validate execution

schurchleycci:
## Summary

- `.env.local` was previously only loaded when a sidecar was involved (remote runs); purely local `chunk validate` runs never read it
- `loadSidecarEnvVars` now always calls `resolveEnvVars` and only skips the `syncToSidecar` step when there is no sidecar
- `RunInline`, `RunNamed`, and `RunAll` in the `validate` package each accept an `envVars map[string]string`; when non-empty, the subprocess environment is built from `os.Environ()` with the file's values overlaid

## Test plan

- [ ] Added `TestRunAllEnvVars` verifying that a key set in the `envVars` map is visible to the subprocess (`echo $CHUNK_TEST_VAR` captures the value)
- [ ] Full test suite passes (`task test`) with race detector
- [ ] Linting clean (`task lint`)

Closes FACT-447

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/zalando/go-keyring v0.2.8
 	golang.org/x/crypto v0.54.0
 	golang.org/x/sync v0.22.0
+	golang.org/x/sys v0.47.0
 	golang.org/x/term v0.45.0
 	gotest.tools/v3 v3.5.2
 )
@@ -261,7 +262,6 @@ require (
 	golang.org/x/exp/typeparams v0.0.0-20260209203927-2842357ff358 // indirect
 	golang.org/x/mod v0.37.0 // indirect
 	golang.org/x/net v0.56.0 // indirect
-	golang.org/x/sys v0.47.0 // indirect
 	golang.org/x/text v0.40.0 // indirect
 	golang.org/x/tools v0.47.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
--- a/internal/cmd/env.go
+++ b/internal/cmd/env.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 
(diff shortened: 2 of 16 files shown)


danmux (commented):
## Required

### `internal/cmd/prune.go:83` — `ClearActiveByOrg` error silently discarded

The original code printed `"Warning: could not clear active sidecar state: %v"` on failure. The replacement `if cleared, _ := sidecar.ClearActiveByOrg(resolvedOrgID); cleared > 0` discards the error entirely. If `AppData()` or `ReadDir` fails the user gets no feedback and has no idea the prune did nothing. The error should surface as a warning.

### `internal/sidecar/active.go:~300` — per-entry errors silently swallowed in `ClearActiveByOrg`

Two silent discards inside the loop:
- `entries, _ := loadStateEntries(...)` — a corrupted state directory is skipped with no indication
- `if rerr := os.Remove(e.path); rerr == nil { removed++ }` — removal failures are swallowed; the returned count doesn't distinguish removed from tried-and-failed

Combined with the prune.go discards above, a prune that silently does nothing is invisible to the user.

danmux (approved):

schurchleycci:
> ## Required
> ### `internal/cmd/prune.go:83` — `ClearActiveByOrg` error silently discarded
> The original code printed `"Warning: could not clear active sidecar state: %v"` on failure. The replacement `if cleared, _ := sidecar.ClearActiveByOrg(resolvedOrgID); cleared > 0` discards the error entirely. If `AppData()` or `ReadDir` fails the user gets no feedback and has no idea the prune did nothing. The error should surface as a warning.
> 
> ### `internal/sidecar/active.go:~300` — per-entry errors silently swallowed in `ClearActiveByOrg`
> Two silent discards inside the loop:
> 
> * `entries, _ := loadStateEntries(...)` — a corrupted state directory is skipped with no indication
> * `if rerr := os.Remove(e.path); rerr == nil { removed++ }` — removal failures are swallowed; the returned count doesn't distinguish removed from tried-and-failed
> 
> Combined with the prune.go discards above, a prune that silently does nothing is invisible to the user.

Thank you! These are addressed now

PR #531, opened 2026-08-25
Make chunk watch default to --all

jesseworld22:
## Summary

`chunk watch` watched only the current working directory unless you passed `--all`. Watching everything is the more useful default for a dashboard:

- **Default** - discovers and watches every project `chunk` has seen before (via the `project-root` breadcrumbs each project registers).
- **`--focus`** (new) - narrows back to the current directory only.
- **`--all`** - kept and still parses, but mark deprecated: it prints a one-line pointer to `--focus` and is hidden from `--help`. Existing invocations, aliases, and script's don't break.

Root selection moved out of `RunE` into a `watchRoots(cwd, focus, args)` helper so it can be tested without a TTY.

## Test plan

- [x] `task test` - 1369 pass, 2 skipped (the usual env-gated `CHUNK_ENV_BUILDER_ACCEPTANCE` / `op` CLI ones)
- [x] `task lint` - 0 issues
- [x] New unit tests in `internal/cmd/watch_test_go`:
    - default mode picks up a project registered via its breadcrumb
    - `--focus' yields exactly `[cwd, args...]` and ignores known projects
    - `--all` is still registered and carries a deprecation message
- [x] `chunk watch --help` shows `--focus`, hides `--all`
- [x] `chunk watch --all` prints `Flag --all has been deprecated, watching all known projects is now the default; use --focus...` and still proceeds
- [x]  Manual: run `chunk watch` in a real terminal with sidecars in two projects, confirm both appear; then `chunk watch --focus` and confirm only the current one does

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -143,7 +143,7 @@ chunk
 │           --json                  # Output as JSON
 │
 ├── watch [dir...]                  # Live TUI dashboard for active sidecars and recent activity
-│   --all                           # Watch all known projects, not just the current directory
+│   --focus                         # Watch only the current directory instead of all known projects
 │
 ├── hook                            # Manage chunk hook execution
 │   --project <path>                # Override project directory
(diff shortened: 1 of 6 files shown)


hanabel1:
Looks like this PR is two separate changes? I think you might have accidentally combined your changes for this the sidecar fallback removal ticket ([FACT 426](https://linear.app/circleci/issue/FACT-426/improve-experience-for-no-active-sidecar-found)) into this.
The watch change is only the two watch files and the doc updates.

jesseworld22:
rebased onto main as one commit (two watch files + doc updates)

hanabel1 (approved):

PR #532, opened 2026-08-25
Show per-session sidecars separately in watch TUI

hanabel1:
- Don't collapse two real sidecars on the same branch; sessions that each created their own sidecar now appear as distinct left-pane entries
- Show sidecar name under snapshot name in left pane
- Print branch + session label header in stop-hook runs so concurrent invocations are distinguishable in the feedback block 

before:
Screenshot 2026-08-25 at 11 09 00 AM
after:
Screenshot 2026-08-25 at 1 51 38 PM

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -231,7 +231,7 @@ func initHook(ctx context.Context, hook *hookContext, workDir string, tree gitut
 	if len(sessionLabel) > 8 {
 		sessionLabel = sessionLabel[:8]
 	}
-	if branch, err := gitutil.CurrentBranch(); err == nil && branch != "" {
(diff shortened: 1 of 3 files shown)


danmux (commented):
**`internal/cmd/validate.go:232`** - `gitutil.CurrentBranch()` resolves from the process CWD, not `workDir`. If the shell CWD differs from the repo root when the Stop hook fires (e.g. agent running from a subdirectory), the header shows the wrong branch or falls back to the no-branch format. Both `workDir` and `tree` are in scope here - is there a path-accepting variant of `CurrentBranch` that can take `workDir`?

danmux (approved):

PR #534, opened 2026-08-25
Add opt in desktop notifications for validate results

hanabel1:
## Summary

Adds an opt-in notifications user config key that sends an os desktop notification when chunk validate finishes, reporting pass/fail, check counts, and the elapsed time. Disabled by default - enable it with `chunk config set notifications true`. 

Screenshot 2026-08-25 at 2 56 13 PM

## Implementation 
- Adds Notifications bool to UserConfig/ResolvedConfig and propagates it through config resolution
- New internal/notify/ package wraps the OS notification call
- finishValidate accepts a notifyDesktop bool and fires notify.Send after reporting the outcome
- config show/config set supports the new key; extracts parseBoolValue to deduplicate boolean parsing

## Test Plan
- [ ] chunk config set notifications true → chunk config show reflects it
- [ ] chunk validate (passing) → desktop notification appears with correct counts and elapsed
- [ ] chunk validate (failing) → desktop notification shows failed message
- [ ] chunk config set notifications false → no notification sent
- [ ] chunk config set notifications invalid → user-facing error with accepted values hint

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -170,7 +170,7 @@ chunk
 - `build-prompt` does not write intermediate files by default. Pass `--debug` to write the raw details JSON, analysis markdown, and PR rankings CSV alongside the prompt — useful when diagnosing unexpected prompt output.
 - `task run` defaults to pipeline-as-tool mode; use `--no-pipeline-as-tool`
   to disable.
-- `config set` user keys: `model`, `telemetry`. Project keys (`.chunk/config.json`):
+- `config set` user keys: `model`, `telemetry`, `notifications`. Project keys (`.chunk/config.json`):
   `orgID`, `validation.sidecarImage`. Credentials use `chunk auth set`, not `config set`.
 - `validate --mark-remote` sets `remote: true` on commands in `.chunk/config.json`
   and exits without running anything. With a `[name]` it marks that one command;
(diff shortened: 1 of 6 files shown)


michael-webster (commented):
### Finding 1 — \`Sender\` interface has a single implementation; \`DefaultSender\` global is a race hazard

**\`internal/notify/notify.go:12\`** | **High**

The \`Sender\` interface has exactly one real implementation (\`osSender\`). The review standard is explicit: no interface with only one implementation — single-method dependency injection should use a \`func\` parameter instead. The \`DefaultSender\` package-level global compounds this: it's swapped in tests without synchronization, so any test that calls \`t.Parallel()\` would produce a data race under \`-race\`.

The intent (let tests observe notifications) is fine — the mechanism should just be a \`func\` parameter threaded through \`finishValidate\`:

\`\`\`go
// notify.go — drop the interface and global entirely
func Send(title, body string) { /* direct osSender logic */ }
\`\`\`

\`\`\`go
// validate.go — pass a func for testability
func finishValidate(..., notifyFn func(title, body string)) error {
    if notifyFn != nil {
        notifyFn("chunk validate passed", ...)
    }
}
\`\`\`

Tests can pass a capturing closure without touching global state.

---

### Finding 2 — \`cmd.Run()\` is synchronous; \`finishValidate\` can block for up to 3 seconds

**\`internal/notify/notify.go:56\`** | **High**

The package comment promises "best-effort" fire-and-forget, but \`cmd.Run()\` is called directly (blocking). If \`osascript\`, \`notify-send\`, or \`powershell\` stalls, \`finishValidate\` silently hangs for the full 3-second timeout — directly counter to "must not disrupt the caller." Suggested fix:

\`\`\`go
go func() {
    ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
    defer cancel()
    _ = cmd.Run()
}()
\`\`\`

michael-webster (approved):
approved mod comments.

danmux (commented):
**`internal/notify/notify.go:12`** - `Sender` has only one production implementation (`osSender`). Project standard: no interface with a single impl. Replace with a plain func variable:

```go
var Send = func(title, body string) { osSend(title, body) }
```

Tests swap `notify.Send` directly and restore with `t.Cleanup`. Drops the interface, `DefaultSender`, and the struct wrapper.

danmux (approved):

PR #537, opened 2026-08-26
Skip interactive org picker in non-interactive (CI) sessions

michael-webster:
## Problem

In agent and CI sessions the \`CI\` environment variable is set, but a PTY may still be allocated — so \`term.IsTerminal\` returns true. When \`chunk\` needs to resolve an org ID and neither \`--org-id\` nor a stored \`orgID\` is found, it falls through to the interactive org picker. With \`CI=true\` the picker is rendered by bubbletea and then waits indefinitely for keyboard input that never arrives, **hanging the session**.

## Fix

Added a \`nonInteractive()\` guard in \`orgPicker\` (\`internal/cmd/sidecar.go\`) immediately before the call to \`tui.SelectFromList\`. This mirrors the same pattern already used in \`auth.go\`. When \`CI=true\` the function returns immediately with a clear \`userError\` wrapping \`tui.ErrNoTTY\`, directing the user to run \`chunk org list\` and set their org ID.

Single-org auto-selection is unaffected — the guard only fires when multiple orgs are found and interactive selection would be required.

## Test plan

- [x] \`task build\` passes
- [x] \`task test\` passes — new test \`TestOrgPicker_MultipleOrgs_NonInteractive\` verifies the fix
- [x] Existing \`TestOrgPicker_MultipleOrgs_NoTTY\` and \`TestOrgPicker_SingleOrg\` continue to pass

### DIFF
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -114,6 +114,13 @@ func orgPicker(ctx context.Context, client *circleci.Client) func() (string, err
 		if len(collabs) == 1 {
 			return collabs[0].ID, nil
 		}
+		if nonInteractive() {
+			return "", &userError{
+				msg:        "No interactive terminal available to select an organization.",
+				suggestion: "Run 'chunk org list' to find your org ID, then set it with 'chunk config set orgID <id>' or pass --org-id.",
+				err:        tui.ErrNoTTY,
+			}
+		}
 		labels := make([]string, len(collabs))
 		for i, c := range collabs {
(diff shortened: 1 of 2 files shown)


michael-webster:
Manually tested on a project with no org id configured. Ran a simulation of the CI build to simulate non-interactive shell:
```
CI=true ./dist/chunk sidecar list
✗ Error: No interactive terminal available to select an organization.

no interactive terminal available

Suggestion: Run 'chunk org list' to find your org ID, then set it with 'chunk config set orgID <id>' or pass --org-id.
```

Ran without the CI env var and got the org picker.

hanabel1 (approved):

PR #538, opened 2026-08-26
Use the CircleCI config to detect validate commands in chunkinit

jesseworld22:
## Summary

- `chunk init` guessed validation commands from root filenames, so a repo whose real build system is outranked by a stray manifest got the wrong ones - carta-web (Django + yarn) got `make test`, which isn't a CI gate there, and backend-platform (bazel/java) got pnpm and pytest. Both teams rewrote the generated config by hand. Same defect class as #322.
- `DetectCommands` now reads `.circleci/config.yml` first - it names the checks that actually gate the branch - and falls back to the existing filename switch when there's no config, the config is dynamic, or nothing in it classifies.
- New `internal/ciconfig` reduces a config (possibly thousands of lines) to the `run` steps of jobs gating the default branch. Handles both `run` forms, custom commands and job parameters with substitution, steps passed to orb commands (`go/with-cache`), `pre-steps`/`post-steps`, `when`/`unless`, anchors, branch filters, approval jobs, and legacy v2 configs.
- It reports rather than guesses what it can't resolve: `Dynamic` for `setup: true`, `SkippedOrbs` for orb internals, `Unresolved`, `Truncated`. An empty result means "couldn't tell", not "no gates".
- `internal/validate` classifies steps into install/test/lint/format, dropping work that belongs to CI but not the inner loop (deploys, uploads, image pushes) and anything not runnable on a laptop (multi-line procedures, `$BASH_ENV`/`~/project`, step-level `working_directory`).
- On this repo, init goes from `task test` / `task lint` / `task fmt` to `task mod-download` / `task ci:test` / `task ci:lint` - the formatter drops out because CI has no standalone fmt gate (see note).

Note: **no `format` command is emitted** when CI has no standalone formatter gate, so Go users lose today's `gofmt -w .` autofix - faithful to "CI is ground truth", but formatting is an autofix, not a gate, so shout if you'd rather keep the toolchain default. And `envbuilder`'s install detection is untouched.

## Test plan

- [x] `TestDetectCommandsPrefersCIConfig` - bazel repo with stray `package.json` + `pyproject.toml` returns `bazel test //...`, not `pnpm test`
- [x] `TestDetectCommandsFallsBackToFilenames` / `TestDetectCommandsIgnoreUnusableCIConfig` - no config or a dynamic one leaves existing behavior intact
- [x] `TestExtractIgnoresJobWorkingDirectory` - job-level `working_directory` is the checkout path, not a subdirectory; reading it as one silently discarded every step
- [x] `TestCommandsFromCIKeepsBuildProfileFlags` - `cargo test --release` survives; `./release.sh` doesn't
- [x] `TestExtractWalksStepsPassedToOrbCommands` - steps wrapped in `go/with-cache` are followed
- [x] `FuzzExtract` - 40k execs, no panics; the config comes from whatever repo the user points chunk at
- [x] `task test` (1462), `task acceptance-test` (283), `task lint` clean; `ciconfig` 94.5% covered
- [ ] Verify against carta-web's and backend-platform's real configs - two bugs so far were invisible to synthetic fixtures and obvious against real input


internal/ciconfig/ciconfig.go
```diff
@@ -0,0 +1,562 @@
...
+// a duplicate, or still carries an unsubstituted parameter.
+func (e *extractor) addRun(jc jobCtx, s step, args map[string]string) {
+	if s.Background {
+		return
+	}
+	cmd := strings.TrimSpace(substitute(s.Command, args))
+	if cmd == "" {
+		return
+	}
+	if paramRef.MatchString(cmd) {
```
schurchleycci: `paramRef` only matches `<< parameters.x >>`, so `<< pipeline.parameters.x >>` — along with `<< pipeline.git.branch >>` and `<< matrix.os >>` — is neither substituted nor counted as `Unresolved`.

A config with `- run: pytest -n << pipeline.parameters.workers >>` yields a candidate whose `Command` is that literal string, and `commandsFromCI` writes it to `.chunk/config.json` as `test=pytest -n << pipeline.parameters.workers >>` with empty `Notes`, so nothing warns the user. `<<` is a bash heredoc operator, so `chunk validate` then hits a shell syntax error, or hangs waiting on the heredoc.

Pipeline parameters are common in 2.1 configs, so this is easy to trip. A separate wider pattern for the unresolved check — something like `<<\s*[a-z]+\.` — would catch all these forms while `paramRef` stays narrow for substitution.

schurchleycci (commented):
Read through the CircleCI config reducer and the step classifier. `go build ./...` and the affected test packages pass; the notes below are behavioral issues I verified against this branch with throwaway probe tests.

Six correctness items and one comment nit, inline. The two I'd most want addressed before merge are the unexpanded `<< pipeline.parameters.x >>` passthrough (produces a `.chunk/config.json` command that is a shell syntax error, silently) and `when: on_fail` steps being eligible for a role slot (a log-dumping cleanup step can displace the real test command).

Everything else looked right to me: the `DetectCommands` signature change has its one non-test caller updated, the `hasInstallCommand` gate avoids a duplicate `install` entry, the `yaml.v3` indirect→direct promotion in `go.mod` is correct now that `internal/ciconfig` imports it, and the depth-bounded custom-command expansion, the `maxCandidates` cap, and the document-order workflow walk all behave as documented.

internal/ciconfig/ciconfig.go
```diff
@@ -0,0 +1,562 @@
...
+		if err := n.Content[i+1].Decode(&steps); err != nil {
+			return nil
+		}
+		return steps
+	}
+	return nil
+}
+
+// decodeRun handles both `run: cmd` and the expanded mapping form.
+func (s *step) decodeRun(n *yaml.Node) {
```
schurchleycci, line 583: `decodeRun` reads `name`/`command`/`background`/`working_directory` but not `when`, so a step's `when` is invisible to the rest of the extractor.

A step with `when: on_fail` is diagnostic work that never runs on a green build, yet it is emitted as a gate candidate — the same class of non-gate step that `background: true` is already filtered for. With a job whose first step is `{name: "Show test failures", when: on_fail, command: ./scripts/dump-logs.sh}` followed by `run: go test ./...`, "first match per role wins" hands the `test` slot to the log dumper and the real test command is discarded.

Decoding `when` and skipping `on_fail` alongside the existing `background` check would fix it. `when: always` is a judgment call — it does run on green builds, but in practice it marks cleanup and artifact upload rather than gates, so it may be worth skipping too.

internal/validate/setup.go
```diff
@@ -35,6 +59,87 @@ func DetectCommands(ctx context.Context, claude *anthropic.Client, workDir strin
...
+	return Detection{
+		Commands: []config.Command{{Name: "test", Run: result, Role: config.RoleGate}},
+		Source:   sourceClaude,
+		Notes:    fallbackNotes(ci),
+	}, nil
+}
+
+// withAutofixDefaults adds the autofix commands from defaults that ci does not
+// already provide, leaving every gate ci named untouched.
+func withAutofixDefaults(ci, defaults []config.Command) []config.Command {
```
schurchleycci, line 117: The intent here is that CI never names a *rewriting* formatter, so the toolchain autofix default fills the `format` slot. But the fallback is keyed on whether CI produced a command *named* `format`, and `roleMarkers[roleFormat]` matches check-only formatters too — `prettier`, `cargo fmt`, `ruff format`, `gofmt`, `:fmt`, `" fmt"`.

So `prettier --check .`, `cargo fmt --check`, or `task ci:fmt-check` classifies as `roleFormat`, occupies the slot, and suppresses the real fixer. The user ends up with a `format` command carrying `RoleAutofix` that cannot fix anything, so pre-commit autofix reports a failure instead of reformatting.

The current tests only exercise the two cases where this doesn't bite: `task ci:diff`, which matches no marker at all, and `goimports -w .`, a genuine fixer. Splitting the format role by whether the command rewrites (`-w`, `--write`, `--fix`) or only reports (`--check`, `-l`, `--diff`, `--dry-run`) would keep check-only formatters as gates without letting them claim the autofix slot.

internal/ciconfig/ciconfig.go
```diff
@@ -0,0 +1,562 @@
...
+			seen[key] = true
+			out = append(out, j)
+		}
+	}
+	return out
+}
+
+// runsOnDefaultBranch reports whether a job's filters let it run on main.
+func runsOnDefaultBranch(f filters) bool {
+	if slices.ContainsFunc(f.Branches.Ignore, matchesDefaultBranch) {
```
schurchleycci: `matchesDefaultBranch` is documented as failing open — an uncompilable regex matches, so an unparseable filter keeps the job rather than dropping it — but that reasoning only holds for `only`. In this `Ignore` branch the same `true` means "main is ignored", so the job is dropped instead of kept.

A filter of `ignore: /release-[/` (invalid regex) therefore drops the job, and `Extract` returns zero candidates — a single malformed filter silently kills detection for that job.

The fail-open value needs to be per-direction: `false` for `ignore`, `true` for `only`.

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,275 @@
...
+		notes = append(notes, fmt.Sprintf("%d step(s) past the scan limit were not read", res.Truncated))
+	}
+	return notes
+}
+
+// candidateRun renders a candidate as something a developer can run from the
+// repo root. A step-level working_directory narrows where the command runs, so
+// it is preserved as a `cd` prefix rather than discarded: in a monorepo the
+// subdirectory checks are the real gates.
+func candidateRun(c ciconfig.Candidate) string {
```
schurchleycci, line 210: The `cd <dir> && ` prefix binds more loosely than the operators in the command it wraps.

`working_directory: web` with `command: yarn lint || true` produces `cd web && yarn lint || true`, which parses as `(cd web && yarn lint) || true` — a failed `cd` is swallowed and the step reports a pass from the wrong directory. The `;`-separated case is worse: `cd web && a; b` runs `b` from the repo root regardless of whether the `cd` succeeded.

Grouping the original command preserves the intended precedence — `cd <dir> && ( <cmd> )`, or parenthesizing only when the command contains shell separators.

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,275 @@
...
+// which names the checks that actually gate the default branch. Commands are
+// empty when there is no config, when the config is dynamic, or when nothing
+// in it classifies — in every case the caller falls back to filename
+// detection, and Notes says why so `chunk init` can tell the user.
+//
+// At most one command is kept per role: the first match wins, so a job's
+// primary test step beats a later variant like an acceptance-only run.
+func commandsFromCI(workDir string) Detection {
+	res, err := ciconfig.Extract(workDir)
+	if err != nil || res == nil {
```
schurchleycci: `if err != nil || res == nil` discards `Extract`'s error, including the `parse %s: %w` failure.

A `.circleci/config.yml` that yaml.v3 rejects therefore yields an empty `Source`, so `fallbackNotes` adds nothing and `chunk init` silently falls back to filename guessing with no indication that the config was found and failed to parse — the "surprising rather than explicable" outcome `Detection.Notes` exists to prevent.

Distinguishing "no config at all" (nothing to explain) from "config present but unreadable" (worth a note) would surface it.

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,275 @@
...
+// roleMarkers classify a step by substring, tried in order. Install is first
+// so `yarn install` is not read as a test, and format precedes lint so a
+// formatter is not mistaken for a checker.
+var roleMarkers = []struct {
+	role    string
+	markers []string
+}{
+	{roleInstall, []string{
+		// Covers npm/yarn/pnpm/bun/bundle/pip/poetry install in one marker.
+		// "apt-get install" and friends are removed by skipMarkers first.
```
schurchleycci: Nit: this points at the wrong list. `apt`/`apt-get` are in `skipTools`, which matches the leading word of each command, not `skipMarkers`. The behavior is correct — `apt-get install x` does get filtered — only the comment misdirects.

internal/validate/setup.go
schurchleycci, line 1: Claude findings:

 internal/validate/setup.go:63 — a repo can end up with no test gate at all. Once any CI-derived command classifies, the code returns early and backfills only autofix roles. A Go repo whose config has command: golangci-lint run ./... plus a multi-line command: | test script yields lint + format and no test — where pre-PR detection produced go test ./.... Multi-line test scripts, orb-provided test jobs, and skip-marker matches are all common, and ciNotes emits nothing for classify-time drops, so .chunk/config.json is silently test-less. The "CI is authoritative for the gates it names" comment reads as deliberate, but silently dropping test is a different thing from deferring to CI. Suggest backfilling any unnamed gate role from commandsFromFilenames, or at minimum a note.

- internal/validate/setup.go:78, :103 — the claude == nil and empty-response paths return bare Detection{}, discarding fallbackNotes(ci). That's exactly the dynamic-or-unparseable-config case the Notes plumbing exists for.

internal/ciconfig/ciconfig.go
schurchleycci, line 1: Claude findings:

- internal/validate/ciconfig.go:70 — ciOnlyMarkers/skipTools don't cover the circleci CLI, so gotestsum -- $(go list ./... | circleci tests split --split-by=timings) is emitted verbatim. That split idiom is standard, and locally it's either not installed or fails without CIRCLE_NODE_TOTAL/CIRCLE_NODE_INDEX.
- internal/validate/ciconfig.go:35 — "artifact" as a bare case-folded substring drops real gates: mkdir -p /tmp/artifacts && pytest --junitxml=/tmp/artifacts/results.xml classifies as roleNone. Writing junit/coverage output under an artifacts path is routine. Match store_artifacts-shaped intent rather than the substring.

- internal/ciconfig/ciconfig.go:237 — strings.Trim(pattern, "/") strips all slashes, so only: /^release\// becomes ^release\, fails to compile, and fails open — treating a release-only job as a default-branch gate. Use TrimSuffix(TrimPrefix(...)). Same site: CircleCI anchors filter regexes to the whole branch name; re.MatchString is unanchored.
- internal/ciconfig/ciconfig.go:595 — in decodeRun, one un-decodable field discards the whole step without incrementing Unresolved. background: << parameters.daemon >> makes an npm test step vanish while the notes claim nothing was missed.
- internal/ciconfig/ciconfig.go:61 — defaultBranches hardcodes main/master; a develop-default repo gets zero gate jobs and a misleading "no runnable checks were found" note. The real default branch isn't plumbed in anywhere, so this is a design gap, not a typo.
- internal/ciconfig/ciconfig.go:186 — scheduled workflows are excluded via triggers:, but workflow-level when:/unless: isn't, so jobs from a workflow that's off by default count as branch gates.

schurchleycci:
> Read through the CircleCI config reducer and the step classifier. `go build ./...` and the affected test packages pass; the notes below are behavioral issues I verified against this branch with throwaway probe tests.
> 
> Six correctness items and one comment nit, inline. The two I'd most want addressed before merge are the unexpanded `<< pipeline.parameters.x >>` passthrough (produces a `.chunk/config.json` command that is a shell syntax error, silently) and `when: on_fail` steps being eligible for a role slot (a log-dumping cleanup step can displace the real test command).
> 
> Everything else looked right to me: the `DetectCommands` signature change has its one non-test caller updated, the `hasInstallCommand` gate avoids a duplicate `install` entry, the `yaml.v3` indirect→direct promotion in `go.mod` is correct now that `internal/ciconfig` imports it, and the depth-bounded custom-command expansion, the `maxCandidates` cap, and the document-order workflow walk all behave as documented.

Whoa I didn't realize Claude had posted a comment on my behalf here 😅

internal/ciconfig/ciconfig.go
```diff
@@ -0,0 +1,772 @@
...
+		if !workflowRuns(w, params) {
+			// The workflow is switched off by default, so nothing in it gates a
+			// push either.
+			continue
+		}
+		for _, j := range w.Jobs {
+			if j.Name == "" || j.Type == "approval" {
+				continue
+			}
+			if !runsOnDefaultBranch(j.Filters, branches) {
```
schurchleycci: Should we be filtering in default branch jobs, or filtering them out? Filtering them out might be a safer starting point, but I could be convinced otherwise

schurchleycci (approved):
Looks like it works! Might be worth getting another set of eyes on it before merging though

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,382 @@
...
+	roleLint        = "lint"
+	roleTest        = "test"
+)
+
+// skipMarkers mark a step that runs in CI but has no business in a developer's
+// inner loop: deploys, uploads, and reporting. Matched case-folded as a
+// substring, because the work is named rather than invoked by a fixed tool —
+// `make deploy`, `npm run upload-coverage`.
+var skipMarkers = []string{
+	"deploy", "publish", "notify", "slack", "upload",
```
danmux, line 40: `"notify"`, `"sonar"`, `"slack"` are bare substring matches. `go test ./internal/notify/...` or `pytest tests/slack_client.py` would be silently dropped - direct regression of the stated goal. Should these match only as leading words like `skipTools` does?

internal/ciconfig/ciconfig.go
```diff
@@ -0,0 +1,772 @@
...
+		return
+	}
+	// background is a string because a config may parameterize it. Whether the
+	// step is a long-running process or a gate is then unknowable until the
+	// value resolves, and unknowable is reported rather than assumed: guessing
+	// "not a process" hands the user a daemon to run before every commit, and
+	// guessing "process" drops a real gate while the notes claim nothing was
+	// missed.
+	if bg := strings.TrimSpace(substitute(s.Background, args)); bg != "" {
+		background, known := boolValue(bg)
```
danmux, line 437: `continue` here drops the step with no record - `res.Unresolved` stays zero. A config with legitimate deep command nesting (> 4 levels) silently misses checks. Shouldn't this increment `e.res.Unresolved`?

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	// Fall back to the step's own label: "Run tests" around an opaque script.
+	label := strings.ToLower(c.Step)
+	// The label is what classified the step, so it is also the only place the
+	// check-vs-rewrite distinction can be read from.
+	return refineFormat(matchRole(label), cmd+" "+label)
+}
+
+// formatCheckFlags mark a formatter told to report rather than rewrite:
+// `prettier --check .`, `cargo fmt --check`, `gofmt -l .`, `ruff format --diff`.
```
danmux, line 321: `refineFormat(matchRole(label), cmd+" "+label)` passes the full command text into `formatCheckFlags`. If the command has an unrelated `-l` or `-d` flag, the regex fires and the formatter is misclassified as check-only. Why include `cmd` here rather than just `label`?

internal/validate/ciconfig.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	cmd := strings.ToLower(c.Command)
+	for _, m := range ciOnlyMarkers {
+		if strings.Contains(cmd, m) {
+			return roleNone
+		}
+	}
+	for _, m := range skipMarkers {
+		if strings.Contains(cmd, m) {
```
danmux: `strings.Contains(cmd, m)` for single-word entries like `"publish"` also drops `go test ./cmd/publish/...` and `npm run publish-docs-preview`. Is the intent to match publish as a verb at the start of a command?

danmux (approved):

PR #539, opened 2026-08-26
replace git sync with rsync

hanabel1:
## Summary

- Replaces BundleSync/Sync with rsync
- Removes the `--checkout` flag from `sidecar sync`; incremental transfers are now handled by rsync
- Workspace resolution no longer requires a git remote — falls back to `filepath.Base(cwd)`
- Installs rsync on the sidecar automatically on first sync if absent
- For now the apt-get update + install on first sync is kinda slow. Baking it into the image is the cleaner fix. Follow up pr ?

## How it works
rsync needs TCP SSH but the sidecar exposes SSH over WebSocket.
The `.git` directory is included so the remote has a functional git repo. Gitignored files are excluded.

## Test plan
- [x] `task build` passes
- [x] `task test` passes
- [x] `task lint` passes
- [x] Manual: sync transfers working tree to sidecar, incremental on re-run
- [x] Manual: `chunk validate` syncs and runs end-to-end
- [x] Manual: gitignored files absent from remote after sync

Screenshot 2026-08-26 at 3 18 58 PM

Screenshot 2026-08-26 at 3 19 50 PM


internal/sidecar/rsync.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+		if err := persistWorkspace(ctx, repoPath); err != nil {
+			status(iostream.LevelWarn, fmt.Sprintf("Could not save workspace: %v", err))
+		}
+	}
+
+	status(iostream.LevelInfo, fmt.Sprintf("Syncing workspace %s...", repoPath))
+
+	if check, err := ExecOverSSH(ctx, sess, "which rsync", nil, nil); err != nil {
+		return fmt.Errorf("rsync: check remote rsync: %w", err)
```
michael-webster, line 69: do we need to make sure our base images all contain rsync?

michael-webster:
## Testing notes + a suggestion on the SSH proxy

I ran through 6 scenarios manually (initial sync, incremental, delete propagation, `.git` included, gitignore exclusion, gitignore rule changes mid-session) — the core design is solid and everything works correctly once two bugs in the SSH command construction are resolved. Full report: https://claude.ai/code/artifact/e25183f3-694d-49ee-b469-541d4b7622f3

### Bugs found

**1. `IdentitiesOnly` missing — breaks anyone with an active ssh-agent**

The `-e` SSH command has no `IdentitiesOnly=yes`. SSH offers every agent key before the explicit `-i` path; the sidecar disconnects with `too many authentication failures` before `chunk_ai` is tried. Fix:

```go
sshArgs := []string{"ssh", "-p", port,
    "-o", "StrictHostKeyChecking=no",
    "-o", "UserKnownHostsFile=/dev/null",
    "-o", "IdentitiesOnly=yes",   // prevent agent key flood
    "-q",
}
if sess.IdentityFile != "" {
    sshArgs = append(sshArgs, "-i", sess.IdentityFile)
} else if sess.UseAgent && sess.AuthSock != "" {
    sshArgs = append(sshArgs, "-o", "IdentitiesOnly=no")
}
```

**2. `ShellEscape` applied to the identity file path**

`ShellEscape` wraps the path in single quotes (`'/home/.ssh/chunk_ai'`). Rsync tokenizes `-e` by whitespace and calls `execve` directly — no shell to interpret them. SSH sees the quote characters as part of the filename and falls through to agent keys (compounding bug 1). Fix: use `sess.IdentityFile` directly.

---

### Suggested alternative: skip the local ssh subprocess entirely

Both bugs exist because the proxy manufactures an `ssh` command string and relies on the local ssh binary for auth. But we already have `dialSSH`, which authenticates correctly via Go's `x/crypto/ssh` over the WebSocket — no ssh-agent involvement, no command string escaping.

The same proxy pattern works without spawning `ssh` at all:

```go
// Dial SSH once via WebSocket — auth handled by Go, not local ssh binary
sshConn, err := dialSSH(ctx, sess)
if err != nil {
    return fmt.Errorf("rsync: dial ssh: %w", err)
}
defer sshConn.Close()

// Open a session and start rsync --server on the remote
sshSess, err := sshConn.NewSession()
if err != nil {
    return fmt.Errorf("rsync: new ssh session: %w", err)
}

// Bridge a local TCP listener to the SSH session's stdin/stdout
ln, err := net.Listen("tcp", "127.0.0.1:0")
if err != nil {
    return fmt.Errorf("rsync: listen: %w", err)
}
_, port, _ := net.SplitHostPort(ln.Addr().String())

remoteCmd := "rsync --server --delete -logDtprze.iLsfxCIvu . " + ShellEscape(repoPath)

go func() {
    conn, err := ln.Accept()
    ln.Close()
    if err != nil {
        return
    }
    sshSess.Stdin = conn
    sshSess.Stdout = conn
    sshSess.Stderr = io.Discard
    if err := sshSess.Start(remoteCmd); err != nil {
        conn.Close()
        return
    }
    sshSess.Wait()
    conn.Close()
}()

// Local rsync connects to the listener; -e just needs to reach localhost TCP
cmd := exec.CommandContext(ctx, "rsync",
    "--archive",
    "--delete",
    "--filter=:- .gitignore",
    "-e", "ssh -p "+port+" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -q",
    src,
    fmt.Sprintf("%s@127.0.0.1:%s", defaultSSHUser, repoPath),
)
```

The local `ssh` call here only needs to reach `127.0.0.1` — there's no key auth involved because the session is already open on the other side. `IdentitiesOnly=yes` with no `-i` flag means ssh won't try any keys at all, it just establishes the connection and the session is already waiting.

This collapses the two-WebSocket-connection approach to one, removes the key-escaping concern, and means the local `ssh` binary only needs to handle a localhost TCP handshake — no auth, no agent.

Note: the `remoteCmd` rsync flags (`-logDtprze.iLsfxCIvu`) need to match what your local rsync version negotiates; you may want to let rsync negotiate them itself rather than hardcoding. Easiest way: keep the `-e` approach but point it at a tiny wrapper that just `exec`s `ssh` to the already-open session's port with no auth.

michael-webster (approved):

PR #542, opened 2026-08-27
Update TUI to account for session aware sidecars

jesseworld22:
## Summary

Sidecars are isolated per session, so several sessions can hold one for the same worktree - but the dashboard rendered a row per sidecar, all labeled with the same branch name. The session ID never reached the TUI: `ActiveSidecar` persists it, `watchd.SidecarState` didn't carry it, so `loadSidecars` dropped it.

## Before / After

Two sessions in one worktree. Left pane, fixture state.

**Before** - two rows, same label, nothing saying who owns what:

── chunk-cli
▶ jesse/update-tui-session…
   22222222-aaaa-4bbb-8c…
  ⠋ validate...
────────────────────────────
  jesse/update-tui-session…
   11111111-aaaa-4bbb-8c…
  ✓ in sync

**After** - branch once for the group, rows named by session, yours pinned first:

── chunk-cli
  jesse/update-tui-session-…
  2 sessions
▶ ● this session
   11111111-aaaa-4bbb-8c…
  ✓ in sync
────────────────────────────
  ○ bbbbbbbb
   22222222-aaaa-4bbb-8c…
  ⠋ validate...

The same problem arrives a second way: one branch checked out in two directories collapsed under one repo header, both rows reading `main`. They now read `work/chunk-cli` and `tmp/chunk-cli` - the shortest path suffix that tells them apart, since two clones share a basename.

A worktree with **one** session is untouched: verified byte-identical against a binary built from `main`.

## Fixed

Found while testing the above. The daemon ones are here because they blocked verifying it at all.

- Rows started without room to finish were clipped mid-render, losing the sync state and age they exist to report - now costed, dropped whole, and followed by a `↓ N more` hint
- A local `chunk validate` run was credited to whichever sidecar `mergeBranches` inserted last - shared worktrees keep the local row separate, and the header counts sessions not rows
- The daemon serves snapshots shaped by the build that started it with no mismatch detection, so `chunk watch` showed pre-session data after an upgrade until the daemon was killed by hand - `/ping` now reports a build identity
- A dashboard whose daemon died froze on stale data forever, and the "daemon unavailable" line was appended past the last row so it was clipped at every height - polls now relaunch, and the message has room
- `renderActivityPane` bounded the selection above but not below, so a negative index could have panicked on `m.sidecars[-1]`

## Test Plan

- [x] `task test` (1490 tests, `-race`), `task acceptance-test` (289), `task lint`
- [x] Single session frame-diffed against `main` - byte-identical
- [x] Two sessions, two checkouts, and a local run all render as described above
- [x] Overflow at 130x24 drops a row whole with `↓ 1 more`; every row complete at 130x26
- [x] Stale daemon replaced at startup, matching build reused; a daemon killed mid-session is relaunched on the next poll

### DIFF
--- a/internal/tui/watch/model.go
+++ b/internal/tui/watch/model.go
@@ -11,6 +11,7 @@ import (
 	"charm.land/lipgloss/v2"
 
 	"github.com/CircleCI-Public/chunk-cli/internal/eventlog"
+	"github.com/CircleCI-Public/chunk-cli/internal/session"
 	"github.com/CircleCI-Public/chunk-cli/internal/upgrade"
 )
 
@@ -24,6 +25,10 @@ const (
 
 	levelDone  = "done"
 	levelError = "error"
+
+	// localRunnerName labels the synthesised row for validate runs that happened
+	// locally rather than on a sidecar.
+	localRunnerName = "local"
 )
 
 var spinFrames = []string{"⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"}
@@ -86,9 +91,14 @@ type ProjectEntry struct {
 // sidecarIDs lists every sidecar UUID (and "" for local runs) whose events
 // should appear in this entry's activity pane; populated by mergeBranches.
 type sidecarInfo struct {
-	id            string
-	sidecarIDs    []string // all IDs to match when filtering events
-	name          string
+	id         string
+	sidecarIDs []string // all IDs to match when filtering events
+	name       string
+	// sessionID is the agent session owning this sidecar, empty for the local
+	// runner and for state written before sidecars were session-scoped. Several
+	// sessions can hold sidecars for one worktree, so this is what distinguishes
+	// two otherwise identical rows.
+	sessionID     string
 	projectName   string
 	repoName      string    // basename of the main worktree (groups linked worktrees together)
 	branch        string    // current git branch for this worktree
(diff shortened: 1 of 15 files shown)


internal/tui/watch/model.go
```diff
@@ -909,38 +955,110 @@ func (m Model) loadData() tea.Msg {
...
-		repoName   string
-		branch     string
-		projectIdx int
-	}
-	seen := map[key]int{}
+	seen := map[groupKey]int{}
 	result := make([]sidecarInfo, 0, len(sidecars))
 	for _, sc := range sidecars {
-		k := key{sc.repoName, sc.branch, sc.projectIdx}
+		k := groupOf(sc)
```
michael-webster, line 1279: `mergeBranches` uses `groupOf` as its deduplication key — the same `{repoName, branch, projectIdx}` key that `sharedWorktrees` uses, with no `sessionID`. When two sessions hold sidecars for the same worktree, `convertSnapshot` produces two `sidecarInfo` entries with distinct `sessionID` values, but `mergeBranches` sees the same `groupKey` for both and folds the second into the first. After the merge `m.sidecars` has one row per worktree, `sharedWorktrees` returns 1 everywhere, `multi` is always `false`, and the group-header / session-labelling paths never fire.

The tests that exercise multi-session rendering (`TestRenderSidecarPane_twoSessionsAreNamed`, etc.) all set `m.sidecars` directly, bypassing `mergeBranches`, so they do not catch this.

The fix is to include `sessionID` in `mergeBranches`'s deduplication key — `groupOf`/`sharedWorktrees` should stay as-is since they intentionally group by worktree identity. An integration-level test that pipes two session snapshots through `convertSnapshot → sortByActivity → mergeBranches` and asserts two rows survive would prevent a regression.

jesseworld22: turns out the guard at `model.go:1080` (from 2eec790) already keeps two real sidecars separate, so they do survive the merge. 

ran two session snapshots through the full path and got 2 rows with `multi` true.

the promotion block was dropping `sessionID.` if a local run sorted ahead of the sidecars, the row lots its ` ● this session` marker. fixed this + added your integration test in both orderings.

kept `sessionID` out of the key since the local runner has no session and would stop merging into its sidecar row - added a comment to make this clear.

hanabel1 (approved):

PR #543, opened 2026-08-28
Replace hand-rolled ANSI with lipgloss, collapse internal/tui into internal/ui

schurchleycci:
## Summary

- Replaces all hand-rolled ANSI escape code helpers in `internal/ui` with [lipgloss v2](https://charm.land/lipgloss/v2)
- Collapses `internal/tui` into `internal/ui` — the dual-package split was awkward and unnecessary; bubbletea components now live alongside the CLI formatting helpers
- Exports top-level style presets (`SuccessStyle`, `WarningStyle`, etc.) with godoc comments, using semantic names instead of color names
- Adds `IconOK`/`IconWarn`/`IconFail` constants to `internal/ui` so status glyphs are defined once and shared between CLI output and the TUI
- Fixes dim/faint text readability: replaces `Faint(true)` (SGR 2, unreliable) with explicit xterm-256 grays, which remain legible regardless of how the terminal renders SGR 2
- **Semantic states use named ANSI colors (1–6)** so they respect the user's terminal theme (Solarized, Dracula, etc.); only dim/muted grays use fixed xterm-256 values where named bright-black varies too much across themes
- Drops decorative color from op tags (`validate`, `setup`, `hook`) and the update notice — the label text already differentiates them; `sync` and `exec` keep colors since those map to semantic states

Screenshot 2026-08-31 at 1 37 33 PM

Screenshot 2026-08-31 at 1 35 23 PM

## Test plan

- [ ] `task test` passes
- [ ] `task lint` passes
- [ ] `chunk watch` — verify color hierarchy looks correct in your terminal (sidecars, activity pane, step results, log lines)
- [ ] `chunk auth status` — verify bold/dim/success output in a dark terminal
- [ ] Test with a configured terminal theme (e.g. Solarized) — success/warning/error should use theme colors


michael-webster (approved):
LGTM

PR #544, opened 2026-08-28
Allow active sidecar state to represent groups

z00b:
- store active sidecars as a slice with legacy state compatibility
- preserve pooled active state when removing or switching sidecars
- update sidecar, validate, watch, sync, and reap callsites to use the new shape


internal/sidecar/active.go
schurchleycci, line 1: Claude findings:

-  Line 27 [correctness]    `chunk sidecar current --json` now emits `sidecar_ids` (array) instead of `sidecar_id` (string) with no backward-compatible shim in the marshaling direction
-  Line 199 [correctness]    `pruneRekeyedState` checks only the primary (first) sidecar ID via `e.active.ID()`, while `PruneID` already iterates all IDs via `e.ids()`
- Line 345 [simplification] Manual filter-append loop in `RemoveActiveSidecar` duplicates `slices.DeleteFunc`, which is already imported

internal/sidecar/reap_test.go
schurchleycci, line 1: Claude finding:

- Line 194 [correctness]    `BundleSync` replaces the entire `SidecarIDs` slice with a single-element slice when the active sidecar changes, silently dropping secondary members

schurchleycci:
LGTM!

schurchleycci (approved):

PR #545, opened 2026-08-28
Record the sidecar command ID in the chunk event log

schurchleycci:
## Summary

- Adds `CommandID string` to `eventlog.Event` so structured log consumers (e.g. `chunk watch`) can read it without parsing message text
- Adds `Log.AppendCommandID` helper that writes a structured command-ID event directly to the event log after every remote exec
- Changes the `execFn` signature throughout to return `commandID string` as a fifth value, threading it from `client.Exec` → `newExecFn` → `RunRemote`/`RunRemoteInline`
- On failure, emits `command_id: <id>` via `statusFn` so the ID appears in the terminal scrollback and is recorded in `events.jsonl`

Closes FACT-451.

## Why this matters

The sandbox-provisioner already supports replaying a command's full output via `GET /api/v3/sidecar/commands/{id}/output`, but there was no way to learn the `{id}` after the fact — it was only held in memory for the duration of the stream. Now it's persisted in `~/.local/share/chunk/<hash>/events.jsonl` for every remote exec run.

## Test plan

- All 1448 existing tests pass
- New `TestAppendCommandID` in `internal/eventlog/eventlog_test.go` verifies the structured event is written with the correct fields

### DIFF
--- a/internal/circleci/client.go
+++ b/internal/circleci/client.go
@@ -320,7 +320,7 @@ func (c *Client) streamCommandOutput(
 		// reporting it as a stalled stream would send someone hunting a network
 		// fault that does not exist.
 		if err == nil && outcome.frames > 0 && outcome.known == 0 {
-			return nil, ErrOutputFormatUnsupported
+			return result, ErrOutputFormatUnsupported
 		}
 
 		if outcome.cursor != "" {
@@ -332,14 +332,14 @@ func (c *Client) streamCommandOutput(
 			return result, nil
 		case outcome.failed != nil:
 			if !outcome.failed.Retryable {
-				return nil, fmt.Errorf("remote command: %s", outcome.failed.Message)
+				return result, fmt.Errorf("remote command: %s", outcome.failed.Message)
 			}
 		case err != nil:
 			if ctx.Err() != nil {
-				return nil, ctx.Err()
+				return result, ctx.Err()
 			}
 			if !isRetryable(err) {
-				return nil, mapErr("stream command output", err)
+				return result, mapErr("stream command output", err)
 			}
 		}
 
@@ -353,13 +353,13 @@ func (c *Client) streamCommandOutput(
 		} else if err != nil {
 			stalls++
 			if stalls > maxStreamStalls {
-				return nil, fmt.Errorf(
+				return result, fmt.Errorf(
 					"stream command output: gave up after %d attempts that returned nothing", stalls-1)
 			}
 		}
 
 		if attempts >= maxStreamAttempts {
-			return nil, fmt.Errorf(
+			return result, fmt.Errorf(
 				"stream command output: gave up after %d reconnects without the command finishing", attempts)
 		}
 
@@ -371,7 +371,7 @@ func (c *Client) streamCommandOutput(
 		delay := min(time.Duration(stalls)*streamRetryBase, 10*streamRetryBase)
 		select {
 		case <-ctx.Done():
-			return nil, ctx.Err()
+			return result, ctx.Err()
 		case <-time.After(delay):
 		}
 	}
--- a/internal/cmd/sidecar.go
+++ b/internal/cmd/sidecar.go
@@ -570,7 +570,7 @@ func newSidecarSyncCmd() *cobra.Command {
 				if as, _ := sidecar.LoadActive(cmd.Context()); as != nil && as.SidecarID == sidecarID {
 					scName = as.Name
 				}
-				syncFn = eventlog.WrapFromDir(dataDir, syncFn, eventlog.OpSync, sidecarID, scName, sidecar.CurrentBranch(cwd))
+				syncFn = eventlog.WrapFromDir(dataDir, syncFn, eventlog.OpSync, sidecarID, scName, sidecar.CurrentBranch(cwd), nil)
 			}
 			err = sidecar.RsyncSync(cmd.Context(), client, sidecarID, identityFile, authSock, workdir, cwd, syncFn)
 			if err != nil {
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -629,7 +629,7 @@ func runValidate(ctx context.Context, client *circleci.Client, rc config.Resolve
 			streams.ErrPrintf("%s\n", ui.Success(fmt.Sprintf("Saved %s to .chunk/config.json", cmdName)))
 		}
 		if sidecarID != "" && allRemote {
-			execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, streams)
+			execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, workDir, envVars, rc, streams)
 			if err != nil {
 				return validate.Result{}, err
 			}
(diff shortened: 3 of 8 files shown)


jesseworld22:
Reviewed against the team's `.chunk/context/review-prompt.md`. Four things, two of which I think need a fix before merge. All line numbers are on `fact-451-record-sidecar-command-id-in-event-log` @ `3182237`.

## Required

### 1. `internal/cmd/validate.go:886` — the workspace probe's command ID gets stamped onto the run-wide summary event

```go
if wsErr := validate.WorkspaceExists(ctx, execFn, dest); wsErr != nil {
    return validate.Result{}, missingWorkspace(sidecarID, dest, commandNames(remoteCfg.Commands), wsErr)
}
```

`validate.WorkspaceExists` runs `test -d '<dest>'` through the same `execFn`, so `newExecFn:843` publishes the *probe's* command ID into the carrier. When the probe fails we return `missingWorkspace(...)` immediately — no per-command terminal event ever fires to claim that ID. Control reaches `finishValidate`, which emits `statusFn(iostream.LevelError, "0/N passed …")`, and `Wrap` takes the pending ID there (`internal/eventlog/eventlog.go:172-174`), writing the probe's ID onto the run summary. Replaying `GET /api/v3/sidecar/commands/{id}/output` for that event then returns the probe's empty output, presented as the failed run's output.

This is the case the `CommandIDs` doc comment says can't happen ("never a later run-wide summary that belongs to no single command"), and the `newExecFn` comment at 840-842 ("the workspace probe's, say — is replaced by the next exec's") only holds when the probe succeeds.

Draining right after the probe closes it:

```go
// The probe is not a command; its ID must not be left for a later terminal
// event to claim.
wsErr := validate.WorkspaceExists(ctx, execFn, dest)
cmdIDs.Take()
if wsErr != nil {
    return validate.Result{}, missingWorkspace(sidecarID, dest, commandNames(remoteCfg.Commands), wsErr)
}
```

Or hand the probe an `execFn` that doesn't publish at all — `WorkspaceExists` already discards the ID with `_`.

### 2. `internal/cmd/validate.go:835-838` — the ID is discarded precisely when the exec fails mid-stream

```go
if err != nil {
    // The ID of the exec that failed is not reported, so clear the carrier
    // rather than let the previous command's ID be recorded against this one.
    cmdIDs.Set("")
    return "", "", 0, "", err
}
```

The comment says the failed exec's ID "is not reported", but `client.Exec` has it — it submits the command, reads `envelope.Data.ID`, and passes it to `streamCommandOutput(ctx, envelope.Data.ID, …)` (`internal/circleci/client.go:248, 271`). Every failure return in that loop is `return nil, err` (retries exhausted, non-retryable `remote command:` error, stall), so the client drops the ID and `cmdIDs.Set("")` makes the loss permanent.

Concretely: a 10-minute remote `go test` streams for 8 minutes, the connection dies unrecoverably, scrollback is truncated, and the one thing that would recover the full output is the ID we just deleted. Nothing else records it either — `RunRemote`'s `err != nil` branch (`internal/validate/validate.go:183-187`) never emits `command_id:`, and `RunRemoteInline`'s (216-218) emits no status at all.

Could `streamCommandOutput` return its partial result alongside the error, or `Exec` return a non-nil `*ExecResponse` carrying just `CommandID` on failure, so this branch records the ID instead of clearing it? It's the branch where the ID is most valuable. This also makes the PR body's "persisted … for every remote exec run" inaccurate as written.

## Suggestions

### 3. `internal/validate/validate.go:198-200` (and `229-231`) — `command_id:` lands in the event log a second time as message text

```go
if commandID != "" {
    status(iostream.LevelInfo, "command_id: "+commandID)
}
```

`status` is the event-log-wrapped `statusFn`, so a failing remote command writes two `events.jsonl` rows carrying the ID: an `info` row with `msg: "command_id: cmd-abc123"`, then the `error` row with the structured field. That reintroduces the message-text parsing this PR set out to remove, and it's asymmetric — a passing command records the ID structurally only, a failing one twice, plus a bare `command_id: …` row in `chunk watch`. `TestRemoteFailureRecordsCommandIDOnTerminalEvent` guards duplicate *structured* fields but not this.

Writing the scrollback line straight to `streams.Err` keeps terminal visibility without the extra event:

```go
if commandID != "" {
    _, _ = fmt.Fprintf(streams.Err, "  command_id: %s\n", commandID)
}
```

### 4. The PR description documents a mechanism that isn't in the diff

The body says it adds a `Log.AppendCommandID` helper "that writes a structured command-ID event directly to the event log after every remote exec", plus a `TestAppendCommandID`. Neither exists. What's here is `eventlog.CommandIDs` (`internal/eventlog/eventlog.go:117-153`), a shared mutable carrier that `Wrap` drains on the *next* `LevelDone`/`LevelError` call — a different mechanism with different failure modes, and the reason attribution can go wrong at all (findings 1 and 2). Someone debugging a mis-stamped event later will go looking for a direct append helper and won't know correctness hinges on which status call fires next. Worth updating the summary and test-plan bullet.

jesseworld22:
claude findings:

1. **`internal/cmd/validate.go:886`** — `WorkspaceExists` runs `test -d` through the same `execFn`, so the probe's ID lands in the carrier. If the probe fails we return early, no command event claims it, and `finishValidate`'s `"0/N passed"` error event takes it instead — replaying that ID gives the probe's empty output. A `cmdIDs.Take()` right after the probe would close it.

2. **`internal/cmd/validate.go:837`** — `cmdIDs.Set("")` drops the ID exactly when it's most useful: stream dies 8 minutes into a 10-minute test, scrollback truncated, no way to replay. `client.Exec` does have it (`envelope.Data.ID`, `internal/circleci/client.go:248`) — could it return the ID alongside the error rather than `nil, err`?

3. **`internal/validate/validate.go:198-200`** (same at `229-231`) — `status(LevelInfo, "command_id: "+id)` goes through the wrapped `statusFn`, so a failure writes the ID to `events.jsonl` twice: once as message text, once structurally. That's the parsing this PR set out to remove. `fmt.Fprintf(streams.Err, ...)` keeps the scrollback line without the extra event.

4. **Description** — mentions `Log.AppendCommandID` and `TestAppendCommandID`; neither is in the diff. Worth describing the `CommandIDs` carrier instead, since that's where 1 and 2 come from.

jesseworld22 (commented):
claude findings:

builds clean, 1451 tests pass with -race, lint clean on the touched packages. Everything below is behavioral.

1. internal/cmd/validate.go:886 — WorkspaceExists runs test -d through the same execFn, so the probe's ID sits in the carrier. If the probe fails we return early, no command event claims it, and finishValidate's "0/N passed" error event takes it instead — replaying that ID returns the probe's empty output. Add cmdIDs.Take() right after the probe.

2. internal/circleci/client.go — streamCommandOutput:272 builds result := &ExecResponse{CommandID: commandID}, then every failure return throws it away: 288, 300, 304, 307, 321, 327, 339 are all return nil, err. That's what forces Set("") at validate.go:837, and it's the case where replay matters most (stream dies mid-run, scrollback truncated). Return result, err on those seven; Exec:246 correctly stays nil. Then newExecFn becomes Set("") when result == nil, else Set(result.CommandID).

3. internal/validate/validate.go:198-200 (same at 229-231) — status(LevelInfo, "command_id: "+id) goes through the wrapped statusFn, so a failure logs the ID twice: once as message text, once structurally. That's the parsing this PR removes. fmt.Fprintf(streams.Err, ...) keeps the scrollback line without the extra event.

4. No test for either failure mode — probe fails, and Exec errors. Both cheap with fakes.FakeCircleCI; the first settles whether https://github.com/CircleCI-Public/chunk-cli/pull/1 is real.

questions:

5. validate.go:811 — DetectOrgAndRepo(cwd) → DetectOrgAndRepo(localWorkDir). No-op normally, but under --project-dir it changes repo, and so dest. Intentional? Unmentioned and untested.

6. Nothing reads Event.CommandID — no refs in internal/cmd/watch.go, internal/tui/watch/, internal/watchd/. Follow-up tracked?

7. Description describes a Log.AppendCommandID helper and TestAppendCommandID; neither exists. The actual mechanism is the CommandIDs carrier drained by Wrap on the next terminal event.

schurchleycci:
Going to close this and start with a fresh branch

PR #547, opened 2026-08-31
delegate install to claude plugin instead of embedding files

hanabel1:
## Summary

- `chunk skill install` now calls `claude plugin install circleci --yes --scope <scope>` instead of embedding and copying SKILL.md files
- Aligns with the skills marketplace model — skills live in [CircleCI-Public/skills](https://github.com/CircleCI-Public/skills) and are distributed via the `circleci` Claude Code plugin
- Removes embedded file dependency so the CLI no longer needs to be updated when skill content changes

## Changes

- **`internal/skills/install.go`** — rewritten to invoke the `claude` CLI; parses output to distinguish fresh install vs. already-installed (→ `Updated` vs `Installed`)
- **`skills/embed.go`** — gutted; no more `//go:embed` directives
- **`internal/cmd/skills.go`** — `resolveScope` simplified (no longer needs `HOME` since no files are written); output messages reference the plugin name
- **`internal/cmd/init.go`** — `installSkillsStep` drops the `workDir` arg; calls `skills.Install(scope)` directly
- **`internal/testing/env/env.go`** — added `PATH` override support via `Extra["PATH"]` for acceptance tests
- **`internal/skills/skills_test.go`** — rewritten using fake binary pattern
- **`acceptance/skills_test.go`** — rewritten using fake binary pattern with invocation logging
- **`internal/cmd/init_test.go`** — replaced file-creation tests with fake-binary tests that verify `claude plugin install circleci` is called

## Events

- `chunk skill install` → calls `claude plugin install circleci --yes --scope project` (default) or `--scope user`
- `chunk skill install` with no `claude` CLI on PATH → prints "skipped (claude CLI not found)", exits 0
- `chunk init` → calls `chunk skill install` at project scope (same delegation)
- `chunk skill list` → calls `claude plugin list --json` to show install state

## Not collected

- Skill file contents are no longer embedded in the binary
- `chunk init` no longer writes `.claude/skills/` files

## Implementation

Uses the fake binary pattern (shell script in temp dir, PATH injection) consistent with this repo's "fakes over mocks" guideline. For "not found" tests, PATH is fully isolated (`t.TempDir()` only) to prevent the real `claude` on the developer's machine from interfering.

## Test plan

- [x] `task test` — 1473 tests, 1 skipped (E2E sidecar test behind flag), 0 failures
- [x] Install success / already-installed / skipped-when-absent / errors
- [x] User scope vs project scope
- [x] `chunk init` calls plugin install
- [x] `skill list` shows current/missing states

### DIFF
--- a/internal/skills/install.go
+++ b/internal/skills/install.go
@@ -1,112 +1,26 @@
 package skills
 
 import (
+	"encoding/json"
 	"fmt"
-	"os"
-	"path/filepath"
-
-	"github.com/CircleCI-Public/chunk-cli/skills"
+	"os/exec"
+	"strings"
 )
 
-// Scope determines where skills are installed.
+// Scope determines where the plugin is installed.
 type Scope string
 
 const (
-	// ScopeUser installs into the user's agent config directories (~/.claude, ~/.agents).
-	// Agents whose config directories do not exist are skipped.
+	// ScopeUser installs the plugin at user scope (~/.claude).
 	ScopeUser Scope = "user"
-	// ScopeProject installs into the project's agent config directories (.claude, .agents).
-	// Directories are created as needed; no pre-existing config dir is required.
+	// ScopeProject installs the plugin at project scope (.claude in the project root).
 	ScopeProject Scope = "project"
 )
 
-// State describes the installation state of a skill for a specific agent.
-type State string
+// pluginName is the name of the CircleCI plugin in the marketplace.
+const pluginName = "circleci"
 
-// Skill installation states.
-const (
-	StateMissing  State = "missing"
-	StateCurrent  State = "current"
-	StateOutdated State = "outdated"
-)
-
-// Skill defines an embedded skill with its metadata.
-type Skill struct {
-	Name        string
-	Description string
-}
-
-// All is the ordered list of bundled skills.
-var All = []Skill{
-	{
-		Name:        "chunk-testing-gaps",
-		Description: `Use when asked to "find testing gaps", "chunk testing-gaps", "mutation test", "mutate this code", or "find surviving mutants". Runs a 4-stage mutation testing process.`,
-	},
-	{
-		Name:        "chunk-review",
-		Description: `Use when asked to "review recent changes", "chunk review", "review my diff", "review this PR", or "review my changes". Applies team-specific review standards from .chunk/review-prompt.md.`,
-	},
-	{
-		Name:        "debug-ci-failures",
-		Description: `Debug CircleCI build failures, analyze test results, and identify flaky tests. Use when asked to "debug CI", "why is CI failing", "fix CI failures", "find flaky tests", or "check CircleCI".`,
-	},
-	{
-		Name:        "chunk-sidecar",
-		Description: `Run build/test/validate on a remote chunk sidecar instead of locally. Use when asked to "validate on the sidecar", "run tests on the sidecar", "sync to sidecar", "check this on the sidecar", "run smarter testing doctor", "diagnose smarter testing", or when edits need remote verification. Also covers creating sidecars, snapshots, and env customization.`,
-	},
-	{
-		Name:        "chunk-sidecar-setup",
-		Description: `Interactive onboarding wizard for setting up a chunk sidecar from scratch. Use when asked to "set up chunk sidecar", "onboard to chunk", "first time sidecar", "configure sidecar from scratch", "walk me through sidecar setup", or "I've never used a chunk sidecar before". Covers auth, orgID, sidecar creation, dependency install, snapshot creation, and handoff to the dev loop.`,
-	},
-}
-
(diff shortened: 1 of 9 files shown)


michael-webster (commented):
I think we want to use the chunk plugin instead of circleci since we have the limitation of only being able to install hooks at a plugin level.

internal/testing/env/env.go
```diff
@@ -37,11 +37,17 @@ func (e *TestEnv) Environ() []string {
 	configDir := filepath.Join(e.HomeDir, ".config")
 	dataDir := filepath.Join(e.HomeDir, ".local", "share")
 
+	path := os.Getenv("PATH")
+	if override, ok := e.Extra["PATH"]; ok {
+		path = override
+		delete(e.Extra, "PATH")
+	}
+
```
michael-webster, line 45: The `delete(e.Extra, "PATH"])` makes `Environ()` non-idempotent — any test that calls `RunCLI` (or `Environ()`) twice on the same `TestEnv` after setting `Extra["PATH"]` will silently lose the override on the second call.

circleci-cli solves this with a `withExtDir` pattern that operates on the env slice at call time rather than mutating the shared struct:

```go
// In the test helper:
func withFakeClaude(env *testenv.TestEnv, binDir string) testenv.RunOpts {
    opts := testenv.RunOpts{}
    for _, e := range env.Environ() {
        if strings.HasPrefix(e, "PATH=") {
            opts.Env = append(opts.Env, "PATH="+binDir+":"+strings.TrimPrefix(e, "PATH="))
        } else {
            opts.Env = append(opts.Env, e)
        }
    }
    return opts
}

// Call site:
result := binary.RunCLI(t, []string{"skill", "install"}, env, dir, withFakeClaude(env, binDir))
```

This keeps `Environ()` a pure read and avoids the delete. If `RunCLI` doesn't yet accept a `RunOpts` variadic, the simpler fix is just removing the `delete` — the override stays in `Extra` across calls and PATH is always prepended correctly.

michael-webster (commented):
Two changes needed here:
- Make sure we use the new chunk plugin instead of the circleci plugin
- For tests, instead of directly deleting the `PATH` env var, let's use the `withExitDir` approach that circleci-cli uses.

internal/skills/install.go
```diff
@@ -1,241 +1,174 @@
...
+		Updated:   make([]string, 0),
 	}
-	embedded, err := skills.Content.ReadFile(filepath.Join(s.Name, "SKILL.md"))
+
+	if _, err := exec.LookPath(agentClaude); err != nil {
+		result.Skipped = true
+		return result
+	}
+
+	args := []string{"plugin", "install", pluginName, "--yes", "--scope", string(scope)}
```
michael-webster, line 93: can we some kind of user confirmation before we run a command that skips the checks?

internal/skills/install.go
```diff
@@ -1,241 +1,174 @@
...
-		return StateMissing
+		result.Errors = append(result.Errors, strings.TrimSpace(string(out)))
+		return result
 	}
-	if string(existing) == string(embedded) {
-		return StateCurrent
+
+	outStr := string(out)
+	switch {
+	case strings.Contains(outStr, "already installed"):
```
danmux, line 102: Detecting 'already installed' by parsing the CLI's output string - if the phrasing changes, installs silently misclassify. Is there a structured signal (exit code or `--json` flag) available?

internal/skills/install.go
```diff
@@ -1,241 +1,174 @@
...
-	if len(subset) == 0 {
-		return nil
-	}
-	all := agents(scope, baseDir)
-	results := make([]AgentInstallResult, 0, len(all))
-	for _, agent := range all {
-		results = append(results, installForAgent(agent, subset))
+	out, err := exec.Command(agentClaude, "plugin", "list", "--json").CombinedOutput() //nolint:gosec
+	if err != nil {
+		return AgentStatus{Agent: agentClaude, Available: true, Skills: pluginSkillStatuses(StateMissing)}
```
danmux, line 122: When `plugin list --json` fails, returns `Available: true` with `StateMissing`. A transient auth or network error triggers a 'missing' status that could prompt a spurious reinstall. Should a command error set `Available: false`?

internal/skills/install.go
```diff
@@ -1,241 +1,174 @@
...
-
-	for _, agent := range all {
-		available := true
-		if agent.SkipIfAbsent {
-			if _, err := os.Stat(agent.ConfigDir); err != nil {
-				available = false
-			}
-		}
+// SkillState is kept for compatibility but always reflects the plugin's overall state.
+func SkillState(_ string, _ Skill) State {
```
danmux, line 167: `SkillState` always returns `StateMissing` regardless of actual state. Any caller expecting real data silently gets wrong results. Can it be removed or unexported if there are no real callers?

danmux (approved):

PR #548, opened 2026-09-01
Interactive logs and live resource usage for sidecars

jesseworld22:
## Summary

Adds interactive logs and live resource usage to `chunk watch`.

- The watch daemon buffers a registered command's output, so it outlives the process that submitted it — for a hook-driven run, that process exits immediately.
- `chunk watch` gains a scrollback pane over one command's output and a CPU/memory/disk line per sidecar.
- `chunk sidecar logs <command-id>` prints the same output for readers with no terminal, from the daemon's buffer or the API.
- Both `chunk validate --remote` and `chunk sidecar exec` register with the daemon, so any command run on a sidecar is readable afterwards.
- `Exec` splits into `SubmitExec` + `StreamOutput` so the command ID exists before output is consumed. No new API surface.

Buffers are in-memory: a daemon restart loses them. Durable recovery needs the event log to carry `command_id`.

## Test plan
- [x] Ran `chunk validate --remote` against a real sidecar with a daemon attached: command registered before output was consumed, `sidecar logs` tailed it mid-run, and output survived the CLI exiting
- [x] Verified live that both `validate` and `sidecar exec` register and replay output after the submitting process exits
- [x] Resource samples verified on 10 live sidecars (cgroup memory, not host)
- [x] Checked `chunk watch` by hand: pane renders, and Enter opens output for both validate and exec rows
- [x] Merged `main` (rsync work) and re-ran the suite


docs/DESIGN_INTERACTIVE_LOGS.md
hanabel1, line 1: you can remove this file. looks like it created a proposal?

internal/cmd/sidecarlogs.go
```diff
@@ -0,0 +1,139 @@
...
+				}
+				if !chunk.Running || !follow {
+					return reportExitCode(chunk.ExitCode, io)
+				}
+				return followFromDaemon(cmd, commandID, chunk.NextOffset, io)
+			}
+
+			// No daemon, or it has forgotten this command: stream from the API.
+			insecureStorage := insecureStorageFlag(cmd)
+			rc, _ := config.Resolve("", "", insecureStorage)
```
hanabel1: the code ignores the error from config.Resolve - if that call fails the error is thrown away and the code keeps going with a broken/empty config. When it eventually doesn't work later, the error message won't tell you why it failed or where it came from.

Fix:
```
rc, err := config.Resolve("", "", insecureStorage)
if err != nil {
    return fmt.Errorf("resolve config: %w", err)
}
```

internal/cmd/sidecarlogs.go
```diff
@@ -0,0 +1,139 @@
...
+// caller. stdout therefore stays pure output for piping, and the status goes to
+// stderr where a human still sees it. A nil code means the command never
+// terminated as far as the reader could tell, which is not a failure either.
+func reportExitCode(code *int, io iostream.Streams) error {
+	if code == nil || *code == 0 {
+		return nil
+	}
+	io.ErrPrintf("exit status %d\n", *code)
+	return nil
+}
```
hanabel1, line 145: A script doing `chunk sidecar logs <id> || handle_failure` silently swallows the failure because the CLI always exits 0. The stderr message implies an intent to propagate failure. Was returning nil unconditionally intentional? If so, the docs should say logs never exits non-zero regardless of remote status. If not, a usererr wrapping the exit code would make it composable in pipelines.

internal/watchd/output.go
```diff
@@ -0,0 +1,303 @@
...
+func (s *outputStore) register(reg CommandReg, stream streamFn) {
+	if reg.CommandID == "" {
+		return
+	}
+	s.mu.Lock()
+	if _, exists := s.cmds[reg.CommandID]; exists {
+		s.mu.Unlock()
+		return
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
hanabel1: These goroutines use context.Background() instead of the daemon's own context, so when the daemon shuts down, it can't cancel them. They'll keep streaming from the CircleCI API until each command naturally finishes. If there are many long-running commands in flight, shutdown could hang. Worth either passing the daemon's context down here, or making sure stopAll() is called before the server closes.

internal/cmd/execregister.go
```diff
@@ -0,0 +1,67 @@
...
+// output being lost rather than never captured.
+func submitAndStream(
+	ctx context.Context, client *circleci.Client,
+	reg watchd.CommandReg, command string, args []string,
+	env map[string]string, onOutput circleci.OutputFn,
+) (*circleci.ExecResponse, error) {
+	commandID, err := client.SubmitExec(ctx, reg.SidecarID, command, args, env)
+	if err != nil {
+		return nil, err
+	}
```
hanabel1, line 38: When SubmitExec fails, the error is returned as-is with no context added. The team standard is to wrap errors ( `fmt.Errorf("submit exec: %w", err)` ) so that when it bubbles up, you know where it came from and can tell it apart from a streaming failure.

hanabel1:
the test plan covers chunk validate --remote run manualy but its also prob also worth triggering a validate through a hook and confirming it shows up in chunk watch.

jesseworld22:
Split into four stacked PRs, per review feedback — each builds and passes tests on its own:

- #551 — `SubmitExec`/`StreamOutput` split (193 lines, pure refactor)
- #552 — daemon output buffering
- #553 — scrollback pane + `chunk sidecar logs`
- #554 — live CPU/memory/disk sampling

They're stacked, so each is based on the one above and they merge in order. #551 is the quick one and unblocks the rest.

The four review threads here were all against code that now lives in #552, and their fixes came across with it. Closing in favour of the split.

PR #549, opened 2026-09-01
Route validate through watch daemon

hanabel1:
**Summary** 
- chunk validate routes through the daemon when it's running 
- daemon serializes validate runs running at the same time so two sessions can't stomp on each other on the same sidecar
- falls back to inline if the daemon isn't running

**Implementation**
- new POST /validate endpoing on the daemon which forks the current binary with sync to run inline and avoid recursion
- sync.mutex on the daemon serializes concurrent requests
- sync, hooks-session-id, and stop-hook-active flags carry hook context from client -> daemon -> subprocess
- hook-invoked run also delegate so the subprocess prints the session header 

**Test Plan**
- [x] chunk validate with daemon running → output appears, exit code is correct (zero and non-zero)
- [x] chunk validate with daemon not running → still works (falls through to inline execution)
- [x] chunk validate with a hook firing while daemon is running → hook output lands in the hook feedback block, not stdout
- [ ] Two concurrent chunk validate calls with daemon up → second one waits, not crashes (serialized by validateMu)
- [ ] Existing validate_test.go suite passes with --sync flag behavior intact (daemon never re-delegates)
Screenshot 2026-09-02 at 12 03 47 PM

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -150,22 +150,24 @@ func reportSkippedAutofix(skipped []string, streams iostream.Streams) {
 }
 
 type validateOpts struct {
-	sidecarID    string
-	identityFile string
-	workdir      string
-	orgID        string
-	dryRun       bool
-	list         bool
-	save         bool
-	remote       bool
-	local        bool
-	markRemote   bool
-	jsonOut      bool
-	inlineCmd    string
-	projectDir   string
-	envVarsFlag  []string
-	envFile      string
-	sync         bool // bypasses daemon delegation; set by daemon when it spawns us
+	sidecarID      string
+	identityFile   string
+	workdir        string
+	orgID          string
+	dryRun         bool
+	list           bool
+	save           bool
+	remote         bool
+	local          bool
+	markRemote     bool
+	jsonOut        bool
+	inlineCmd      string
+	projectDir     string
+	envVarsFlag    []string
+	envFile        string
+	sync           bool   // bypasses daemon delegation; set by daemon when it spawns us
+	hookSessionID  string // hook session ID forwarded from client to daemon subprocess
+	stopHookActive bool   // stop_hook_active forwarded from client to daemon subprocess
 }
 
 func newValidateCmd() *cobra.Command {
@@ -206,6 +208,10 @@ func newValidateCmd() *cobra.Command {
 	cmd.Flags().StringVar(&opts.envFile, "env-file", defaultEnvFile, "Env file to load (default: .env.local; pass a path to override)")
 	cmd.Flags().BoolVar(&opts.sync, "sync", false, "")
 	_ = cmd.Flags().MarkHidden("sync")
+	cmd.Flags().StringVar(&opts.hookSessionID, "hook-session-id", "", "")
+	_ = cmd.Flags().MarkHidden("hook-session-id")
+	cmd.Flags().BoolVar(&opts.stopHookActive, "stop-hook-active", false, "")
+	_ = cmd.Flags().MarkHidden("stop-hook-active")
 
 	cmd.AddCommand(newValidateVariantsCmd())
 
@@ -343,7 +349,17 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 	}
 
 	hook := detectHook(cmd.InOrStdin())
+	// When running as a daemon subprocess, hook context arrives via flags.
+	if opts.hookSessionID != "" && hook == nil {
+		hook = &hookContext{sessionID: opts.hookSessionID, stopHookActive: opts.stopHookActive}
+	}
 	ctx := cmd.Context()
+
+	// Delegate hook runs to the daemon before initHook so the subprocess prints
+	// the session header (not the client, which would cause it to appear twice).
+	if done, err := tryHookDelegate(cmd, hook, opts.sync, streams); done {
+		return err
+	}
 
 	// The working-tree fingerprint answers both hook-only questions about the
 	// tree: whether there is anything to validate at all, and whether this exact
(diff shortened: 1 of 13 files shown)


internal/cmd/validate.go
```diff
@@ -598,6 +602,79 @@ func validateEnvFlag(envVarsFlag []string) error {
...
+	if hook == nil || sync || !watchd.IsDaemonRunning() {
+		return false, nil
+	}
+	// If hooks are disabled in this environment, don't delegate — the daemon
+	// subprocess won't inherit this process's env var, so initHook must handle it.
+	if os.Getenv(config.EnvChunkHooksDisabled) != "" {
+		return false, nil
+	}
+	insecureStorage := insecureStorageFlag(cmd)
+	rc, _ := config.ResolveCircleCI(insecureStorage)
```
michael-webster: The error from `ResolveCircleCI` is silently discarded here. If token resolution fails (no token configured, keychain error), `rc.CircleCIToken` is empty-valued and gets forwarded to the daemon as-is — the subprocess runs without auth and the user gets a silent failure.

The non-hook path goes through `maybeEnsureCircleCIClient` which handles errors explicitly (prompts, propagates). This path bypasses all of that.

Suggested fix: return `false, nil` on error to fall back to inline execution, which has proper auth handling:
```go
rc, err := config.ResolveCircleCI(insecureStorage)
if err != nil {
    return false, nil // fall back to inline; inline path handles auth
}
```

internal/watchd/client.go
```diff
@@ -75,6 +75,43 @@ func stopDaemon(pid int, sockPath string) error {
...
+	if err != nil {
+		return ValidateResponse{}, err
+	}
+	body, err := json.Marshal(ValidateRequest{Args: args, CircleCIToken: circleCIToken, Env: os.Environ()})
+	if err != nil {
+		return ValidateResponse{}, fmt.Errorf("marshal validate request: %w", err)
+	}
+	resp, err := longUnixClient(sockPath).Post("http://watchd/validate", "application/json", bytes.NewReader(body))
+	if err != nil {
+		return ValidateResponse{}, fmt.Errorf("connect to watch daemon: %w", err)
```
michael-webster: There's a TOCTOU window between the `IsDaemonRunning()` check in the callers and this HTTP POST. If the daemon stops between the ping succeeding and this `Post` call, the user gets an opaque `"connect to watch daemon: ..."` error instead of falling back to inline execution.

In `tryHookDelegate`, the caller returns `(true, err)` on this error path — the `done=true` prevents any inline fallback at the call site. Neither caller currently catches dial errors as a signal to retry inline.

Consider returning a typed sentinel error (e.g. `ErrDaemonUnavailable`) so callers can distinguish a mid-flight connection failure from a real validate error and fall through to inline execution:
```go
var ErrDaemonUnavailable = errors.New("daemon unavailable")

// in RunValidate:
if err != nil {
    return ValidateResponse{}, fmt.Errorf("%w: %w", ErrDaemonUnavailable, err)
}
```

internal/cmd/validate.go
```diff
@@ -150,21 +151,24 @@ func reportSkippedAutofix(skipped []string, streams iostream.Streams) {
...
+	save           bool
+	remote         bool
+	local          bool
+	markRemote     bool
+	jsonOut        bool
+	inlineCmd      string
+	projectDir     string
+	envVarsFlag    []string
+	envFile        string
+	sync           bool   // bypasses daemon delegation; set by daemon when it spawns us
```
michael-webster: could we name use useDaemon instead of sync? just makes it clearer that we're controlling the daemon usage.

internal/watchd/ipc.go
```diff
@@ -48,3 +49,15 @@ func unixClient(sockPath string) *http.Client {
 		},
 	}
 }
+
+// longUnixClient is like unixClient but with no total-request timeout, suitable
+// for long-running operations like /validate.
+func longUnixClient(sockPath string) *http.Client {
```
michael-webster, line 103: this is probably okay given the time we're targeting for validation, but probably need to consider polling in the future.

internal/watchd/validate.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		http.Error(w, "executable: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	// Append --sync so the subprocess skips daemon delegation and runs inline.
+	// The socket is owner-only (0700) and exe is the current binary, so the
+	// args are not an injection risk.
+	args := append(append([]string(nil), req.Args...), "--sync")
+	cmd := exec.CommandContext(r.Context(), exe, args...) //nolint:gosec
```
michael-webster: do we actually need to spawn the binary here or could we just call the internal validate logic?

michael-webster (changes requested):
I don't think we want to completely spawn the chunk cli to validate from the daemon. I think we just want to call whatever validate would do from the daemon as library code.

internal/watchd/validate.go
```diff
@@ -0,0 +1,68 @@
...
+	if id := session.IDFromSlice(req.Env); id != "" {
+		ctx = session.WithID(ctx, id)
+	}
+	if req.CircleCIToken != "" {
+		req.Env = append(append([]string(nil), req.Env...), "CIRCLE_TOKEN="+req.CircleCIToken)
+	}
+	ctx = envctx.WithEnv(ctx, req.Env)
+
+	var stdout, stderr bytes.Buffer
+	exitCode := d.runner(ctx, req.Args, req.Env, &stdout, &stderr)
```
danmux, line 66: `d.runner` can be nil - `RunDaemon(ctx, nil)` is called in tests. Any POST to `/validate` on a nil-runner daemon panics. Needs a nil guard?

internal/envctx/envctx.go
```diff
@@ -0,0 +1,44 @@
...
+// Subsequent Getenv and Environ calls on that ctx consult env first.
+func WithEnv(ctx context.Context, env []string) context.Context {
+	if len(env) == 0 {
+		return ctx
+	}
+	return context.WithValue(ctx, envKey{}, env)
+}
+
+// Getenv returns the value of key from the env slice stored in ctx.
+// Falls back to os.Getenv when ctx carries no slice or the key is absent.
```
danmux: Doc says absent key falls back to `os.Getenv`, but the impl returns `""` instead. Is the fallback intentionally omitted once a slice is present, or is the doc wrong?

internal/watchd/validate.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var req ValidateRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, "decode request: "+err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// Seed the context with the caller's env and session ID so the in-process
+	// validate run behaves as if launched in the caller's environment.
+	ctx := r.Context()
```
danmux: `r.Context()` is canceled when the HTTP connection drops while `validateMu` is still held - the in-process runner gets interrupted mid-run. Intentional, or should this use `context.WithoutCancel`?

danmux (approved):

michael-webster (approved):

PR #552, opened 2026-09-02
Buffer remote command output in the watch daemon

jesseworld22:
**2 of 4** — splitting #548. Base: `main` — #551 has landed, and this is rebased on #543 (the `internal/tui` → `internal/ui` collapse).

## Summary

The watch daemon owns a registered command's output stream, so output outlives the process that submitted it — for a hook-driven run, that process exits immediately.

- Per-command buffer capped in bytes, keeping the tail and reporting truncation.
- Registration is best-effort and never starts the daemon: it sits on the hook path in front of a command the developer is waiting on.
- `chunk validate --remote` and `chunk sidecar exec` both register through one helper.
- Buffers are in memory; a daemon restart loses them.
- The daemon's CircleCI client is resolved once by the `_daemon` subcommand and passed into `RunDaemon`, rather than resolved lazily inside the daemon behind a mutex.

## Test plan
- [x] Live against a real sidecar: registered before output was consumed, output survived the CLI exiting — *verified before the credential refactor*
- [x] Stop-hook path — registered while running, replayable after the hook exited — *verified before the credential refactor*
- [ ] Re-run both live paths against the resolved-in-cmd client
- [ ] Unauthenticated daemon: command still recorded, output pane explains why it is empty, `AuthError` surfaced
- [ ] `chunk auth login` with a `chunk watch` on screen: daemon is replaced and output starts streaming
- [ ] Expired/rotated token mid-session: 401 reloads and the stream resumes without restarting the daemon


michael-webster:
The `credentials` type does a lot of work to lazily resolve a CircleCI client inside the daemon — but a background process with no terminal can't approve keychain prompts anyway, so the lazy init machinery handles a case that can't succeed. Whether resolution happens lazily on first command or eagerly at startup, the outcome on a keychain-approval-required machine is the same: `ErrNeedsAuth`, no output streaming.

I think the simpler design is to resolve credentials in the `cmd/` layer before calling `RunDaemon`, pass the resolved `*circleci.Client` in (or nil if unauthenticated), and delete `credentials` entirely. The daemon records commands either way; a nil client just means no output is streamed, which the existing `AuthError` field on `Snapshot` already surfaces to the user. Token refresh on expiry belongs inside `circleci.Client` (transparent 401 retry) rather than here.

This also eliminates the mutex-around-slow-keychain-I/O problem on the hook path. Tracked in FACT-460.

internal/httpcl/client.go
```diff
@@ -95,6 +111,41 @@ type Client struct {
...
+// reload re-reads the stored credential after used was rejected, and reports
+// whether the client now holds a different one — i.e. whether a retry is worth
+// making.
+//
+// The lock is held across the reload so that concurrent 401s produce one read
+// rather than one per caller: reading can mean a keychain round trip, and a
+// daemon streaming several commands can hit this from several goroutines at
+// once. Whoever arrives second sees the token has already moved and retries
+// without reading again.
+func (c *Client) reload(used string) bool {
```
hanabel1, line 140: reload() acquires the write lock before calling c.reloadToken(), which reads from disk or the OS keychain. While that lock is held, every goroutine calling c.token() (read lock) blocks - including all concurrent in-flight streaming connections in the daemon. The collapse logic is correct, but the I/O can happen outside the lock:

```
func (c *Client) reload(used string) bool {
    if c.reloadToken == nil {
        return false
    }
    c.mu.RLock()
    cur := c.authToken
    c.mu.RUnlock()
    if cur != used {
        return true
    }
    tok, err := c.reloadToken() // I/O outside the lock
    if err != nil || tok == "" || tok == used {
        return false
    }
    c.mu.Lock()
    defer c.mu.Unlock()
    if c.authToken != used {
        return true // lost the race
    }
    c.authToken = tok
    return true
```

hanabel1, line 140: reload() acquires the write lock before calling c.reloadToken(), which reads from disk or the OS keychain. While that lock is held, every goroutine calling c.token() (read lock) blocks - including all concurrent in-flight streaming connections in the daemon. The collapse logic is correct, but the I/O can happen outside the lock:

```
func (c *Client) reload(used string) bool {
    if c.reloadToken == nil {
        return false
    }
    c.mu.RLock()
    cur := c.authToken
    c.mu.RUnlock()
    if cur != used {
        return true
    }
    tok, err := c.reloadToken() // I/O outside the lock
    if err != nil || tok == "" || tok == used {
        return false
    }
    c.mu.Lock()
    defer c.mu.Unlock()
    if c.authToken != used {
        return true // lost the race
    }
    c.authToken = tok
    return true
```

internal/watchd/daemon.go
```diff
@@ -29,10 +32,41 @@ type projectState struct {
...
+	out *outputStore
+}
+
+// authMessage renders a credential-resolution failure for the dashboard. An
+// empty logs pane with no explanation sends people hunting the wrong fault.
+func authMessage(err error) string {
+	if err == nil {
+		return ""
+	}
+	if errors.Is(err, authprompt.ErrNeedsAuth) {
```
danmux: `watchd` importing `authprompt` to check `authprompt.ErrNeedsAuth` is a lateral business-package dependency. The `cmd/` caller already holds `authErr` - could it format the auth message string there and pass a plain string, or define the sentinel in `watchd` itself?

internal/watchd/output.go
```diff
@@ -0,0 +1,326 @@
...
+	endedAt  time.Time
+	// streamErr records why streaming stopped early, so an empty pane can say
+	// why it is empty instead of looking like a command that printed nothing.
+	streamErr string
+}
+
+func newBuffer() *buffer {
+	return &buffer{running: true}
+}
+
```
danmux, line 80: When `len(p) > MaxCommandBytes`, `append(b.data, p...)` first allocates `len(b.data) + len(p)` before trimming. A single large chunk temporarily spikes allocation well above the 256KB cap. Shouldn't the `len(p) > MaxCommandBytes` case be handled before the append?

internal/httpcl/client.go
```diff
@@ -95,6 +111,41 @@ type Client struct {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.authToken != used {
+		return true
+	}
+	tok, err := c.reloadToken()
+	// A failed reload is not worth surfacing: the caller already has a 401,
+	// which is the more useful error of the two.
+	if err != nil || tok == "" || tok == used {
+		return false
```
danmux, line 160: `c.mu.Lock()` is held across `c.reloadToken()`, which can be a keychain round trip. Every concurrent `c.token()` read-lock call blocks for the duration. For the daemon streaming multiple commands, all streams stall during a single 401 recovery. Is the one-reload guarantee worth blocking all concurrent reads?

internal/httpcl/client_test.go
```diff
@@ -468,3 +471,189 @@ func TestRetries504Exhausted(t *testing.T) {
...
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{}`))
+	}))
+	defer srv.Close()
+
+	var reloads atomic.Int32
+	cl := hc.New(hc.Config{
```
danmux, line 399: New reload tests use `t.Fatalf`/`t.Errorf` directly - inconsistent with the rest of the file and the project's `gotest.tools/v3/assert` convention.

danmux (approved):

michael-webster (approved):

PR #553, opened 2026-09-02
Add a scrollback pane and sidecar logs

jesseworld22:
**3 of 4** — splitting #548. Base: `jesse/logs-2-daemon-buffer`.

## Summary

Consumes the buffer the previous PR fills.

- `chunk watch` gains a scrollback pane over one command's output: `Enter` opens, `g`/`G` jump to top / re-follow, `Esc` closes the pane rather than quitting. Scrolling up detaches so arriving output does not yank the view away.
- Carriage returns are resolved, so a progress bar renders as its last frame rather than a hundred stacked copies.
- `chunk sidecar logs <command-id>` is the same data for readers with no terminal. It exits non-zero only when *reading* failed.

## Test plan
- [x] `chunk watch` by hand: pane renders, `Enter` opens output for validate and exec rows
- [x] `chunk sidecar logs` replayed a command after the submitting process exited

### DIFF
--- a/internal/cmd/sidecarlogs.go
+++ b/internal/cmd/sidecarlogs.go
@@ -1,0 +1,145 @@
+package cmd
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/config"
+	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
+	"github.com/CircleCI-Public/chunk-cli/internal/ui"
+	"github.com/CircleCI-Public/chunk-cli/internal/watchd"
+)
+
+// newSidecarLogsCmd prints a remote command's output.
+//
+// This is the non-TUI door onto the same data `chunk watch` shows, which matters
+// because the most common reader here has no terminal at all: an agent that just
+// ran validate and wants to know why it failed can pipe this, where it cannot
+// drive a dashboard.
+//
+// It prefers the watch daemon's buffer and falls back to the API, so it works
+// whether or not a daemon is running.
+func newSidecarLogsCmd() *cobra.Command {
+	var follow bool
+
+	cmd := &cobra.Command{
+		Use:   "logs <command-id>",
+		Short: "Print the output of a command that ran on a sidecar",
+		Long: "Print the output of a command that ran on a sidecar.\n\n" +
+			"Command IDs appear in the chunk watch dashboard. Output is read from the\n" +
+			"watch daemon's buffer when it has it, and from the CircleCI API otherwise.\n\n" +
+			"Exits non-zero only when reading failed. A command that itself exited\n" +
+			"non-zero is reported on stderr as \"exit status N\" and does not change this\n" +
+			"command's own status, so a failing remote command is not a failing read.",
+		Args:         cobra.ExactArgs(1),
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			commandID := args[0]
+			io := iostream.FromCmd(cmd)
+
+			// The daemon first: it needs no network round trip and it holds output
+			// for commands whose submitting process has long since exited.
+			if chunk, err := watchd.FetchOutput(commandID, 0); err == nil && chunk.Found {
+				if chunk.Truncated {
+					io.ErrPrintln("warning: earlier output was dropped from the buffer")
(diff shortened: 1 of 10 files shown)


internal/ui/watch/model.go
```diff
@@ -292,6 +335,19 @@ func (m Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.daemonErr = msg.err
 		return m, tea.Tick(pollInterval, func(time.Time) tea.Msg { return tickMsg{} })
 
+	case outputMsg:
+		return m.withOutputChunk(msg)
+
+	case outputTickMsg:
```
hanabel1, line 341: The outputTickMsg handler unconditionally reschedules itself whenever the pane is open — it doesn't check m.output.running. Once a command exits, every tick fires, fetchOutput returns Found:true, Running:false with no new data, withOutputChunk returns m, nil, and the tick fires again at 200ms forever. The eviction case (Found:false) has the same problem — the already-queued tick from the prior tea.Batch(...) still fires and reschedules. The comment in updateOutputKey shows explicit awareness of tick chain leaks ("starting another here would leave two chains polling"), which makes this omission notable. Should the handler gate the reschedule on m.output.running?

internal/ui/watch/output.go
```diff
@@ -0,0 +1,197 @@
...
+		} else {
+			canvas = append(canvas, r)
+		}
+		col++
+	}
+	return string(canvas)
+}
+
+// visibleLines returns the slice of scrollback to draw, plus whether the view is
+// showing the end of the output.
```
danmux, line 133: Allocates a full copy of `p.lines` (up to 5000 strings) every render frame when `pending != ""`. Runs on every 200ms tick and every keypress. Could just append inline: `append(all, resolveCR(p.pending))`?

internal/cmd/sidecarlogs.go
```diff
@@ -0,0 +1,145 @@
...
+			insecureStorage := insecureStorageFlag(cmd)
+			rc, err := config.Resolve("", "", insecureStorage)
+			if err != nil {
+				return fmt.Errorf("resolve config: %w", err)
+			}
+			client, err := ensureCircleCIClient(cmd.Context(), cmd, rc, io, ui.PromptHidden)
+			if err != nil {
+				return err
+			}
+			result, err := client.StreamOutput(cmd.Context(), commandID, "", func(_ string, data []byte) {
```
danmux, line 76: `--follow` is checked only in the daemon path. Does `StreamOutput` block until the command exits? If it returns early for a still-running command, `--follow` is silently a no-op on the API fallback path.

internal/cmd/sidecarlogs_test.go
```diff
@@ -0,0 +1,164 @@
...
+	got := remoteCommandLabel("cd '/repo' && " + long)
+	assert.Check(t, len(got) <= 120, "label should be capped, got %d chars", len(got))
+}
+
+func TestReportExitCode(t *testing.T) {
+	// Reading succeeded in every one of these cases, so none of them is an error:
+	// making `logs` fail because the command it reports on failed would leave the
+	// exit status meaning nothing.
+	var errOut strings.Builder
+	io := iostream.Streams{Out: &errOut, Err: &errOut}
```
danmux: `Out` and `Err` both point to `errOut`. `reportExitCode` writing to `Out` instead of `Err` would still pass the `Contains(errOut, ...)` assertion. Two separate buffers would catch the wrong stream.

internal/ui/watch/outputmodel.go
```diff
@@ -0,0 +1,237 @@
...
+// validate commands sequentially, so overlapping candidates are rare. When
+// nothing matches, no output affordance is shown — which is the honest outcome,
+// since the daemon genuinely has nothing for that invocation.
+//
+// Recording the command ID on the events themselves would make this exact rather
+// than probable; that is what the event log's command_id field is for once it
+// lands.
+func (m Model) commandForInvocation(g invocationGroup) *watchd.CommandState {
+	if len(g.events) == 0 || m.selectedIdx >= len(m.sidecars) {
+		return nil
```
danmux: `commandForInvocation` takes `g invocationGroup` as a parameter but resolves the sidecar via `m.selectedIdx`, not from `g`. If ever called for a non-selected invocation, it silently returns wrong results.

danmux (approved):

PR #554, opened 2026-09-02
Sample live CPU, memory and disk per sidecar

jesseworld22:
**4 of 4** — splitting #548. Base: `jesse/logs-3-scrollback`.

## Summary

Live CPU, memory and disk per sidecar in `chunk watch`.

- One long-lived sampler per sidecar, started only while a dashboard is attached.
- Runs over the async exec API rather than SSH, so there is no persistent connection for the daemon to supervise.
- Memory comes from cgroup files; `/proc/meminfo` reports the host's, which would make every sidecar look like it had hundreds of gigabytes free.
- Backoff and a cooloff after repeated failures, so a dead sidecar is not retried forever.

Most droppable piece of the stack — if it needs to go, don't merge it.

## Test plan
- [x] Resource samples verified live on 10 sidecars (cgroup memory, not host)
- [x] `chunk watch` by hand: resource line renders per row

### DIFF
--- a/internal/watchd/resources.go
+++ b/internal/watchd/resources.go
@@ -1,0 +1,558 @@
+package watchd
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/circleci"
+)
+
+const (
+	// SampleInterval is how often the remote sampler emits a reading.
+	SampleInterval = 2 * time.Second
+
+	// StaleSamples is how many intervals a sample may age before the dashboard
+	// should treat it as stale. A sampler that dies must look stalled rather than
+	// look like an idle sidecar, so the last value is kept and marked, not
+	// discarded.
+	//
+	// The budget has to cover more than SampleInterval: the reading crosses an SSH
+	// connection before the daemon sees it, and the dashboard then renders that
+	// same snapshot until its next 5s poll while re-evaluating the age every
+	// frame. Measured against real sidecars, a healthy sampler reaches ~10s of
+	// apparent age just before a poll lands, so a tighter bound flags working
+	// samplers as stale for part of every cycle. Twelve seconds clears that and
+	// still catches a dead sampler within two polls.
+	StaleSamples = 6
+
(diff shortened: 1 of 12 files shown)


internal/watchd/resources.go
```diff
@@ -0,0 +1,558 @@
...
+		if err == nil {
+			// The remote loop ended on its own; start another immediately.
+			failures = 0
+			backoff = time.Second
+			continue
+		}
+
+		failures++
+		if failures >= maxSamplerFailures {
+			log.Printf("watchd: pausing sampling of %s for %s after %d failures: %v",
```
danmux: `log.Printf` in a business-logic package - should these be propagated as errors or dropped, rather than going directly to stderr?

internal/watchd/resources.go
```diff
@@ -0,0 +1,558 @@
...
+			toStop = append(toStop, s)
+			delete(r.samplers, id)
+		}
+	}
+	// Started under the lock: a goroutine whose sampler was stopped between being
+	// recorded and adopting its cancel function could never be cancelled at all.
+	for _, req := range toStart {
+		ctx, cancel := context.WithCancel(context.Background())
+		req.sampler.setCancel(cancel)
+		go r.run(ctx, req.sidecar, req.sampler)
```
danmux: Goroutines launched while holding `r.mu`. `setCancel` also acquires `s.mu` under that lock, establishing `r.mu -> s.mu` ordering. If anything in the goroutine's call chain ever acquires `r.mu`, this deadlocks. Is holding the lock for the goroutine spawn strictly necessary?

internal/watchd/resources.go
```diff
@@ -0,0 +1,558 @@
...
+		consumeSamples(frames, s)
+		close(parsed)
+	}()
+	go func() {
+		frameErr <- readFrames(pr, frames)
+		close(frames)
+	}()
+
+	_, streamErr := client.StreamOutput(ctx, commandID, "", func(_ string, data []byte) {
+		_, _ = pw.Write(data)
```
danmux, line 516: `pw.Write(data)` inside the `StreamOutput` callback writes into an `io.Pipe` whose read side feeds an unbuffered `frames` channel. If `consumeSamples` is slow, the pipe fills and `pw.Write` blocks, stalling the callback. Does `StreamOutput` tolerate a blocking callback?

danmux (approved):

PR #558, opened 2026-09-03
allow opt-in autolaunch of the daemon

hanabel1:
## Summary

- Adds an `autoLaunchDaemon` user setting (default `false`) — users opt in rather than being surprised by a background process
- Adds `--daemon` / `--no-daemon` global flags to override the setting on any individual run
- Fixes a 404 stop-hook error caused by delegating to a stale daemon from an older build

## Events

- `chunk config set autoLaunchDaemon true` enables auto-launch globally
- `chunk validate --no-daemon` skips the daemon for that run even when the setting is on
- `chunk validate --daemon` forces daemon launch for that run even when the setting is off

## Not collected

- Auto-launch is not attempted for: shell completions, `chunk watch` (manages its own daemon), the `_daemon` subcommand, and `receive-telemetry`

## Implementation

- `IsDaemonCompatible()` added to `watchd` — checks reachability **and** build ID match; `shouldUseDaemon` and `tryHookDelegate` now use this instead of `IsDaemonRunning()`, so a stale daemon from a previous build is skipped rather than delegated to
- `RunValidate` treats a 404 response as `ErrDaemonUnavailable` (belt-and-suspenders fallback to inline execution)
- `--no-daemon` moved from a hidden local validate flag to a persistent root flag so it applies to all commands; validate reads it via `cmd.Flags().GetBool`

## Test plan

- [ ] `task build && task test` passes
- [ ] `chunk config set autoLaunchDaemon true` → subsequent commands start the daemon if not running
- [ ] `chunk validate --no-daemon` skips daemon even when setting is on
- [ ] Stopping the running daemon and running the stop hook with a stale (pre-/validate) daemon no longer errors with 404

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -14,6 +14,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/session"
 	"github.com/CircleCI-Public/chunk-cli/internal/telemetry"
 	"github.com/CircleCI-Public/chunk-cli/internal/upgrade"
+	"github.com/CircleCI-Public/chunk-cli/internal/watchd"
 )
 
 type updateCheckKey struct{}
@@ -46,7 +47,7 @@ func NewRootCmd(version string) *cobra.Command {
 				return err
 			}
 			startUpdateCheck(cmd)
-			return nil
+			return maybeAutoLaunchDaemon(cmd)
 		},
 		PersistentPostRunE: func(cmd *cobra.Command, _ []string) error {
 			printUpdateNotice(cmd)
@@ -108,6 +109,9 @@ Configuration:
 
 	rootCmd.PersistentFlags().Bool("insecure-storage", false, "do not use the system's secure storage for storing tokens")
 	_ = rootCmd.PersistentFlags().MarkHidden("insecure-storage")
+
+	rootCmd.PersistentFlags().Bool("daemon", false, "auto-launch the watch daemon for this run even if autoLaunchDaemon is disabled")
+	rootCmd.PersistentFlags().Bool("no-daemon", false, "skip the watch daemon for this run even if autoLaunchDaemon is enabled")
 
 	telemetry.RecordForSubcommands(rootCmd)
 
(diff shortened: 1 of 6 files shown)


internal/watchd/client.go
```diff
@@ -109,7 +122,14 @@ func RunValidate(args []string, circleCIToken string) (ValidateResponse, error)
 	defer func() { _ = resp.Body.Close() }()
 	if resp.StatusCode != http.StatusOK {
 		msg, _ := io.ReadAll(io.LimitReader(resp.Body, 512))
-		return ValidateResponse{}, fmt.Errorf("watch daemon returned %s: %s", resp.Status, bytes.TrimSpace(msg))
+		err := fmt.Errorf("watch daemon returned %s: %s", resp.Status, bytes.TrimSpace(msg))
+		// 404 means the daemon is running but does not have the /validate
+		// endpoint — it is from an older build. Treat it as unavailable so
+		// callers fall back to inline execution instead of surfacing the error.
+		if resp.StatusCode == http.StatusNotFound {
+			return ValidateResponse{}, fmt.Errorf("%w: %w", ErrDaemonUnavailable, err)
```
danmux, line 223: Comment says callers fall back to inline on 404 wrapping `ErrDaemonUnavailable` - is that fallback actually wired in `validate.go`, or does the 404 surface to the user?

internal/cmd/root.go
```diff
@@ -207,6 +212,51 @@ func startUpdateCheck(cmd *cobra.Command) {
...
+// maybeAutoLaunchDaemon starts the watch daemon if autoLaunchDaemon is
+// configured (or --daemon is passed) and the command is not excluded.
+// Errors are silently ignored: the daemon is an optimization and every command
+// works correctly without it.
+func maybeAutoLaunchDaemon(cmd *cobra.Command) {
+	if !shouldAutoLaunch(cmd) {
+		return
+	}
+	_ = watchd.EnsureRunning([]string{watchCmdName, watchDaemonSubcmd})
+}
```
danmux, line 265: `EnsureRunning` error is silently dropped even when the user explicitly passed `--daemon`. Shouldn't an explicit flag surface an error if the daemon fails to start?

internal/cmd/validate.go
```diff
@@ -335,14 +333,20 @@ func resolveWorkDir(opts *validateOpts) (string, error) {
...
-// us in-process.
+// us in-process. Delegation is skipped for daemons from a different build since
+// they may not support the /validate endpoint.
 func shouldUseDaemon(hook *hookContext, noDaemon bool) bool {
-	return hook == nil && !noDaemon && watchd.IsDaemonRunning()
+	return hook == nil && !noDaemon && watchd.IsDaemonCompatible()
 }
 
 func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) error {
 	streams := iostream.FromCmd(cmd)
```
danmux, line 344: `GetBool` error silently discarded - if `--no-daemon` isn't wired (e.g. test context without root persistent flags), `noDaemon` stays false and the call unexpectedly delegates to the daemon.

danmux (approved):

PR #560, opened 2026-09-04
fix(sidecar): stop forcing IdentitiesOnly=yes for rsync over SSH

### LINKED ISSUE #559: sidecar sync: rsync over SSH-WebSocket proxy fails with connection unexpectedly closed (0 bytes)
## Summary

`chunk sidecar sync` fails with rsync `connection unexpectedly closed (0 bytes received so far)` (exit 255 or 12), while `chunk sidecar exec` works on the **same** sidecar. Reproduced by a customer across multiple sidecars (including freshly created from snapshot) on chunk CLI `0.7.166` / `0.7.167` (macOS arm64, Homebrew rsync 3.5.0).

## Why this looks like a CLI bug

After #539, sync uses system `rsync` + OpenSSH pointed at a local TCP→WebSocket proxy (`internal/sidecar/rsync.go` `startSSHProxy` / `bridgeConn`). Exec uses golang `x/crypto/ssh` over WebSocket (`dialSSH` / `ExecOverSSH`).

Customer ruled out remote rsync version skew (rebuilt matching 3.5.0 on the sidecar). Disk/workspace/ownership checks via exec are fine. `bridgeConn` currently swallows WebSocket dial errors (closes TCP with no log), which matches the `0 bytes received` symptom.

## Environment

- chunk CLI: 0.7.166, 0.7.167
- Client: macOS 15.7.5 arm64, Homebrew rsync 3.5.0 (protocol 32)
- Sidecar: Ubuntu 24.04, stock rsync 3.2.7 (also tested with rebuilt 3.5.0)

## Error

```
Syncing workspace /home/user/vcr-cloud-desktop...
✗ Error: The sync operation failed.

rsync: exit status 255
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(285) [sender=3.5.0]
```

(also seen exit 12 / protocol data stream on another sidecar)

## Suggested direction

1. Surface `bridgeConn` dial/proxy errors instead of silent TCP close.
2. Prefer driving rsync over the same golang SSH transport as `exec` (custom `--rsh` / session stdio), or fall back to that path when OpenSSH-via-proxy fails.
3. Add `--verbose`/`--debug` for sync to print the constructed rsync/ssh command and proxy errors.

## Regression window

#539 merged 2026-09-01 (~`v0.7.165`). Pre-rsync (`v0.7.164` and earlier) used git-bundle sync over ExecOverSSH.

## Impact

Blocks `chunk validate` and git pre-commit hooks that wrap it. Preview feature, but hard blocker for affected developers.

nanophate:
## Summary

- Set `-o IdentitiesOnly=yes` only when passing an explicit `-i` identity file
- Set `-o IdentitiesOnly=no` explicitly for ssh-agent sessions, and pin the agent with `-o IdentityAgent=<sess.AuthSock>`
- Pin `ProxyCommand`, `ProxyJump`, `ControlPath` and `RequestTTY` so a user `~/.ssh/config` cannot hijack the `127.0.0.1` hop, rather than discarding the whole config with `-F /dev/null`
- Drop `-q` so ssh diagnostics reach the rsync error, and filter ssh's expected `Warning: Permanently added ...` line out of that error detail

## Why

`chunk sidecar sync` failed with `connection unexpectedly closed (0 bytes received)` when the sidecar session authenticated through ssh-agent, while `exec` on the same sidecar worked.

The `rsync -e ssh` command always passed `-o IdentitiesOnly=yes`, then appended `-o IdentitiesOnly=no` for agent sessions. OpenSSH honours the first occurrence of an option, so the second had no effect:

```
$ ssh -o IdentitiesOnly=yes -o IdentitiesOnly=no -G example.com | grep identitiesonly
identitiesonly yes
```

`OpenSession` registers the agent key with the sidecar, then ssh could not offer that key, exited 255, and `-q` hid the reason. rsync only saw an empty stream.

## Why not `-F /dev/null`

An earlier revision of this PR isolated the hop by discarding the user's config entirely. That also discards `/etc/ssh/ssh_config` and macOS `UseKeychain`, so a passphrase-protected `--identity-file` key has no way to obtain its passphrase and **blocks indefinitely on a prompt inside the rsync child** — a hang with no output. (`-F <unopenable path>` is also fatal in ssh, unlike `UserKnownHostsFile`, which matters for the Windows build target.)

It is also mostly unnecessary. Command-line options are parsed first, so a user config can never override `-p` or any `-o` set here; only options left unset can leak in:

```
Host * in user config          resolved
-----------------------------  ---------------------------
Port 9999                      port 2299            (our -p wins)
StrictHostKeyChecking yes      stricthostkeychecking false
UserKnownHostsFile ~/...       userknownhostsfile /dev/null
ProxyCommand /bin/false        proxycommand /bin/false   <- leaks
IdentityAgent /tmp/evil.sock   identityagent /tmp/evil.sock  <- leaks
IdentitiesOnly yes             identitiesonly yes            <- leaks
```

Pinning those three is enough, and keeps the rest of the user's config working. Note the third one is why the agent path now sets `IdentitiesOnly=no` explicitly rather than leaving it unset: a user config with `IdentitiesOnly yes` would otherwise suppress the agent key and reproduce this very bug.

Pinning `IdentityAgent` to `sess.AuthSock` is also what keeps the rsync child on the same agent `ExecOverSSH` dials (`ssh.go:279`); previously the child resolved `SSH_AUTH_SOCK` from the ambient environment, which diverges whenever a caller sets `variants.Options.AuthSock` to a non-ambient socket.

## Test plan

- [x] `go test -race ./...` (35 packages) and `task lint` (0 issues)
- [x] Unit: every pinned option asserted per session type; `IdentitiesOnly` set at most once; `-i` absent on the agent path (tokenised, so a trailing `-i` is caught); `rsyncErrDetail` filtering
- [x] Integration against a local `sshd` with only the agent key authorized, driving the strings the code actually generates, with a hostile `ssh_config` setting `IdentitiesOnly yes`, `ProxyCommand /bin/false`, a bogus `IdentityAgent`, `ControlMaster auto` and `RequestTTY yes`:
  - agent path → `exit=0`, file synced
  - identity-file path → `exit=0`, file synced
  - previous arguments under the same config → `Permission denied (publickey)`
- [x] Confirmed OpenSSH prefers agent keys, so omitting `-i` on the agent path does not flood the server with default `~/.ssh/id_*` offers: with five file identities configured, ssh made exactly one offer (the agent key)
- [ ] Manual: `chunk sidecar sync` on macOS with `SSH_AUTH_SOCK` set and no `--identity-file`
- [ ] Manual: `chunk sidecar sync --identity-file ~/.ssh/chunk_ai` with a passphrase-protected key in the macOS Keychain

Fixes #559

### DIFF
--- a/internal/sidecar/rsync.go
+++ b/internal/sidecar/rsync.go
@@ -97,21 +97,7 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
 		return fmt.Errorf("rsync: parse proxy addr: %w", err)
 	}
 
-	sshArgs := []string{"ssh", "-p", port,
-		"-o", "StrictHostKeyChecking=no",
-		"-o", "UserKnownHostsFile=/dev/null",
-		"-o", "IdentitiesOnly=yes", // prevent agent key flood before explicit key is tried
-		"-q",
-	}
-	if sess.IdentityFile != "" {
-		// Pass path directly — rsync tokenizes -e by whitespace and calls execve,
-		// so shell quoting (ShellEscape) would embed literal quote characters in
-		// the filename and cause ssh to fall through to agent keys.
-		sshArgs = append(sshArgs, "-i", sess.IdentityFile)
-	} else if sess.UseAgent && sess.AuthSock != "" {
-		sshArgs = append(sshArgs, "-o", "IdentitiesOnly=no")
-	}
-	sshCmd := strings.Join(sshArgs, " ")
+	sshCmd := sshCommand(sess, port)
 
 	src := strings.TrimRight(cwd, "/") + "/"
 	dst := fmt.Sprintf("%s@127.0.0.1:%s", defaultSSHUser, repoPath)
@@ -135,6 +121,32 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
 
 	status(iostream.LevelDone, "Synced")
 	return nil
+}
+
+// sshCommand builds the command rsync passes to -e in order to reach the
+// sidecar through the local proxy listening on port.
+//
+// IdentitiesOnly=yes is set only alongside an explicit -i. OpenSSH honours the
+// first occurrence of an option, so setting it unconditionally cannot be undone
+// later: an IdentitiesOnly=no appended for agent sessions is silently ignored,
+// leaving ssh unable to offer the agent key OpenSession registered. With no -i
+// to restrict, IdentitiesOnly=yes also does not mean "no keys" — ssh falls back
+// to the default ~/.ssh/id_* filenames, so it offers unrelated keys instead.
+//
+// -q is deliberately absent so ssh diagnostics reach the rsync error instead of
+// being swallowed.
+func sshCommand(sess *Session, port string) string {
+	args := []string{"ssh", "-p", port,
+		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
+	}
+	if sess.IdentityFile != "" {
+		// Pass path directly — rsync tokenizes -e by whitespace and calls execve,
+		// so shell quoting (ShellEscape) would embed literal quote characters in
+		// the filename and cause ssh to fall through to agent keys.
+		args = append(args, "-o", "IdentitiesOnly=yes", "-i", sess.IdentityFile)
+	}
+	return strings.Join(args, " ")
 }
 
 // startSSHProxy starts a local TCP listener on a random port and bridges each
--- a/internal/sidecar/rsync.go
+++ b/internal/sidecar/rsync.go
@@ -133,10 +133,11 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
 // to restrict, IdentitiesOnly=yes also does not mean "no keys" — ssh falls back
 // to the default ~/.ssh/id_* filenames, so it offers unrelated keys instead.
 //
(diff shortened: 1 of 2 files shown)


internal/sidecar/rsync.go
```diff
@@ -137,6 +123,69 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
...
+	if sess.IdentityFile != "" {
+		// Pass path directly — rsync tokenizes -e by whitespace and calls execve,
+		// so shell quoting (ShellEscape) would embed literal quote characters in
+		// the filename and cause ssh to fall through to agent keys.
+		args = append(args, "-o", "IdentitiesOnly=yes", "-i", sess.IdentityFile)
+	} else if sess.UseAgent && sess.AuthSock != "" {
+		// Set IdentitiesOnly=no explicitly: a user ssh_config with
+		// "IdentitiesOnly yes" would otherwise stop ssh offering the agent key
+		// OpenSession registered, reproducing the failure this helper fixes.
+		// Pinning IdentityAgent keeps the agent in step with ExecOverSSH, which
```
danmux, line 182: If `sess.AuthSock` contains spaces, rsync's whitespace split on the `-e` string corrupts this option into separate tokens. Same limitation as `-i` (noted at line 172) but not documented here?

danmux (approved):

PR #562, opened 2026-09-04
Record sidecar command ID in the event log

schurchleycci:
## Summary

- Adds `command_id` to `Event` — populated on the per-command pass/fail event for commands that ran on a sidecar, linking to the command's output via `GET /api/v3/sidecar/commands/{id}/output`
- Adds `Recorder.SetCommandID(id)` — called by `newExecFn` after each remote exec; `record()` attaches the pending ID to the next `LevelDone`/`LevelError` Status call and clears it, so each exec stamps exactly its own terminal event and never a run-wide summary
- Threads `*eventlog.Recorder` through `newExecFn`, `runValidate`, and `runSplitCommands`; clears the workspace probe's ID after `WorkspaceExists` so it cannot leak onto a real command's event
- Removes the redundant `statusFn` parameter from `runValidate` and `runSplitCommands` — both already receive `*eventlog.Recorder`, which exposes `Status` directly

## Test plan

- [ ] New eventlog tests: ID stamps the right event, ID consumed once, ID not placed on `Final`
- [ ] Full test suite green

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -472,7 +472,7 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		return failBeforeRun(rec, start, err)
 	}
 
-	result, execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, opts.workdir, allRemote, envVars, cfg, rec.Status, streams)
+	result, execErr := runValidate(ctx, circleCIClient, rc, workDir, name, opts.inlineCmd, opts.save, opts.sidecarID, opts.workdir, allRemote, envVars, cfg, rec.Status, rec, streams)
 	if execErr == nil && resultCache != nil {
 		if err := resultCache.Put(cacheKey, validate.CachedResult{CachedAt: time.Now()}); err != nil {
 			streams.ErrPrintf("  %s\n", ui.ErrDim(fmt.Sprintf("chunk validate: cache write failed: %v", err)))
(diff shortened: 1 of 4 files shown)


internal/eventlog/eventlog.go
```diff
@@ -186,6 +192,14 @@ func Record(dataDir string, fn iostream.StatusFunc, op Op, sidecarID, sidecarNam
 	return el.Recorder(fn, op, sidecarID, sidecarName, branch)
 }
 
+// SetCommandID records the sandbox-provisioner command ID of the exec that just
+// finished. The next pass/fail event for that command will carry this ID.
+// Pass "" when an exec failed without reporting an ID so a stale one from a
+// prior exec is not attributed to it.
+func (r *Recorder) SetCommandID(id string) {
```
danmux, line 200: Does this race if `validate.RunRemote` runs commands concurrently? `SetCommandID` writes `pendingCommandID` from inside `execFn` while `record()` reads and clears it from a `Status` call, with no mutex protecting the field.

internal/cmd/validate.go
```diff
@@ -877,31 +881,33 @@ func hostForwardEnv(token string) map[string]string {
...
-		statusFn(iostream.LevelInfo, fmt.Sprintf("running locally: %s", commandNames(localCfg.Commands)))
+		rec.Status(iostream.LevelInfo, fmt.Sprintf("running locally: %s", commandNames(localCfg.Commands)))
 	}
 	var combined validate.Result
 	var runErr error
 	if len(remoteCfg.Commands) > 0 {
-		execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, streams)
+		execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, rec, streams)
 		if err != nil {
 			return validate.Result{}, unreachableSidecar(sidecarID, commandNames(remoteCfg.Commands), err)
```
danmux, line 906: This manually clears the side-effect `execFn` left on `rec` after `WorkspaceExists`. If a future exec is inserted between `WorkspaceExists` and this line, the probe ID silently stamps the next real command's terminal event. Shouldn't `WorkspaceExists` take a plain exec func that never touches the recorder?

danmux (approved):

internal/eventlog/eventlog.go
schurchleycci: Added a mutex to protect the field 👍

internal/cmd/validate.go
schurchleycci: Good call, updated!

PR #564, opened 2026-09-04
Add sidecar pool primitives

z00b:
- add reusable Pool and Target abstractions for sidecar coordination
- refactor sync and session setup for fanout, stale probing, and retry-aware SSH key registration
- extend fake CircleCI and tests to support deterministic pool creation and reuse

### DIFF
--- a/internal/sidecar/pool.go
+++ b/internal/sidecar/pool.go
@@ -1,0 +1,406 @@
+package sidecar
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/circleci"
+	"github.com/CircleCI-Public/chunk-cli/internal/gitremote"
+	"github.com/CircleCI-Public/chunk-cli/internal/iostream"
+)
+
+// PoolEntry is a single ready-to-use sidecar checked out from a Pool.
+type PoolEntry struct {
+	ID       string
+	RepoPath string
+	Client   *circleci.Client
+}
+
+// Pool manages a fixed set of sidecars as a work queue for parallel tasks.
+type Pool struct {
+	free         chan *PoolEntry
+	updates      chan struct{}
+	ids          []string
+	entries      []*PoolEntry
+	client       *circleci.Client
+	workDir      string
+	orgID        string
+	image        string
+	name         string
+	identityFile string
+	authSock     string
+
+	mu           sync.Mutex
+	pendingSyncs int
+	checkedOut   int
+	syncErr      error
+}
+
+type poolState struct {
+	SidecarIDs    []string `json:"sidecar_ids"`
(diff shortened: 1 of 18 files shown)


internal/sidecar/session.go
```diff
@@ -77,21 +81,85 @@ type Session struct {
 	AuthSock     string // SSH_AUTH_SOCK path (only used when UseAgent is true)
 }
 
+// readProbeKey resolves the SSH public key to use for a staleness probe.
+func readProbeKey(authSock, identityFile string) (string, error) {
+	if identityFile == "" && authSock != "" {
+		if pubKey, err := agentPublicKey(context.Background(), authSock); err == nil {
+			return pubKey, nil
+		}
```
danmux, line 89: `readProbeKey` passes `context.Background()` to `agentPublicKey`. The 15s probe timeout in `IsDefinitelyStale` can't cancel this leg - if the agent socket stalls, the stale check blocks past the deadline.

internal/sidecar/pool.go
```diff
@@ -0,0 +1,385 @@
...
+	var state poolState
+	return &state, json.Unmarshal(data, &state)
+}
+
+func savePoolState(workDir, name string, state *poolState) {
+	data, err := json.Marshal(state)
+	if err != nil {
+		return
+	}
+	_ = os.WriteFile(poolStatePath(workDir, name), data, 0o600)
```
danmux: `savePoolState` silently discards the `os.WriteFile` error. If the state file isn't written, the next run recreates all sidecars from scratch with no warning.

internal/sidecar/pool.go
```diff
@@ -0,0 +1,385 @@
...
+		p.notifyUpdate()
+	}()
+}
+
+func (p *Pool) finishSync(entry *PoolEntry, err error) {
+	p.mu.Lock()
+	if err != nil {
+		p.syncErr = errors.Join(p.syncErr, fmt.Errorf("sidecar %s: %w", entry.ID, err))
+	} else {
+		p.free <- entry
```
danmux, line 392: `Rebuild` names every replacement `"%s-rebuilt"` - same name for all replacements. Is a unique suffix (timestamp, old ID, or index) intentionally omitted?

internal/sidecar/target.go
```diff
@@ -0,0 +1,142 @@
...
+		if stream == circleci.StreamStderr {
+			w = streams.Err
+		}
+		_, _ = w.Write(data)
+	}
+	execFn := func(ctx context.Context, script string) (string, string, int, error) {
+		result, err := t.Client.Exec(ctx, t.SidecarID, "sh", []string{"-c", script}, envVars, onOutput)
+		if err != nil {
+			return "", "", 0, err
+		}
```
danmux, line 82: `execFn` always returns `"", ""` for stdout/stderr, but the signature promises them. All output goes through the `onOutput` callback - callers expecting output via return values get nothing silently.

danmux (approved):

z00b:
addressed review comments and compacted into the original commit. merging.

PR #573, opened 2026-09-08
Write chunk validate results to the filesystem

jesseworld22:
**Standalone** — based on `main`, merges in any order.

This was the base of the #580 → #581 → #584 stack. That series no longer depends on it: the only pieces it borrowed (`config.CanonicalProjectRoot` and the `newTestDaemon` test helper) are now carried in #580, so whichever of the two lands second drops its copy.

---

## Summary

`chunk validate` results never showed up in `chunk watch` for a project that had never had a sidecar.

Nothing was ever lost — every run already writes to `events.jsonl`, and the daemon reads those files off disk. What was missing was **discovery**: the daemon only finds a project by its `project-root` breadcrumb, and only saving sidecar state wrote one. So `--local` runs, and repos with no remote commands, logged somewhere nothing would ever look. Now every run registers its project. The same gap was hiding buffered remote command output (#552), which is keyed by project root too.

Registering everywhere then exposed a second bug, spotted in the real dashboard: one project drawn as two rows over one log, because `ProjectDataDir` keys by the *resolved* path while the daemon keys by the breadcrumb *string*, and different writers spelled roots differently. An audit found two more instances of the same thing. All of them now go through one rule, `config.CanonicalProjectRoot`.

## Test plan

Every test below was confirmed to fail with the change it covers reverted.

- [x] A `--local` run registers its project and closes with a 1/1 tally; a breadcrumb plus a log and no sidecar state is discovered and replayed.
- [x] One project however its root is spelled — including breadcrumbs written before this change — and commands registered through a symlinked root are still listed (`got 0` before the fix).
- [x] `--focus` roots match the keys the daemon reports.
- [x] Real dashboard through a pty: `--local` and hook runs in sidecar-less repos render `✓ 1/1` and `✗ 0/1`; control case with the breadcrumb removed shows "No activity yet".
- [x] `task test` (1558), `task lint`, `task ci:diff`, CI green.

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -448,8 +448,6 @@ func runValidateCmdE(cmd *cobra.Command, args []string, opts *validateOpts) erro
 		// the IsDaemonCompatible check and the POST (connection refused), and the
 		// daemon lacks the /validate endpoint because it is from an older build
 		// (404). Both fall through to inline execution.
-		// daemon disappeared between the IsDaemonRunning check and the POST;
-		// fall through to inline execution.
 	}
 
 	// allRemote is true unless --local is passed explicitly. opts.remote is
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -541,12 +541,15 @@ func newValidateRecorder(statusFn iostream.StatusFunc, sidecarID string, activeS
 	}
 	// A missing data dir leaves the recorder reporting without recording.
 	dataDir, _ := config.ProjectDataDir(workDir)
-	// Register the project alongside the log about to be written to it. Nothing
-	// is streamed to the watch daemon — it reads these same files — and only
-	// saving sidecar state used to register a project, so a run with no sidecar
-	// (--local, or a repo with no remote commands) logged its results where no
-	// daemon would ever look for them. Best-effort: a run still records without
-	// the breadcrumb, and still reports without the log.
+	// Register the project alongside the log about to be written to it. A run's
+	// results are never sent to the watch daemon — it reads this same log off
+	// disk — and only saving sidecar state used to register a project, so a run
+	// with no sidecar (--local, or a repo with no remote commands) logged its
+	// results where no daemon would ever look for them. Registered remote
+	// commands were reached the same way: the daemon buffers their output under a
+	// project root it lists only once it has discovered that project.
+	// Best-effort: a run still records without the breadcrumb, and still reports
+	// without the log.
 	_ = sidecar.RegisterProjectRoot(dataDir, workDir)
 	return eventlog.Record(dataDir, statusFn, op, sidecarID, scName, sidecar.CurrentBranch(workDir))
 }
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -540,6 +540,13 @@ func newValidateRecorder(statusFn iostream.StatusFunc, sidecarID string, activeS
 	}
 	// A missing data dir leaves the recorder reporting without recording.
 	dataDir, _ := config.ProjectDataDir(workDir)
+	// Register the project alongside the log about to be written to it. Nothing
+	// is streamed to the watch daemon — it reads these same files — and only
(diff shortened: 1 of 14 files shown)


schurchleycci:
Couple minor comments:

Env var constant not used in daemon tests (internal/watchd/daemon_test.go, line 1072)

t.Setenv("XDG_DATA_HOME", ...) uses a raw string; validate_test.go correctly uses config.EnvXDGDataHome. If the constant is renamed, these tests silently stop isolating config.

Stale/duplicate comment (internal/cmd/validate.go, ~line 522–525)

After the new explanatory block, the old // daemon disappeared between the IsDaemonRunning check and the POST comment remains (and now references the renamed function). Two overlapping comments explaining the same fallthrough — consolidate into one.

internal/watchd/output.go
```diff
@@ -208,6 +209,11 @@ func (s *outputStore) register(reg CommandReg, stream streamFn) {
 	if reg.CommandID == "" {
 		return
 	}
+	// Canonicalised here rather than at each client: the clients send the working
+	// directory they were handed, project discovery canonicalises, and a command
+	// filed under a spelling no project carries is output the dashboard cannot
+	// reach. Doing it on ingest also covers clients built before this.
+	reg.ProjectRoot = config.CanonicalProjectRoot(reg.ProjectRoot)
```
hanabel1, line 216: CanonicalProjectRoot("") resolves "." to the process's working directory. A command registered with no root would then appear under whichever project happens to live there. Add a guard:

```
if reg.ProjectRoot != "" {
    reg.ProjectRoot = config.CanonicalProjectRoot(reg.ProjectRoot)
}
```

internal/watchd/daemon_test.go
```diff
@@ -153,3 +216,83 @@ func TestStopForCredentialChangeIsANoopWithNoDaemon(t *testing.T) {
...
+
+// The duplicate rows this guards against were found by hand: one project drawn
+// twice, the two rows sharing a single event log. The daemon keys projects by
+// the breadcrumb string while ProjectDataDir keys the directory by the resolved
+// path, so two spellings of one root share a log but count as two projects. It
+// took a symlinked path to see it, which on darwin is every temp directory and
+// on linux is none — so the symlink here is built rather than assumed, and the
+// spelling is flipped between polls the way a validate run and chunk watch used
+// to flip it.
+func TestPollListsOneProjectPerRootHoweverItIsSpelled(t *testing.T) {
```
hanabel1, line 228: "project-root" is spelled out directly instead of referencing sidecar.projectRootFile. If the filename changes, the test silently passes with zero projects instead of catching the regression. Export projectRootFile or add a test helper in sidecar so watchd tests can reference the canonical name.

internal/cmd/validate.go
```diff
@@ -653,12 +653,20 @@ func ensureRequestedValidateCommand(workDir, name, inlineCmd string, cfg *config
 // wrapEventLogStatusFn wraps statusFn with event log recording when a sidecar
 // is active. Returns statusFn unchanged when no sidecar is involved, so callers
 // with empty sidecar IDs never write events with a blank sidecar_id.
```
hanabel1: This is now false. The pr removes the guard that made that true

internal/cmd/validate_test.go
```diff
@@ -334,15 +381,6 @@ func TestPlanValidationRemoteFlagOverridesLocalConfig(t *testing.T) {
 	assert.Equal(t, plan.PoolSize, 1)
 }
 
-func TestValidateRejectsRemoteAndLocalFlagsTogether(t *testing.T) {
```
hanabel1, line 361: is there a reason this was removed?

hanabel1 (approved):

internal/cmd/validate_test.go
jesseworld22: was not intentional - it got dropped when i resolved the 'validate_test.go' conflict with main.

test is now restored - compared the test function sets against main to confirm nothing else went missing

PR #574, opened 2026-09-09
Drop --identity-file and SSH agent support; standardize on chunk_ai key

### LINKED ISSUE #559: sidecar sync: rsync over SSH-WebSocket proxy fails with connection unexpectedly closed (0 bytes)

schurchleycci:
## Summary

- `rsync` and all other SSH paths now unconditionally use `~/.ssh/chunk_ai`, which chunk auto-generates and registers with the sidecar — removes the fragile `UseAgent`/`IdentitiesOnly=no` branch that had no file path to give rsync's SSH subprocess
- `bridgeConn` now sends WebSocket dial errors to a buffered channel rather than closing silently; a failed proxy connection is surfaced in the returned error instead of appearing as \"0 bytes received so far\"
- Drops `--identity-file` from all commands (`validate`, `validate variants`, `sidecar ssh`, `sidecar setup`, `sidecar sync`)
- Drops SSH agent (`SSH_AUTH_SOCK` / `--auth-sock`) support — chunk never needed it once it started generating its own key
- Removes `useSSHIdentityFile` config key (was stored but never acted upon)
- Simplifies `OpenSession`, `Sync`, `BundleSync`, pool assembly — all identity/agent parameters removed
- Documents automatic SSH key management in `docs/GETTING_STARTED.md` and removes `--identity-file` from `docs/CLI.md`

Fixes #559.

## Test plan

- [x] `task test` passes (all 1550 tests green)
- [x] `task lint` passes (0 issues)
- [ ] `chunk sidecar sync` works end-to-end with the `chunk_ai` key
- [ ] Reproducer from #559 (rsync exit 255 / "0 bytes received") is resolved
- [ ] `chunk sidecar ssh --identity-file` correctly errors as unknown flag

### DIFF
--- a/acceptance/sidecar_test.go
+++ b/acceptance/sidecar_test.go
@@ -376,35 +376,23 @@ func TestSidecarsSyncCheckoutFlagRemoved(t *testing.T) {
 		"--checkout must be an unknown flag; got: %s", combined)
 }
 
-// TestSidecarsSshSyncFlags verifies that SSH/sync flags are accepted and
+// TestSidecarsSshFlags verifies that SSH --identity-file is accepted and
 // code progresses past flag parsing (fails at SSH step, not at parsing).
-func TestSidecarsSshSyncFlags(t *testing.T) {
-	tests := []struct {
-		name string
-		args []string
-	}{
-		{"ssh identity-file", []string{"sidecar", "ssh", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}},
-		{"sync identity-file", []string{"sidecar", "sync", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}},
-	}
+func TestSidecarsSshFlags(t *testing.T) {
+	cci := fakes.NewFakeCircleCI()
+	srv := httptest.NewServer(cci)
+	defer srv.Close()
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			cci := fakes.NewFakeCircleCI()
-			srv := httptest.NewServer(cci)
-			defer srv.Close()
+	env := testenv.NewTestEnv(t)
+	env.CircleCIURL = srv.URL
 
-			env := testenv.NewTestEnv(t)
-			env.CircleCIURL = srv.URL
+	result := binary.RunCLI(t, []string{"sidecar", "ssh", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}, env, env.HomeDir)
 
-			result := binary.RunCLI(t, tt.args, env, env.HomeDir)
-
-			// Commands should fail at SSH key step, not at flag parsing
-			assert.Assert(t, result.ExitCode != 0, "expected non-zero exit (SSH fails)")
-			combined := result.Stdout + result.Stderr
-			assert.Assert(t, strings.Contains(combined, "SSH key not found"),
-				"expected SSH key error (proves flags accepted), got: %s", combined)
-		})
-	}
+	// Should fail at SSH key step, not at flag parsing
+	assert.Assert(t, result.ExitCode != 0, "expected non-zero exit (SSH fails)")
+	combined := result.Stdout + result.Stderr
+	assert.Assert(t, strings.Contains(combined, "SSH key not found"),
+		"expected SSH key error (proves flags accepted), got: %s", combined)
 }
 
 func TestSidecarsExecWithArgs(t *testing.T) {
--- a/acceptance/sidecar_test.go
+++ b/acceptance/sidecar_test.go
@@ -376,9 +376,9 @@ func TestSidecarsSyncCheckoutFlagRemoved(t *testing.T) {
 		"--checkout must be an unknown flag; got: %s", combined)
 }
 
-// TestSidecarsSshFlags verifies that SSH --identity-file is accepted and
-// code progresses past flag parsing (fails at SSH step, not at parsing).
-func TestSidecarsSshFlags(t *testing.T) {
+// TestSidecarsSshNoKey verifies that `sidecar ssh` fails with a clear error
+// when no default key exists at ~/.ssh/chunk_ai.
+func TestSidecarsSshNoKey(t *testing.T) {
 	cci := fakes.NewFakeCircleCI()
 	srv := httptest.NewServer(cci)
 	defer srv.Close()
@@ -386,13 +386,13 @@ func TestSidecarsSshFlags(t *testing.T) {
 	env := testenv.NewTestEnv(t)
 	env.CircleCIURL = srv.URL
 
-	result := binary.RunCLI(t, []string{"sidecar", "ssh", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}, env, env.HomeDir)
+	// No key at env.HomeDir/.ssh/chunk_ai — the default path.
+	result := binary.RunCLI(t, []string{"sidecar", "ssh", "--sidecar-id", "sb-111"}, env, env.HomeDir)
 
-	// Should fail at SSH key step, not at flag parsing
-	assert.Assert(t, result.ExitCode != 0, "expected non-zero exit (SSH fails)")
+	assert.Assert(t, result.ExitCode != 0, "expected non-zero exit (SSH key missing)")
 	combined := result.Stdout + result.Stderr
 	assert.Assert(t, strings.Contains(combined, "SSH key not found"),
-		"expected SSH key error (proves flags accepted), got: %s", combined)
+		"expected SSH key not found error, got: %s", combined)
 }
 
 func TestSidecarsExecWithArgs(t *testing.T) {
--- a/internal/sidecar/rsync.go
+++ b/internal/sidecar/rsync.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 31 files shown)


internal/sidecar/rsync.go
hanabel1, line 1: rsyncTo now auto-generates `~/.ssh/chunk_ai `if missing before calling OpenSession. The bundle sync path (`syncPreparedSidecar -> OpenSession`) does not - it returns a `KeyNotFoundError` with a manual `ssh-keygen` suggestion. 
The new `GETTING_STARTED.md` section says you do not need to create keys - but a fresh install that skips `sidecar setup` and runs `chunk validate `goes through bundle sync and hits an error that contradicts that promise. Is skipping setup an unsupported path? If the auto-generate behavior should be universal, moving it into `OpenSession` would make the guarantee hold uniformly.

internal/sidecar/session.go
```diff
@@ -146,39 +126,26 @@ func addSSHKey(ctx context.Context, client *circleci.Client, sidecarID, pubKey s
...
-			}, nil
-		}
-	}
-
-	if identityFile == "" {
-		identityFile = filepath.Join(sshDir, defaultKeyName)
-	}
+	identityFile := filepath.Join(sshDir, defaultKeyName)
 
 	if _, err := os.Stat(identityFile); err != nil {
```
michael-webster: **🔴 Blocker — Race condition in first-time SSH key generation**

The `os.Stat` → `GenerateKeyPair` sequence is not atomic. `bundleSyncFanOutSince` (`sync.go`) fans out `syncPreparedSidecar` (→ `OpenSession`) concurrently per sidecar, and `validate variants` fans out one goroutine per variant via `RsyncSyncEphemeral` → `OpenSession`. On a machine with no `~/.ssh/chunk_ai` yet, every goroutine passes the `os.Stat` gate simultaneously, each calls `GenerateKeyPair`, and each writes different ed25519 material to the same path. Last write wins — but each goroutine already called `addSSHKey` with *its own* (now-overwritten) public key. Those sidecars reject the private key that survived on disk: SSH auth failure that disappears on the second run.

**Triggering condition:** `assemblePool` entering the `newCount == 0` branch — saved pool state with all existing sidecars alive, but no key on the current machine (new machine, or key deleted). The `sidecar setup` path is safe because `sidecarSetupEnsureSSHKey` runs serially first. The `validate variants` path is not.

**Suggested fix:** add a new `sidecar.EnsureDefaultKeyPair()` helper called once before `sidecar.NewPool` in `setupValidatePool` and before `variants.Run` in `newValidateVariantsCmd`; or protect the stat→generate block here with a package-level `sync.Mutex`.

internal/sidecar/rsync.go
```diff
@@ -113,10 +113,22 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
 		detail := rsyncErrDetail(stderr.String())
-		if detail != "" {
+		var proxyDetail string
+		select {
+		case pe := <-proxyErr:
```
michael-webster, line 129: **🟡 Strong suggestion — `proxyErr` read may miss the error it was added to surface**

This non-blocking read fires immediately after `cmd.Run()` returns, before `stopProxy()`. `bridgeConn` sends to `errCh` only after its WebSocket dial resolves. For the primary failure scenario (dial failure → `bridgeConn` closes TCP → rsync sees close → rsync exits) the causal ordering is safe. But if rsync exits for an independent reason (its own timeout, OS signal) while the WebSocket dial is still in-flight, `default` fires and the proxy error is silently dropped — the returned error is `"rsync: exit status 255"` with no context, exactly the unhelpful message this PR is fixing.

**Suggested fix:** drain after stopping the proxy so the bridge goroutine has a chance to finish:

```go
stopProxy()
var proxyDetail string
select {
case pe := <-proxyErr:
    proxyDetail = pe.Error()
case <-time.After(200 * time.Millisecond):
}
```

acceptance/sidecar_test.go
```diff
@@ -376,35 +376,37 @@ func TestSidecarsSyncCheckoutFlagRemoved(t *testing.T) {
...
-		args []string
-	}{
-		{"ssh identity-file", []string{"sidecar", "ssh", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}},
-		{"sync identity-file", []string{"sidecar", "sync", "--sidecar-id", "sb-111", "--identity-file", "/tmp/fake-key"}},
-	}
+// TestSidecarsSshNoKey verifies that `sidecar ssh` generates the default
+// keypair at ~/.ssh/chunk_ai instead of asking the user to run ssh-keygen.
+// Reaching a sidecar has to work without `sidecar setup` having run first.
+func TestSidecarsSshNoKey(t *testing.T) {
+	sshSrv := fakes.NewSSHServerAcceptingAnyKey(t)
```
michael-webster, line 390: **🔵 Nit — `TestSidecarsSshNoKey` cannot catch a key mismatch between registration and auth**

`NewSSHServerAcceptingAnyKey` accepts any private key, so this test cannot detect a regression where `OpenSession` generates a keypair but registers a *different* public key with the sidecar (e.g. via the race in the blocker above). The test verifies key files exist on disk and exit code is 0, but not that the registered key and the private key used for the SSH handshake are the same pair. A mismatched-key bug passes silently.

Could be tightened: after the binary exits, read the generated public key back from disk, verify it matches what `cci` recorded for the `add-key` request body, and confirm a second invocation with a `NewSSHServer(t, pubKey)` also succeeds.

internal/sidecar/session.go
schurchleycci: Fixed in c76d1fa. Added `sidecar.EnsureKeyPair(path) (generated bool, err error)` which does the check-then-generate under a package-level mutex, and `OpenSession` now calls it instead of doing the stat→generate itself. Putting the lock inside the helper rather than adding an `EnsureDefaultKeyPair()` call before `NewPool`/`variants.Run` means the guarantee holds for any path that reaches a sidecar, so a future fan-out site cannot forget it. `sidecarSetupEnsureSSHKey` now goes through the same helper too (it uses the returned `generated` bool for its status line), so there is one code path.

`TestEnsureKeyPairConcurrent` in `internal/sidecar/ssh_test.go` fans out 8 goroutines at a missing key and asserts exactly one generates and all observe the same public key. It fails with `8 (generated int) != 1` if the mutex is removed.

internal/sidecar/rsync.go
schurchleycci: Fixed in c76d1fa, as suggested — `stopProxy()` now runs before the read, and the `default` arm became a 200ms timeout (`proxyErrDrainTimeout`). `stopProxy` is idempotent (`cancel()` plus an ignored second `ln.Close()`), so the `defer` is harmless.

acceptance/sidecar_test.go
schurchleycci: Tightened in c76d1fa along the lines you described. After the first run the test now unmarshals the recorded `add-key` body and asserts `public_key` equals the on-disk `chunk_ai.pub`, then reruns `sidecar ssh` against a server authorising only that key.

Rather than standing up a second `NewSSHServer` (which fails on a host key mismatch, since the pinned `chunk_ai_known_hosts` entry survives in the test home), I added `SSHServer.RestrictToKey(pubKey)` so the second run reuses the same host key and address and only the auth policy narrows. Verified it has teeth by sabotaging `OpenSession` to register a freshly generated key: the body assertion fails with a key mismatch.

internal/sidecar/rsync.go
schurchleycci: Good catch — you were right that the guarantee was not uniform, and the fix was the one you suggested: generation moved into `OpenSession` (f4bf8a4), so it is no longer special to `rsyncTo`. Every path that reaches a sidecar goes through `OpenSession` — `rsync.go`, `sidecar.go`, `sync.go` (including `syncPreparedSidecar`, the bundle sync path you traced), `target.go`, and `sidecar ssh` — so skipping `sidecar setup` and running `chunk validate` now generates the key rather than erroring. The `GETTING_STARTED.md` promise holds.

Two follow-ups since your comment:

- Michael spotted that the check-then-generate was not atomic, and the fan-out paths (`bundleSyncFanOutSince` per sidecar, `validate variants` per variant) could each generate a different key over the same path. c76d1fa moves it into `sidecar.EnsureKeyPair`, which holds a package mutex, so concurrent callers share one key. `sidecarSetupEnsureSSHKey` goes through the same helper now, so there is a single code path.
- `PublicKeyNotFoundError` still exists, but only for the narrow case where the private key is present and `chunk_ai.pub` is not — a half-deleted pair, which chunk cannot safely regenerate over. Its suggestion is now "delete `~/.ssh/chunk_ai*` and chunk will regenerate the pair on the next run" rather than a manual `ssh-keygen` invocation, matching the docs.

So: skipping setup is a supported path.

internal/sidecar/session.go
```diff
@@ -56,63 +54,187 @@ func GenerateKeyPair(path string) error {
...
+// goroutines in this process via keyGenMu, and other chunk processes sharing the
+// same HOME via an O_EXCL lock file.
+func EnsureKeyPair(path string) (generated bool, err error) {
+	keyGenMu.Lock()
+	defer keyGenMu.Unlock()
+
+	exists, err := keyExists(path)
+	if err != nil || exists {
+		return false, err
 	}
```
danmux, line 130: keyGenMu is held for the entire acquireKeyGenLock duration, which can spin-poll for up to 5s on a stale lock. With a pool of N sidecars all calling OpenSession concurrently, all N queue behind one goroutine for up to 5s. Should the mutex be released before entering the poll loop and re-acquired only for GenerateKeyPair?

internal/sidecar/rsync.go
```diff
@@ -169,32 +185,27 @@ func sshCommand(sess *Session, port string) string {
...
-	return strings.Join(args, " ")
+		// IdentitiesOnly=yes keeps ssh from falling through to the default
+		// ~/.ssh/id_* keys, which the sidecar has never been told about.
+		"-o", "IdentitiesOnly=yes",
+		// Pass the path directly — rsync tokenizes -e by whitespace and calls
+		// execve, so shell quoting (ShellEscape) would embed literal quote
+		// characters in the filename and cause ssh to reject it.
+		"-i", sess.IdentityFile,
+	}, " ")
 }
```
danmux, line 196: Previously guarded behind if sess.IdentityFile != "". The unconditional form now passes "-i " (empty string) if a Session is constructed outside OpenSession. Session is exported — was the guard intentionally removed, or should there be a defensive check?

danmux (approved):

PR #575, opened 2026-09-09
Fix rsync sidecar sync when running from a git worktree

schurchleycci:
## Summary

- **Wrong workspace path**: `filepath.Base(cwd)` returned the worktree slug (e.g. `worktree-green-meadow-822b`) instead of the repo name. In a worktree, the saved workspace is bypassed entirely and the path is derived from the git remote origin URL, matching what every other code path does.
- **Broken `.git` on sidecar**: In a worktree, `.git` is a plain file containing a `gitdir:` pointer to an absolute local path that doesn't exist on the sidecar. rsync was copying it verbatim, causing every git command in the workspace to fail. Fix: exclude `.git` from rsync, then call `initWorktreeGitRepo` which removes any stale `.git` entry, runs `git init`, and sets the origin URL via `git config` so git remote detection works in the workspace.

## Test plan

- [ ] All 1505 existing tests pass (verified on sidecar)
- [ ] `chunk validate` from a git worktree syncs to `/home/user/<repo-name>` (not the worktree directory name)
- [ ] Tests that call `newExecFn` (e.g. `TestOpenAPIExecPassesEnvVars`) pass on the sidecar — git remote detection works in the synced workspace
- [ ] Subsequent syncs are idempotent: `rm -rf .git` + `git init` + `git config remote.origin.url` run cleanly each time

### DIFF
--- a/internal/cmd/validate.go
+++ b/internal/cmd/validate.go
@@ -29,6 +29,7 @@ import (
 	"github.com/CircleCI-Public/chunk-cli/internal/sidecar"
 	"github.com/CircleCI-Public/chunk-cli/internal/ui"
 	"github.com/CircleCI-Public/chunk-cli/internal/validate"
+	"github.com/CircleCI-Public/chunk-cli/internal/watchd"
 )
 
 func newStatusFunc(streams iostream.Streams) iostream.StatusFunc {
@@ -639,7 +640,7 @@ func runValidate(ctx context.Context, client *circleci.Client, rc config.Resolve
 			streams.ErrPrintf("%s\n", ui.Success(fmt.Sprintf("Saved %s to .chunk/config.json", cmdName)))
 		}
 		if sidecarID != "" && allRemote {
-			execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, rec.SetCommandID, streams)
+			execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, workDir, envVars, rc, rec.SetCommandID, streams)
 			if err != nil {
 				return validate.Result{}, err
 			}
@@ -650,7 +651,7 @@ func runValidate(ctx context.Context, client *circleci.Client, rc config.Resolve
 
 	// All-remote execution (--remote flag): send everything to the sidecar.
 	if sidecarID != "" && allRemote {
-		execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, rec.SetCommandID, streams)
+		execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, workDir, envVars, rc, rec.SetCommandID, streams)
 		if err != nil {
 			return validate.Result{}, err
 		}
@@ -663,7 +664,7 @@ func runValidate(ctx context.Context, client *circleci.Client, rc config.Resolve
 		if name != "" {
 			if cmd := cfg.FindCommand(name); cmd != nil && cmd.Remote {
 				rec.Status(iostream.LevelInfo, fmt.Sprintf("running %s on sidecar %s", name, sidecarID))
-				execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, envVars, rc, rec.SetCommandID, streams)
+				execFn, dest, err := newExecFn(ctx, client, sidecarID, workdir, workDir, envVars, rc, rec.SetCommandID, streams)
 				if err != nil {
 					return validate.Result{}, err
 				}
@@ -820,11 +821,14 @@ func reapAbandonedSidecars(ctx context.Context, client *circleci.Client, workDir
 // stdout and stderr are therefore always empty — callers print output only when
 // it was not already streamed, so there is nothing left for them to do.
 //
+// projectRoot is the local repository root, used only to attribute registered
+// commands to a project for the watch daemon.
+//
 // Pass rec.SetCommandID for setCommandID when each exec's command ID should be
 // recorded on the event log. Pass nil for probe-only use (e.g. WorkspaceExists)
 // where the exec result should not touch the recorder.
 func newExecFn(
-	ctx context.Context, client *circleci.Client, sidecarID, workdir string,
+	ctx context.Context, client *circleci.Client, sidecarID, workdir, projectRoot string,
 	envVars map[string]string, rc config.ResolvedConfig, setCommandID func(string), streams iostream.Streams,
 ) (func(context.Context, string) (string, string, int, error), string, error) {
 	cwd, _ := os.Getwd()
@@ -850,7 +854,24 @@ func newExecFn(
 		_, _ = w.Write(data)
 	}
 	execFn := func(ctx context.Context, script string) (string, string, int, error) {
-		result, err := client.Exec(ctx, sidecarID, "sh", []string{"-c", script}, merged, onOutput)
+		// A nil setCommandID marks a probe: a command chunk issues on its own
+		// behalf, with no event log entry to attribute and nothing a developer
+		// would go looking for in the dashboard. Both omissions follow from that
+		// one fact, so they read off the same flag.
+		var reg *watchd.CommandReg
+		if setCommandID != nil {
+			// Registering before output is consumed is what lets the watch daemon
+			// tail this command while it runs, and keep its output after this
+			// process exits — which for a hook-driven run is immediately.
+			reg = &watchd.CommandReg{
+				SidecarID:   sidecarID,
+				ProjectRoot: projectRoot,
+				Op:          string(eventlog.OpValidate),
+				Name:        remoteCommandLabel(script),
+			}
+		}
+		result, err := submitAndStream(ctx, client, sidecarID, reg,
+			"sh", []string{"-c", script}, merged, onOutput)
 		if err != nil {
 			if setCommandID != nil {
 				// result is nil on all error paths; clear any stale pending ID.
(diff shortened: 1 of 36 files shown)


internal/sidecar/rsync.go
```diff
@@ -51,12 +55,40 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
...
 	if persist {
-		repo := filepath.Base(cwd)
-		repoPath, err = ResolveWorkspace(ctx, workdir, repo)
-		if err != nil {
-			return fmt.Errorf("rsync: resolve workspace: %w", err)
+		if isWorktree && worktreeOriginURL != "" && workdir == "" {
+			// Skip the saved workspace in a worktree: it may have been recorded as
+			// the worktree directory name by older code. Derive from the repo name
+			// so the path matches what every other code path expects.
+			_, repo, _ := gitremote.ParseRemoteURL(worktreeOriginURL)
```
jesseworld22, line 81: ParseRemoteURL matches only github\.com[:/]org/repo (internal/gitremote/detect.go:10). For any non-github.com origin — GitHub Enterprise (git@ghe.company.com:org/repo.git), GitLab, a plain local path remote — it returns ("", "", err). The error is discarded, so repo == "" and DefaultWorkspace("") yields "/home/user/" (internal/sidecar/sync.go:37-39).

Downstream that value is used for:
- dst = user@127.0.0.1:/home/user/ with --archive --delete — every file in the sidecar's home not present in the local tree is deleted (~/.ssh, caches, toolchains).
- initWorktreeGitRepo → rm -rf '/home/user//.git' and git init in $HOME.
- persistWorkspace(ctx, "/home/user/"), which records the bad path for all future syncs.

This is precisely the case ResolveWorkspace guards against, and the new branch routes around that guard:

// sync.go:43-44
// Returns an error if no repo-specific path can be determined (repo empty and no
// saved workspace), because the bare home dir is not safe to pass to rm -rf.

Suggested fix: check the error and fall through to ResolveWorkspace (or return it) rather than discarding it, so the empty-repo guard still applies.

internal/sidecar/rsync.go
```diff
@@ -51,12 +55,40 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
...
+
+	// In a worktree, fetch the origin URL once; it is used for workspace
+	// resolution (to parse the repo name) and to configure the remote on the
+	// sidecar after init.
+	var worktreeOriginURL string
+	if isWorktree {
+		if out, execErr := exec.CommandContext(ctx, "git", "-C", cwd, "remote", "get-url", "origin").Output(); execErr == nil {
+			worktreeOriginURL = strings.TrimSpace(string(out))
+		}
+	}
```
jesseworld22, line 73: if out, execErr := exec.CommandContext(ctx, "git", "-C", cwd, "remote", "get-url", "origin").Output(); execErr == nil {
    worktreeOriginURL = strings.TrimSpace(string(out))
}

When this command fails, worktreeOriginURL stays empty and nothing is reported. Two things then happen silently:

1. The else branch runs, DetectOrgAndRepo also fails, repo falls back to filepath.Base(cwd) — the worktree slug — and ResolveWorkspace returns the stale saved workspace. That is the exact bug this PR sets out to fix, re-entered without a diagnostic.
2. initWorktreeGitRepo still runs but returns early at if originURL == "", so the sidecar gets a git repo with no origin. The PR's stated goal ("git remote detection works in the synced workspace", newExecFn/TestOpenAPIExecPassesEnvVars) fails, and the user sees only downstream test failures.

This is a realistic path, not a hypothetical: a worktree whose main repo directory has been moved or deleted has a dangling gitdir: pointer, so every git command in it fails while .git still stats as a file — isWorktree is true and the origin lookup errors.

Should this either surface a status(iostream.LevelWarn, ...) or return a wrapped error? Also, internal/gitremote already owns origin-remote lookup; adding an OriginURL(dir) helper there would avoid a second git remote get-url origin invocation living in internal/sidecar.

internal/sidecar/rsync.go
```diff
@@ -119,10 +158,46 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
...
+		return fmt.Errorf("remove stale .git on sidecar: exit %d: %s", result.ExitCode, result.Stderr)
+	}
+	if result, err := ExecOverSSH(ctx, sess, "git -C "+ShellEscape(repoPath)+" init -q", nil, nil); err != nil {
+		return fmt.Errorf("git init on sidecar: %w", err)
+	} else if result.ExitCode != 0 {
+		return fmt.Errorf("git init on sidecar: exit %d: %s", result.ExitCode, result.Stderr)
+	}
+	if originURL == "" {
+		return nil
+	}
```
jesseworld22, line 190: --exclude=.git matches at any depth, not just the repo root — nested repos, submodules and git-repo test fixtures get dropped from the sidecar too. Was --exclude=/.git intended?

Also, rm -rf .git + git init runs on every sync, not just the first, so any sidecar-side git state is rebuilt each time. ensureRemoteRepo (sync.go:293) already uses a test -d .git conditional for this — worth matching, with only git config remote.origin.url re-run each sync?

internal/sidecar/rsync.go
```diff
@@ -51,12 +55,40 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
 		return fmt.Errorf("rsync: open session: %w", err)
 	}
 
+	// Detect a git worktree early: in a worktree .git is a file (not a dir)
+	// containing a gitdir pointer to an absolute local path. This affects both
+	// workspace resolution (use repo name, not directory name) and how the
+	// remote git repo is set up after the sync.
+	gitInfo, statErr := os.Stat(filepath.Join(cwd, ".git"))
+	isWorktree := statErr == nil && !gitInfo.IsDir()
```
jesseworld22: All three callers pass cwd straight from os.Getwd() (internal/cmd/validate.go:741, internal/cmd/sidecar.go:573, and sidecarSetupSync), with no repo-root resolution. Running chunk validate from, say, <worktree>/internal/cmd means .git does not exist at cwd, isWorktree is false, and the code takes the else branch — DetectOrgAndRepo succeeds (git searches upward) but ResolveWorkspace then returns the stale saved workspace, i.e. the worktree-slug path the PR is trying to escape. The same invocation from the worktree root goes to a different destination.

Would git rev-parse --git-dir compared against --git-common-dir (they differ in a worktree, match in a normal checkout) be a more robust detection that works from any directory in the tree?

jesseworld22 (commented):
additional claude findings:

internal/sidecar/rsync.go:62 — reuse

▎ findGitRoot (internal/sidecar/active.go:265) already does this upward walk. Hand-rolling os.Stat(filepath.Join(cwd, ".git")) is why detection fails when run from a subdirectory of the worktree — .git isn't at cwd, so isWorktree is false and we fall back to the stale saved workspace. Could findGitRoot take a dir param and be reused here?

---

internal/sidecar/rsync.go:87 — scope

▎ This else branch changes non-worktree behavior too: repo was unconditionally filepath.Base(cwd), now it's DetectOrgAndRepo(cwd) with the basename only as fallback. For a checkout cloned into a differently-named directory (git clone …/chunk-cli.git my-work), the destination moves my-work → chunk-cli. Existing users are shielded by the saved sidecar.json workspace, but a fresh sidecar lands somewhere new. Intentional, or should the normal-checkout path stay as it was?

---

internal/sidecar/rsync.go — tests

▎ ~60 lines of new branching with no test coverage; rsync_test.go only covers sshCommand and rsyncErrDetail. Pulling workspace resolution into a pure helper (func worktreeWorkspace(cwd string) (string, bool)) would make it testable without a live session — and would have caught the empty-repo case above.

internal/cmd/validate.go
```diff
@@ -832,7 +832,10 @@ func newExecFn(
 	envVars map[string]string, rc config.ResolvedConfig, setCommandID func(string), streams iostream.Streams,
 ) (func(context.Context, string) (string, string, int, error), string, error) {
 	cwd, _ := os.Getwd()
-	_, repo, _ := gitremote.DetectOrgAndRepo(cwd)
+	_, repo, repoErr := gitremote.DetectOrgAndRepo(cwd)
+	if repoErr != nil {
+		repo = filepath.Base(cwd)
+	}
```
hanabel1: When `DetectOrgAndRepo` fails, repo falls back to` filepath.Base(cwd)` - the worktree slug. That value gets passed into `RsyncSync` as a non-empty `workdir`, which makes the new `isWorktree && workdir == ""` guard in `rsyncTo` false. The fix  is bypassed and the sidecar workspace ends up named after the slug anyway. 

This only bites non-GH remotes, but its a silent failure - the user gets a wrong path with no warning.

internal/sidecar/rsync.go
```diff
@@ -51,12 +55,54 @@ func rsyncTo(ctx context.Context, client *circleci.Client,
...
+		resolved := false
+		if isWorktree && workdir == "" {
+			// Skip the saved workspace in a worktree: it may have been recorded as
+			// the worktree directory name by older code. Derive from the repo name
+			// so the path matches what every other code path expects.
+			if ws, ok := worktreeWorkspace(worktreeOriginURL); ok {
+				repoPath = ws
+				resolved = true
+			}
+			// If worktreeWorkspace returns false (non-GitHub remote, empty URL, or
```
hanabel1: if `worktreeWorkspace` returns false  and `DetectOrgAndRepo` also fails, repo again becomes `filepathBase(cwd)`. The origin URL was already fetched via `git remote get-url origin` and is sitting in `worktreeOriginURL` - could the repo name be parsed from that directly instead of falling back to the directory name?

hanabel1 (approved):
the comments only affect non-GH remotes in a worktree but I think the core fix works correctly. worth following up/fixing!

PR #576, opened 2026-09-09
Detect merge conflicts in the daemon

jesseworld22:
## Summary

- Daemon previews each project's branch merging into the default branch on `origin` or `upstream` - every 60s, refetching the target ref every 3 min
- Uses `git merge-tree --write-tree`, so nothing touches the working tree, index, or HEAD. Needs git 2.38+; an older git reports "no answer", never a clean merge
- New `chunk conflicts` reads the stored answer over the daemon socket; `chunk init` adds `chunk conflicts --hook` to the commit hook group (Claude Code settings only — `.codex/hooks.json` is unchanged)
- Advisory by construction: always exits 0 and reports via `hookSpecificOutput.additionalContext`, so it can't block a commit like the `chunk validate` entries beside it
- Silent when there's nothing to advise; "no answer" (detached HEAD, unfetched target, no daemon, a project the daemon has never been pointed at) stays distinct from "no conflict"
- Committed history only - uncommitted work is invisible to the preview, and every notice says so

Committed using squash commit.

## Test plan

- [x] `task test` - 1826 tests, 2 skipped, `-race`, green; `task lint` clean
- [x] New unit coverage: `mergetree_test.go`, `watchd/conflicts_test.go`, `watchd/notice_test.go`, `cmd/conflicts_test.go`
- [x] `/conflicts` endpoint tested against a real daemon over a real socket
- [x] Verified by hand end-to-end: real daemon, real remote, real conflict - fetch, preview, and both output modes
- [ ] Live check that the agent receives `additionalContext` on `PreToolUse` - not yet verified; only the hook JSON shape is covered, by `cmd/conflicts_test.go`

## Regular Command:
Screenshot 2026-09-16 at 11 48 09 AM

## Hook:

Screenshot 2026-09-16 at 11 49 09 AM

## JSON:

Screenshot 2026-09-16 at 11 51 25 AM


internal/cmd/conflicts_test.go
```diff
@@ -0,0 +1,80 @@
...
+	out, _, runErr := runConflictsCmd(t, dir, "--hook")
+	assert.NilError(t, runErr)
+	if out == "" {
+		return // nothing to advise, which is the documented quiet path
+	}
+	var parsed hookOutput
+	decodeErr := json.Unmarshal([]byte(out), &parsed)
+	assert.NilError(t, decodeErr)
+	assert.Check(t, cmp.Equal(parsed.HookSpecificOutput.HookEventName, "PreToolUse"))
+}
```
hanabel1, line 209: This test never actually runs its assertions. the test that checks that hook mode-stdout is always valid via json always exits early via `if out == ""`. bc the test points chunk_watch_dir at an empty directory, the daemon is always unreachable and the hook mode has no output, and the` json.Unmarshal` block is dead code. If someone accidentally writes plain text to stdout in hook mode, the test won't catch it.

internal/watchd/conflicts.go
```diff
@@ -0,0 +1,228 @@
...
+	if targetErr != nil {
+		state.Unavailable = fmt.Sprintf("%s has never been fetched", target)
+		return state, fetchedAt
+	}
+	if headErr != nil {
+		state.Unavailable = "this branch has no commits yet"
+		return state, fetchedAt
+	}
+	state.HeadSHA = headSHA
+	state.TargetSHA = targetSHA
```
hanabel1: targetSHA and headSHA share a 10-second deadline. if the first call uses most of that time, the secon times out and stores "this branch has no commits yet" which isn't what happened. (it timed out)

internal/cmd/conflicts.go
```diff
@@ -0,0 +1,126 @@
...
+	enc := json.NewEncoder(streams.Out)
+	_ = enc.Encode(out)
+	return nil
+}
+
+// reportConflictLookupError explains a missing answer, but only where an
+// explanation is wanted. A hook stays silent: the daemon is optional, and a
+// hook that announced its absence would fire on every commit for everyone not
+// running `chunk watch`.
+func reportConflictLookupError(streams iostream.Streams, err error, hookMode, jsonOut bool) {
```
hanabel1: success gives you a Conflict Report object. An error gives you {"error": "..."}. Any script that expects .known or .conflict to always be present will break on the error case. The fix is to always return a ConflictReport. On failure, just leave know: false and put the error text in ConflictState.Unavailable.

hanabel1:
Screenshot 2026-09-15 at 11 50 20 AM
this looks good tho

internal/gitutil/mergetree.go
```diff
@@ -0,0 +1,192 @@
...
+// the background. The cost of a full fetch is somebody else's bandwidth and a
+// pack the developer did not ask for.
+func FetchRemoteBranch(ctx context.Context, dir, remote, branch string) error {
+	refspec := fmt.Sprintf("+refs/heads/%s:refs/remotes/%s/%s", branch, remote, branch)
+	cmd := exec.CommandContext(ctx, "git", "-C", dir,
+		"fetch", "--quiet", "--no-tags", "--no-write-fetch-head", remote, refspec)
+	// A fetch must never turn into a credential prompt: this process has no
+	// terminal to prompt on, and a git that blocks waiting for input would hang
+	// the timer it runs on until the context expires.
+	cmd.Env = append(cmd.Environ(), "GIT_TERMINAL_PROMPT=0", "GIT_ASKPASS=", "SSH_ASKPASS=")
```
hanabel1: Appending puts the overrides at the end. getenv() returns the first match, so if the parent process already has GIT_TERMINAL_PROMPT=1 or SSH_ASKPASS set, git ignores the appended values and may block waiting for credentials — stalling the fetch goroutine for up to 30s.

Fix: filter those keys out of os.Environ() before appending them.

internal/watchd/client.go
```diff
@@ -79,6 +79,48 @@ func RegisterCommand(reg CommandReg) {
...
+// session end for everyone who does not.
+var ErrDaemonUnreachable = errors.New("no watch daemon is running")
+
+// FetchConflicts asks the running daemon whether root's branch still merges
+// cleanly into its merge target.
+//
+// Unlike RegisterCommand this reports its errors, because the caller decides
+// how loudly to fail: a hook stays quiet, a person running the command by hand
+// gets told why there is no answer.
+func FetchConflicts(root string) (ConflictReport, error) {
```
hanabel1, line 110: Every connection error - including a 2s timeout while the daemon is busy - returns ErrDaemonUnreachable. The user sees "No watch daemon is running. Run chunk watch." even when the daemon is up.

internal/watchd/daemon.go
```diff
@@ -206,7 +219,27 @@ func (d *daemon) initProject(root string) *projectState {
...
+// which is the difference between "no answer" and "no conflicts".
+func (d *daemon) conflictReport(root string) ConflictReport {
+	want := canonicalRoot(root)
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	for _, ps := range d.projects {
+		if ps.root != root && ps.canonRoot != want {
+			continue
+		}
+		return ConflictReport{Root: ps.root, Conflict: ps.conflict, Known: true}
```
hanabel1, line 240: ps.conflict is nil for the first 5s after the daemon starts. conflictReport returns known: true, conflict: nil during that window - any consumer that checks known and then dereferences Conflict without a nil guard will panic. No test covers this state. 

Fix: add a test serving Conflict Report {Known:true, Conflict: nil} and assert it exits cleanly with no output.

internal/watchd/client.go
```diff
@@ -79,6 +79,63 @@ func RegisterCommand(reg CommandReg) {
...
+// working daemon to go start one.
+var ErrDaemonTimeout = errors.New("the watch daemon did not answer in time")
+
+// FetchConflicts asks the running daemon whether root's branch still merges
+// cleanly into its merge target.
+//
+// Unlike RegisterCommand this reports its errors, because the caller decides
+// how loudly to fail: a hook stays quiet, a person running the command by hand
+// gets told why there is no answer.
+func FetchConflicts(root string) (ConflictReport, error) {
```
danmux, line 110: Any non-timeout error from client.Get falls through to ErrDaemonUnreachable, telling the user to start the daemon. A permission-denied on the socket is a distinct case — the daemon is there but unreadable — and this sends them to start something they already have.

docs/CLI.md
```diff
@@ -208,6 +213,27 @@ chunk
   env var → `orgID` in `.chunk/config.json` → interactive org picker (TTY only).
   Non-interactive sessions (agents, CI) should set `orgID` in project config or
   pass `--org-id` / `CIRCLECI_ORG_ID`.
+- **`conflicts` is advisory and always exits 0.** It reports what the watch
+  daemon already computed and never previews a merge itself, so it returns
+  immediately. It is deliberately incapable of failing: it runs from the same
+  commit hook group as `chunk validate`, whose non-zero exit blocks a commit,
+  and a merge conflict with the default branch is information rather than a
+  reason to stop. With no daemon running, `--hook` prints nothing at all —
```
danmux, line 228: The docs say "no answer -> known=false -> reason in conflict.unavailable", but when the daemon is running and does not recognise the root it returns {known: false, conflict: null} — conflict is absent, not present. A script following these docs would nil-deref on conflict.unavailable. Document the two known=false sub-cases separately, or have the daemon-unknown path populate conflict.unavailable?

danmux (approved):

danmux (approved):

PR #578, opened 2026-09-09
Implement memory buffer cap

hanabel1:
### Summary

Commands that produce a lot of output were held in unbounded []byte buffers, which could exhaust memory on sidecars running verbose test suites.

- Raised the default per-command output cap from 256 KB → 10 MB
- Made the cap configurable via CHUNK_OUTPUT_BUFFER_SIZE (bytes) so sandbox-provisioner can tune it at deploy time without rebuilding images
- Each buffer now carries its own maxBytes field; the FIFO eviction logic was already correct, just not wired to a configurable size

### Implementation

- MaxCommandBytes default: 256 << 10 → 10 << 20 (10 MB)
- outputBufferCap() reads CHUNK_OUTPUT_BUFFER_SIZE; invalid/non-positive values fall back to the default
- buffer.maxBytes replaces the package-level constant in append()
- outputStore.bufCap holds the configured cap and passes it to each newBuffer() call in register()
- daemon.go calls newOutputStore(ctx, outputBufferCap()) at startup

### Test plan

- [ ] All existing buffer tests pass (go test -race ./internal/watchd/...)
- [ ] Full test suite passes (task test)
- [ ] Lint passes (task lint)
- [ ] Set CHUNK_OUTPUT_BUFFER_SIZy command truncates and reportsTruncated: true


schurchleycci (approved):
One minor thing: it'd be nice to print out a warning if a user has set the env var to an invalid value. I think the fallback itself is fine, I just mean making it not a silent fallback

PR #580, opened 2026-09-10
Track async validation tasks and discard stale results

jesseworld22:

1. #580 – Track async validation tasks and discard stale results  ← **this PR**

Carries `config.CanonicalProjectRoot` and the `newTestDaemon` test helper verbatim from #573, which this series used to stack on. Nothing else in #573 is used here – `taskStore` is deliberately in-memory – so whichever of the two lands second drops its copy.

---

## Summary

Somewhere to keep a validation run nobody is waiting for, and a rule for throwing out answers that went out of date.

- **The store** – release the caller and something has to hold which project the run was for, whether it's still going, and what it concluded.
- **Staleness** – the tree is fingerprinted at the start and re-checked both when the run ends and when the result is read. If it moved, the verdict is dropped; a tree that can't be fingerprinted at all is refused outright.
- **Discards are announced, not silent** – usually the run moved the tree itself, and saying nothing is indistinguishable from no run happening. Keeping the result instead means pinning the tree, which is #584.
- **Interface** – `POST /validate/async` and `GET /validate/collect`; `chunk validate --async` and `chunk validate results`. Any refusal falls back to running inline.
- **Hook** – `chunk init` adds `chunk validate results` on `UserPromptSubmit`, the next moment an agent can be told. Codex sessions don't get it.
- **Delivery** – results are stamped delivered rather than deleted, and only once the response is written, so a lost response isn't a lost failure.
- **Runs use the project that asked** – `ProjectRoot` reaches the runner as `--project`, and tasks key by repo root. Previously repo B's task could hold repo A's exit code and pass.
- **Nothing triggers a background run yet** – that's #581.

## Test plan

- [x] Every behavioural fix confirmed by reverting it and watching the test fail with the reported symptom, not by the suite passing.
- [x] Staleness, delivery, eviction and per-project keying, the last over real git repos rather than fakes.
- [x] Both endpoints over the real mux and socket, including a failed write leaving results still owed.
- [x] `task test` (`-race`), `task lint`.


schurchleycci:
Couple things I'm (not Claude 😅) wondering about:
- do we want to discard the results or mark them as delivered?
- the name of the `--collect` flag feels a bit opaque to me. Not sure what would be a better name for it though 🤔  

+ a few more Claude findings:

High

1. internal/watchd/validate.go:105 — async run executes in the daemon's cwd, not req.ProjectRoot (unchanged: High)
makeValidateRunner (internal/cmd/watchdaemon.go:65) replays the caller's args verbatim; resolveWorkDir (internal/cmd/validate.go:331) falls back to os.Getwd(); launchDaemon (internal/watchd/client.go:392) never sets child.Dir; the socket is user-global (internal/watchd/paths.go:9), so one daemon serves many repos by design. req.ProjectRoot is used only for keying and fingerprinting, never execution — so repo B's task holds repo A's exit code and the staleness check confirms B is unchanged, delivering a false pass. Note the synchronous /validate path shares the defect (validate.go:639), so the cwd bug is pre-existing; async is what makes it authoritative. Fix both by forwarding --project.

2. internal/watchd/asyncvalidate.go:161 — a run invalidates its own result (raised: Medium → High)
Pre-run fingerprint at :109, compared at :161 and :218. Fingerprint uses git status --porcelain -z -uall, so any non-ignored artifact the run creates flips the digest → stale → dropped silently, indistinguishable from "no runs happened". Raised because it fails closed for a whole class of projects (coverage output, regenerated goldens, touched lockfiles) and, with finding 1, leaves no configuration where the feature demonstrably works. Correction to the original note: the hook cache keys on the pre-run fingerprint, which avoids self-invalidating that result but still misses on the next run for the same reason — not a clean precedent.

Medium

3. internal/cmd/validate.go:723 — tasks keyed by cwd, not repo root (unchanged: Medium)
CanonicalProjectRoot (internal/config/paths.go:64) is Clean + EvalSymlinks only and doesn't walk upward. A run started in /repo/sub keys there; the --collect hook (cwd = project root) asks /repo; runCollect prints nothing. gitutil.RepoRoot already exists. Correction: "fingerprints already normalise that way" is half true — Fingerprint resolves --show-toplevel internally for the digest root but returns no root, so nothing normalizes the key today. Fix still correct.

4. docs/CLI.md:77 / docs/HOOKS.md — --async, --collect, and the collect hook undocumented (raised: Low → Medium)
CLI.md's validate flag list (lines 59–78) omits both flags; HOOKS.md covers only the Stop hook. Raised because chunk init now writes a new UserPromptSubmit entry into the user's .claude/settings.json (internal/settings/settings.go, merge.go) that runs before every prompt, with nothing telling them it exists — a doc gap with a user-visible side effect, not cosmetic.

Low

5. internal/watchd/validate.go:133 — collect deletes tasks before encoding the response (lowered: Medium → Low)
Ordering confirmed: delete under the mutex at asyncvalidate.go:222, encode afterward at validate.go:133; any later failure loses the result, and runCollect maps a client timeout to ErrDaemonUnavailable → silent nil. Lowered because the stated trigger is unlikely: the 5s budget (internal/watchd/ipc.go:94) is generous for a digest capped at 64 MB over changed paths only. The client-disconnect / encode-error variant is unconditional though, so still worth the cheap fix — marshal to a buffer before deleting, or split into peek-then-acknowledge.

6. internal/config/paths.go:92 — ProjectDataDir("") hashes to a shared sha256("") bucket (unchanged: Low)
Sharpening: this is an actual behaviour change, not purely latent. Previously clean was "." and EvalSymlinks(".") succeeded, giving a cwd-specific hash; now it's sha256(""), and the migration branch at :96 fires spuriously ("" != "."). No live impact — all in-tree callers pass a real root. Prefer erroring on "".

Nit

7. internal/watchd/asyncvalidate.go:144 — "Fingerprinted before taking the lock" comment is misplaced (unchanged: trivial)
The claim is true of the sequence (lock, read, unlock, fingerprint at :156) but sits 12 lines from what it describes. The identical rationale at :191 in collect is placed correctly — worth matching.

internal/watchd/asyncvalidate.go
schurchleycci: Could a filename more like `tasks.go` make sense here?

internal/settings/merge.go
```diff
@@ -31,6 +31,23 @@ const legacyCommitMatcher = "Bash(git commit*)"
...
+//
+// It goes on UserPromptSubmit rather than Stop because by the time a background
+// run has an answer, Stop has already fired — the agent has finished and gone.
+// UserPromptSubmit runs before Claude processes the next prompt and its stdout
+// is added to the agent's context, so it is the first moment a deferred result
+// can be delivered. Identified by this exact string during merge, so it must
+// stay in sync with the command written by Build.
+const ResultsCommand = "chunk validate results"
+
+// legacyCollectCommand is the flag spelling ResultsCommand replaced, written by
```
schurchleycci: I don't think we need `legacyCollectCommand` since `--collect` never shipped

internal/watchd/tasks.go
schurchleycci, line 1: Could a filename more like `tasks.go` make sense here?

internal/cmd/validate.go
```diff
@@ -235,6 +237,16 @@ func newValidateCmd() *cobra.Command {
...
 	cmd.Flags().StringVar(&opts.envFile, "env-file", defaultEnvFile, "Env file to load (default: .env.local; pass a path to override)")
+	cmd.Flags().BoolVar(&opts.async, "async", false, "Run in the background via the watch daemon and report on a later run")
+	// --collect is what "validate results" used to be. Hidden rather than
+	// removed: the flag is written into .claude/settings.json by chunk init, so
+	// every project already set up by an earlier version has a hook invoking it,
+	// and dropping it would make those hooks fail until the settings are
+	// regenerated. Merge rewrites the entry to the subcommand (see
+	// settings.legacyCollectCommand); this keeps the old spelling working in the
+	// meantime.
+	cmd.Flags().BoolVar(&opts.collect, "collect", false, "Deprecated: use 'chunk validate results'")
```
schurchleycci: Same here, I don't think we need the `--collect` stuff since it never shipped

internal/settings/merge.go
```diff
@@ -153,13 +170,16 @@ func mergePermissionsAllow(merged, generated map[string]interface{}) {
...
+//   - UserPromptSubmit: entries whose command is ResultsCommand, or the legacy
+//     flag spelling it replaced.
 func mergeHooks(merged, generated map[string]interface{}) {
 	genHooks, ok := generated["hooks"].(map[string]interface{})
 	if !ok {
 		return
 	}
 	mergeHookType(merged, genHooks, "PreToolUse", ownsCommitEntry, isChunkCommitGroup)
 	mergeHookType(merged, genHooks, "Stop", ownsStopEntry, nil)
+	mergeHookType(merged, genHooks, "UserPromptSubmit", ownsResultsEntry, nil)
```
schurchleycci: I think we should defer the hooks install until we actually request a background run (which I don't think this PR does?) and have results to report

internal/watchd/tasks.go
```diff
@@ -0,0 +1,427 @@
...
+}
+
+// collect returns the finished tasks for root that nobody has been told about
+// yet — including ones discarded as stale, with their verdicts stripped — and
+// records them delivered in one step. It is peek followed by acknowledge, for
+// callers with nothing that can fail in between.
+//
+// The daemon's handler does not use it: there, encoding and writing the response
+// sit between the two halves, and both can fail. See handleCollect.
+func (s *taskStore) collect(root string) []TaskState {
```
schurchleycci: Looks like `collect` is only used in tests - I think we can drop it

internal/watchd/tasks.go
```diff
@@ -0,0 +1,427 @@
...
+			// that basis. Dropped in silence that is indistinguishable from no run
+			// having happened, so a project whose commands are not perfectly
+			// gitignored gets nothing, forever, with nothing to say why.
+			reported.Stale = true
+			reported.ExitCode = 0
+			reported.Output = ""
+		}
+		out = append(out, reported)
+		kept = append(kept, id)
+	}
```
schurchleycci: I think these two branches below may be unreachable?

internal/watchd/validate.go
```diff
@@ -35,6 +55,133 @@ type ValidateResponse struct {
...
+// ErrDaemonUnavailable, and the results hook reads that as "no daemon running,
+// nothing to report" and prints nothing at all. A failed validation would
+// disappear into what looks like an ordinary quiet turn.
+//
+// Acknowledging after the write leaves one window this cannot close: a client
+// that dies between the daemon's successful write and printing what it read.
+// Shutting that too needs the client to acknowledge in a second call, which is a
+// round trip on every prompt for a case an unacknowledged result already
+// survives — it is simply reported again next turn.
+func (d *daemon) handleCollect(w http.ResponseWriter, r *http.Request) {
```
michael-webster, line 164: **`handleCollect` accepts any HTTP method**

`handleAsyncValidate` (line 76) rejects non-POST with 405, but `handleCollect` has no `r.Method` guard. The stdlib mux applies no method enforcement, so a `DELETE` or `POST` to `/validate/collect?root=X` reaches the full peek-and-acknowledge cycle and permanently marks pending results as delivered — silently draining them before the hook can read them.

```suggestion
func (d *daemon) handleCollect(w http.ResponseWriter, r *http.Request) {
	if r.Method != http.MethodGet {
		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
		return
	}
	root := r.URL.Query().Get("root")
```

internal/watchd/validate.go
```diff
@@ -35,6 +55,133 @@ type ValidateResponse struct {
...
+		var stdout, stderr bytes.Buffer
+		// root, not the daemon's cwd: the task is filed and fingerprinted against
+		// this project, so validating any other one would report an answer about
+		// the wrong repo — and the staleness check, comparing this project against
+		// itself, would confirm it.
+		exitCode := d.runner(ctx, root, args, env, &stdout, &stderr)
+		// stderr carries the progress lines and the tally; stdout is usually
+		// empty for a validate run. Both are kept so whoever collects the result
+		// sees what the developer would have seen.
+		return exitCode, stdout.String() + stderr.String()
```
michael-webster, line 129: **Unbounded output captured per async task**

`stdout.String() + stderr.String()` is stored in `TaskState.Output` with no size cap. `MaxTasksPerProject=20` tasks per project, each holding potentially megabytes of `go test -v` output, sit in the daemon heap until collected. If the `UserPromptSubmit` hook is not running (or the developer is not using an agent), tasks accumulate uncollected and heap growth is only bounded by eviction — up to 20 × (however verbose the validate command is) × (however many projects are active).

Consider capping before storing, e.g. keeping the last N bytes (the tail is where failures typically appear):
```go
out := stdout.String() + stderr.String()
const maxOutput = 64 * 1024
if len(out) > maxOutput {
    out = out[len(out)-maxOutput:]
}
return exitCode, out
```

internal/watchd/tasks.go
```diff
@@ -0,0 +1,427 @@
...
+	ids := s.byProject[root]
+	for len(ids) > MaxTasksPerProject {
+		victim := s.oldestLocked(ids, func(e *taskEntry) bool {
+			return !e.state.Running && !e.state.DeliveredAt.IsZero()
+		})
+		if victim == -1 {
+			victim = s.oldestLocked(ids, func(e *taskEntry) bool { return !e.state.Running })
+		}
+		if victim == -1 {
+			// Every retained task is still running. Going over the cap is better
```
michael-webster, line 453: **`evictLocked` goes over cap when all tasks are running — realistically reachable**

The `break` here is intentional (do not cancel in-flight work), but there is no compensating guard in `start()`. On a busy agentic session (`/loop`, automated prompts), `UserPromptSubmit` fires every 1–3 seconds; a 30-second validate run means 10–30 new tasks start before the first one finishes. Once `byProject[root]` has 21+ running tasks, every subsequent `start()` grows both maps by one with no eviction and no upper bound.

Consider returning an error from `start()` when the in-flight count already exceeds `MaxTasksPerProject`, so the caller falls back to synchronous execution rather than allowing unbounded growth:
```go
if len(s.inFlight(root)) >= MaxTasksPerProject {
    return "", errors.New("too many runs already in flight")
}
```

internal/watchd/client.go
```diff
@@ -231,6 +237,86 @@ func RunValidate(args []string, circleCIToken string) (ValidateResponse, error)
...
+
+// CollectValidateResults returns the finished async results for projectRoot and
+// clears them from the daemon, so a result is reported once and not repeated.
+//
+// Best-effort: with no daemon running there is nothing to collect and nothing to
+// report, which is not an error worth surfacing on a hook path.
+func CollectValidateResults(projectRoot string) ([]TaskState, error) {
+	sockPath, err := SocketPath()
+	if err != nil {
+		return nil, err
```
michael-webster: **`SocketPath()` failure not wrapped as `ErrDaemonUnavailable`**

The function's own doc says *"Best-effort: with no daemon running there is nothing to collect and nothing to report, which is not an error worth surfacing on a hook path."* — but that intent only covers connection failures (line 303, which is correctly wrapped). A `SocketPath()` failure (missing `$HOME`, broken XDG config, containerised runner) returns bare and surfaces as an error message before every agent prompt.

`RunValidate` and `StartAsyncValidate` are guarded upstream by `IsDaemonRunning()`/`IsDaemonCompatible()`, which absorb `SocketPath()` errors. `CollectValidateResults` has no such guard.

```suggestion
	sockPath, err := SocketPath()
	if err != nil {
		return nil, fmt.Errorf("%w: %w", ErrDaemonUnavailable, err)
	}
```

internal/watchd/tasks.go
michael-webster: **`collect()` has no non-test callers — doc comment could mislead future readers**

`handleCollect` deliberately uses `peek`+`acknowledge` separately (write must succeed before acknowledging). `collect()` collapses those two steps and cannot safely be used in any HTTP handler. There are no production callers — only test callers in `asyncvalidate_test.go`.

The doc comment says "for callers with nothing that can fail in between", which sounds like a legitimate building block. A future reader adding a second endpoint or an internal consumer is likely to reach for `collect()` and miss the ordering requirement. Either add a `// only safe in tests` note or remove it and leave `peek`/`acknowledge` as the only API.

michael-webster (commented):
**Multi-agent review** — 6 reviewers × adversarial verification. 5 confirmed findings below.

internal/watchd/tasks_test.go
```diff
@@ -0,0 +1,1079 @@
...
+
+	for _, args := range [][]string{
+		{"init"},
+		{"config", "user.email", "test@example.com"},
+		{"config", "user.name", "test"},
+		{"commit", "--allow-empty", "-m", "init"},
+	} {
+		cmd := exec.Command("git", args...)
+		cmd.Dir = dir
+		out, err := cmd.CombinedOutput()
```
danmux, line 620: os.MkdirTemp + manual t.Cleanup(os.RemoveAll) here — project convention is t.TempDir() for all per-test directories including git repos.

internal/watchd/validate.go
```diff
@@ -35,6 +55,140 @@ type ValidateResponse struct {
...
+func (d *daemon) handleCollect(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	root := r.URL.Query().Get("root")
+	if root == "" {
+		http.Error(w, "root required", http.StatusBadRequest)
+		return
+	}
```
danmux, line 173: For a real http.ResponseWriter, w.Write will not return an error on client disconnect — the guard only catches OS-level write errors. The comment implies the window closes on disconnect, but it does not. Worth clarifying?

danmux (approved):

PR #581, opened 2026-09-10
Add risk assessment to the daemon

jesseworld22:
**Stack** - merge bottom-up; each diff is against the one above it:

1. #581 – Add risk assessment to the daemon  ← **this PR**

---

## Summary

The Stop hook blocks on every check, every turn - right for a risky change, expensive for a typo. The daemon now judges the change first and decides who waits.

- **The deal** - a Stop hook offers to be released (`allow_async` on `POST /validate`); the daemon runs the checks inline, or backgrounds them and answers with a task ID, collected next turn via #580's hook.
- **Released:** under 500 lines, or docs and text only. **Held:** bigger, unmeasurable, or last run failed.
- **Measured from the last passing state, not `HEAD`** - five turns of 100 lines shouldn't add up to a blocking 500.
- **Failures block one turn later** - a background failure is remembered, and the next run held until one passes.
- **Only the Stop hook is released** -`PreToolUse` looks identical from here, so the decision reads `hook_event_name`.
- **A score with its workings, 0–100** - size, files, failure debt; docs and text score zero. It reports, it doesn't decide.
- **Judged against this repo** - finished runs in `risk-history.jsonl` (sizes and verdicts, no code) answer whether this is big *here*.
- **History only tightens** - it can hold a run the rules would have released, never release one.
- **Configurable** - `asyncValidate` (`auto`/`always`/`never`), `asyncValidateMaxLines`.

## Test plan

- [x] `TestSmallTurnsDoNotAccumulateIntoALargeChange` confirmed to fail with the baseline removed - it holds turn 5 on "large change, 500 lines" made of five identical released turns.
- [x] Units: the policy table, endpoint round trips, the failure-debt loop, the `PreToolUse` exclusion, the score (caps, docs ceiling, parts, advice) and the history (comparison band, percentile, survives a restart, can't release a large change).
- [x] By hand against a real daemon: small change let go and collected next turn; 600 lines held; a 900-line docs change let go; a background failure collected with its output and the next run blocking with exit 2; `asyncValidate` modes honoured; everything blocking with no daemon; six turns of 150 lines all let through with 900 lines outstanding against `HEAD`.


internal/watchd/score_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+func TestScoreRisesWithBreadth(t *testing.T) {
+	narrow := scoreChange(500, false, change(100, "a.go"), nil, historyEvidence{})
+	wide := scoreChange(500, false, change(100, "a.go", "b.go", "c.go", "d.go", "e.go"), nil, historyEvidence{})
+	assert.Assert(t, wide.Score > narrow.Score, "%d is not above %d", wide.Score, narrow.Score)
+}
+
+// A docs change is not a small source change, it is a change nothing should
+// fail on. A thousand lines of it must not out-score three lines of Go.
+func TestADocsChangeIsCappedBelowASmallSourceChange(t *testing.T) {
```
hanabel1: The test name says docs should score below a small source change, but the assertion says `docs.Score >= source.Score`. With current weights, 5000 lines of docs hits the `intertCeil` of 10, while 3 lines of Go scores 2 - so the assertion passes while docs is actually outscoring source. Should be <, not >=.

internal/watchd/history.go
```diff
@@ -0,0 +1,223 @@
...
+	rec := historyRecord{
+		At:     time.Now(),
+		Lines:  risk.Lines,
+		Files:  risk.Files,
+		Inert:  risk.Inert,
+		Passed: passed,
+	}
+	canonical := config.CanonicalProjectRoot(root)
+
+	h.mu.Lock()
```
hanabel1: h.mu is held while h.load reads from disk. Since there's one mutex for all projects, a first-access disk read for project A blocks concurrent calls for every other project. The fix is to check without the lock, read disk, then re-acquire and store - or key a sync.Map by canonical root.

docs/HOOKS.md
```diff
@@ -12,6 +12,133 @@ Quality checks that run automatically as Claude Code works.
 when the working tree is clean. When there are changes, it runs all configured
 commands so problems are surfaced before the agent stops working.
 
+<<<<<<< HEAD
+=======
+**UserPromptSubmit** — runs `chunk validate results`, which prints what
+background runs (`chunk validate --async`) concluded and then forgets them. It
```
danmux, line 16: Unresolved merge conflict markers here. This would ship broken docs and likely fail CI doc-linting.

danmux (changes requested):
Unresolved merge conflict markers in docs/HOOKS.md — needs resolving before merge.

danmux (approved):

PR #582, opened 2026-09-11
Add PR monitoring to the watch daemon

hanabel1:
## Summary

- The watch daemon now polls GitHub for the open PR on each project's current branch and surfaces it as advisory `pr` data in the snapshot JSON
- Agents get visibility into failing/in-progress CI checks, review thread comments, and `CHANGES_REQUESTED` reviews without being blocked by them
- A branch change resets the fetch clock so new PRs are picked up promptly; polling otherwise runs at a 60-second interval to respect GitHub rate limits

## Events surfaced

- **Check state** — overall rollup (`SUCCESS`, `FAILURE`, `PENDING`, etc.) plus individual check names and conclusions
- **Review thread comments** — one entry per thread (the opening comment), with author, body, timestamp, and resolved status
- **Changes requested** — boolean flag when any reviewer has an outstanding `CHANGES_REQUESTED` review
Screenshot 2026-09-11 at 8 55 51 AM

## Not collected

- PR body / description
- General issue comments (only inline review thread comments)
- Commit-level comments

## Opt-out

PR monitoring is automatically disabled when no GitHub token is configured. All failures are logged and silently swallowed — the daemon never errors out due to a missing or broken GitHub token.

## Implementation

- `internal/github/pr.go` — `FetchPRForBranch` method using a new `branchPRQuery` GraphQL query; handles the `CheckRun | StatusContext` union via `__typename`
- `internal/watchd/prmonitor.go` — `prMonitor` with per-branch inflight tracking and `maybeRefresh` called from the main poll path
- `internal/watchd/types.go` — `PRState`, `PRCheck`, `PRComment` types; `ProjectSnapshot.PR` field
- `internal/watchd/daemon.go` — org/repo resolved once from git remote per project; `prm` wired into `updateProject` and `RunDaemon`
- `internal/cmd/watchdaemon.go` — GitHub client resolved non-interactively alongside the existing CircleCI client

## Test plan

- [x] All existing tests pass (`task test`)
- [x] `go build ./...` clean
- [x] `golangci-lint` clean after `gofmt` fix


internal/watchd/prmonitor.go
```diff
@@ -0,0 +1,141 @@
...
+		return
+	}
+
+	var state *PRState
+	if info != nil {
+		state = convertPRInfo(info)
+	}
+
+	pm.mu.Lock()
+	pm.states[root] = state
```
schurchleycci, line 97: **Cached PR state is keyed by project root only, with no branch dimension.**

`states` is `map[string]*PRState` (line 30), while `lastFetch` and `inflight` are keyed by `prProjectKey{root, branch}`. Two things fall out of that:

- Right after a `git checkout`, the snapshot still reports the previous branch's PR. This one is self-healing — the new `(root, branch)` key has a zero `lastFetch`, so the next poll fetches immediately — so it's a single poll interval of wrong data.
- The error path is the real problem. Lines 85-88 record `lastFetch[key]` and return without touching `states[root]`. If the new branch's fetch keeps failing (revoked token, SAML 403, rate limit), the dashboard shows the *previous* branch's PR indefinitely — wrong number, wrong URL, wrong check state, and no staleness signal.

Storing the branch alongside the state and clearing on mismatch fixes both, and incidentally bounds `states` growth (today it accumulates one entry per project forever, which is small but unbounded).

internal/github/pr.go
```diff
@@ -0,0 +1,177 @@
...
+		info.Comments = append(info.Comments, PRCommentInfo{
+			Author:    author,
+			Body:      c.Body,
+			CreatedAt: ts,
+			Resolved:  thread.IsResolved,
+		})
+	}
+
+	// Flag CHANGES_REQUESTED.
+	for _, r := range pr.Reviews.Nodes {
```
schurchleycci, line 169: **`ChangesRequested` is derived from raw review history, so it never clears.**

This scans `reviews(last: 20)` (queries.go:25) and breaks on the first `CHANGES_REQUESTED`. Dismissal is actually fine — a dismissed review's state flips to `DISMISSED` — but **re-approval is not**: the historical `CHANGES_REQUESTED` node stays in the list with that state, so the PR reads as blocked for the rest of its life.

`pullRequest.reviewDecision` is the right field here. One caveat on swapping unconditionally: `reviewDecision` can be `null` on repos with no review requirement in branch protection. Suggest using it when non-null and keeping a fallback that takes the *latest* review state per reviewer rather than first-match-wins.

internal/github/pr.go
```diff
@@ -0,0 +1,177 @@
...
+func (c *Client) FetchPRForBranch(ctx context.Context, owner, repo, branch string) (*PRStatusInfo, error) {
+	var resp graphQLResponse[branchPRData]
+	vars := map[string]any{"owner": owner, "repo": repo, "branch": branch}
+	if err := c.doWithRetry(ctx, branchPRQuery, vars, &resp); err != nil {
+		return nil, err
+	}
+	if hasResolutionError(resp.Errors) {
+		return nil, nil // repo not found or not accessible — not fatal for monitoring
+	}
+	if resp.Data == nil || resp.Data.Repository == nil {
```
schurchleycci, line 104: **Non-resolution GraphQL errors are silently read as "no open PR".**

`do` (client.go:84-92) only surfaces transport and HTTP-status errors, and `doWithRetry` (retry.go:79-84) returns nil whenever `do` does. GraphQL `RATE_LIMITED`, SAML `FORBIDDEN`, and `INSUFFICIENT_SCOPES` all come back as **HTTP 200** with a populated `errors[]` and `data: null`. `hasResolutionError` (repos.go:50-57) only substring-matches `"Could not resolve"`, so those fall through to this `resp.Data == nil` branch and return `nil, nil`.

The caller caches that as "no PR" and logs nothing, so there's no way to distinguish a broken token from a branch that genuinely has no PR. Returning an error for unrecognized `resp.Errors` would let prmonitor.go:83's existing `log.Printf` do its job.

internal/github/queries.go
```diff
@@ -12,6 +12,62 @@ const orgReposQuery = `
   }
 `
 
+// branchPRQuery fetches the open PR for a given branch with check status and comments.
+const branchPRQuery = `
+  query BranchPR($owner: String!, $repo: String!, $branch: String!) {
+    repository(owner: $owner, name: $repo) {
+      pullRequests(first: 1, headRefName: $branch, states: [OPEN]) {
```
schurchleycci, line 19: **`headRefName` isn't scoped to an owner.**

A PR opened from a fork lives in the base repo with the fork's branch name as its `headRefName`, so this matches on name alone. Both directions bite:

- A common branch name (`renovate/...`, `fix-tests`) can surface a stranger's fork PR, with their checks and their review comments. `first: 1` with no `orderBy` makes the pick arbitrary when several match.
- Conversely, `DetectOrgAndRepo` reads `origin` (gitremote/detect.go:23), so a developer whose `origin` is their own fork queries the fork and finds nothing.

Worth noting the mechanics for whoever picks this up: `headRepositoryOwner` is **not** an argument on `pullRequests` — the filters are `headRefName`, `baseRefName`, `states`, `labels`, `orderBy`. So this needs `headRepositoryOwner { login }` in the selection set plus client-side filtering, with `first` raised above 1. The fork half needs separate handling (resolve `upstream`, or follow `repository.parent`) and is probably a follow-up rather than this PR.

internal/cmd/watchdaemon.go
```diff
@@ -38,7 +39,18 @@ func newWatchDaemonCmd() *cobra.Command {
 			if err == nil {
 				client, err = authprompt.ResolveCircleCIClient(rc, nil)
 			}
-			return watchd.RunDaemon(cmd.Context(), client, authMessage(err), makeValidateRunner())
+
+			// Resolve the GitHub client for PR monitoring. Failure is not fatal:
+			// PR monitoring is advisory and the daemon runs fine without it.
+			var ghClient *github.Client
+			if fullRC, rcErr := config.Resolve("", "", false); rcErr == nil && fullRC.GitHubToken != "" {
```
schurchleycci, line 46: **The env token is discarded when the config file can't be read.**

`config.Resolve` returns a fully populated `ResolvedConfig` *and* `Load()`'s error (config.go:367, 403), but this gates on `rcErr == nil`. So a corrupt `~/.chunk/config.json` disables PR monitoring even when `CHUNK_GITHUB_TOKEN` is set. Narrow trigger — `Load` only errors on unmarshal failure or a non-ENOENT read error — but checking `fullRC.GitHubToken != ""` is a one-line change and matches this comment's own "failure is not fatal" intent.

Separately: `Resolve` also calls `resolveAnthropicAPIKey`, which hits the keychain (config.go:379 → :204) for a key this daemon never uses. To be fair to the comment 15 lines up, `ResolveCircleCI` *already* does a `keyring.Get` for the CircleCI token, so this isn't keychain-free code being made keychain-touching — it's just one unnecessary keychain read in a detached process. A small `ResolveGitHub(insecureStorage)` mirroring `ResolveCircleCI` would cover both points.

internal/watchd/daemon.go
```diff
@@ -228,6 +245,10 @@ func (d *daemon) updateProject(ps *projectState) {
 	annotateActivity(sidecars, ps.events)
 	d.res.annotate(sidecars)
 
+	// Kick off a background PR fetch if one is due. Uses a background context
+	// derived from the daemon's own: the fetch must survive this poll returning.
+	d.prm.maybeRefresh(context.Background(), ps.root, branch, ps.org, ps.repo)
```
schurchleycci, line 250: **Comment and code disagree: this passes `context.Background()`, not something derived from the daemon's context.**

Flagging this as more than a stale comment because output.go:176 documents the exact opposite convention for this package ("deriving from the daemon's context rather than `context.Background()` ties a streamer's lifetime to the daemon's own"), and output_test.go:304 has a test enforcing it. As written, an in-flight PR fetch can outlive daemon shutdown.

`newPRMonitor(ctx, ghClient)` mirroring `newOutputStore(ctx)` would make this consistent — `poll` has no ctx to hand down, so the monitor holding it at construction is the natural fix.

internal/watchd/daemon.go
```diff
@@ -47,15 +52,19 @@ type daemon struct {
 	out *outputStore
 	// res samples resource usage, only while a dashboard is attached.
 	res *resourceSampler
```
schurchleycci, line 54: **This comment is inaccurate, and it's the reason for a vestigial nil guard.**

`prm` is never nil — line 104 assigns `newPRMonitor(ghClient)` unconditionally. The nil-ness lives on `prm.client` instead (prmonitor.go:27, checked at :47). That leftover shows up as the `pm == nil` guard in `annotate` (prmonitor.go:104), which `maybeRefresh` and `fetch` don't have.

Suggest fixing the comment rather than adding guards to match — the asymmetry is harmless today, but the comment actively misleads anyone deciding whether they need a nil check.

schurchleycci (commented):
Went through the review findings against the code on this branch. All five filed items check out; a couple of the proposed fixes need adjusting, and two of the "not filed" items are worth more than they were credited. Details inline.

**Summary**

| # | Finding | Verdict |
|---|---|---|
| 1 | `states` keyed by root only | Valid — highest priority; error path makes stale state permanent |
| 2 | `ChangesRequested` from raw review history | Valid — `reviewDecision`, but keep a fallback for its `null` case |
| 3 | GraphQL errors read as "no open PR" | Valid — silent failure, nothing logged |
| 4 | `headRefName` unscoped to owner | Valid problem, but `headRepositoryOwner` isn't a `pullRequests` filter |
| 5 | Env token dropped on config read failure | Valid; the keychain half is real but overstated |

Correctly **not** filed: the two-lock check-then-set on `inflight` is safe — `poll` runs `updateProject` sequentially (daemon.go:184-186) — and unbounded `lastFetch` growth is negligible. Raised two of the others inline (daemon.go:250, daemon.go:54).

The `resources_test.go` race fix looks right to me too.

internal/watchd/prmonitor.go
```diff
@@ -0,0 +1,141 @@
...
+
+// prMonitor fetches and caches the open PR state for each watched project's
+// current branch. It is advisory: failures are logged and silently ignored so
+// a missing or unauthenticated GitHub token never breaks the daemon.
+type prMonitor struct {
+	client *github.Client // nil when no GitHub credentials
+
+	mu        sync.Mutex
+	states    map[string]*PRState // latest PR state per project root (nil = no open PR)
+	lastFetch map[prProjectKey]time.Time
```
michael-webster, line 31: **Blocker — stale PR data surfaced after branch switch.**

`states` is keyed by `root` only (`map[string]*PRState`), but `lastFetch`/`inflight` are keyed by `prProjectKey{root, branch}`. When the user switches branches, a background fetch for the new branch fires promptly — but the old branch's `PRState` remains in `states[root]` until that fetch completes (up to 30 s). During that window `annotate` unconditionally writes the old PR number, title, checks, and review status onto a snapshot whose `Branch` field already reflects the new branch. Consumers can't detect the mismatch because `PRState` carries no branch field.

Fix: either key `states` by `prProjectKey` as well, or clear `states[root]` inside `maybeRefresh` when it detects a branch change before the new fetch returns.

internal/github/pr.go
```diff
@@ -0,0 +1,177 @@
...
+			Body:      c.Body,
+			CreatedAt: ts,
+			Resolved:  thread.IsResolved,
+		})
+	}
+
+	// Flag CHANGES_REQUESTED.
+	for _, r := range pr.Reviews.Nodes {
+		if r.State == "CHANGES_REQUESTED" {
+			info.ChangesRequested = true
```
michael-webster, line 171: **Blocker — `ChangesRequested` is permanently true after a reviewer approves.**

`reviews(last: 20)` returns nodes oldest-first. This loop breaks on the first `CHANGES_REQUESTED` without checking whether a later review from the same author supersedes it. A reviewer who blocked and then approved will cause `ChangesRequested: true` to be set forever — the `APPROVED` node is never reached.

Simplest fix: add `reviewDecision` to the GraphQL query. GitHub already computes the correct aggregate (`APPROVED`, `CHANGES_REQUESTED`, `REVIEW_REQUIRED`) accounting for re-reviews and dismissals:

```graphql
pullRequests(first: 1, headRefName: $branch, states: [OPEN]) {
  nodes {
    reviewDecision
    # remove the reviews(last: 20) block
    ...
  }
}
```

Then `info.ChangesRequested = pr.ReviewDecision == "CHANGES_REQUESTED"` and the manual loop is gone.

internal/watchd/prmonitor.go
```diff
@@ -0,0 +1,141 @@
...
+		states:    make(map[string]*PRState),
+		lastFetch: make(map[prProjectKey]time.Time),
+		inflight:  make(map[prProjectKey]bool),
+	}
+}
+
+// maybeRefresh fires a background fetch for the project's branch if one is due.
+// It is called from the daemon's poll path and must not block.
+func (pm *prMonitor) maybeRefresh(ctx context.Context, root, branch, org, repo string) {
+	if pm.client == nil || branch == "" || org == "" || repo == "" {
```
michael-webster, line 47: **Suggestion — `prMonitor` has no tests.**

This file introduces 141 lines of concurrency-sensitive logic (rate-limiting, inflight deduplication, branch-change clock reset) with zero coverage. Per project rules, tests should exercise real goroutines under `-race` with fake HTTP servers rather than mocks.

Key scenarios to cover:
- `maybeRefresh` does not double-fire when already inflight
- `maybeRefresh` resets the poll clock on branch change (new `prProjectKey` → immediate fetch)
- `annotate` returns the cached state and leaves `snap.PR` nil when no state is stored
- `fetch` records `lastFetch` on error so the back-off timer advances even on failure

(The stale-state bug above would have been caught by a test that checks the state key after a branch switch.)

internal/github/pr.go
```diff
@@ -0,0 +1,177 @@
...
+	Conclusion string `json:"conclusion"`
+	// StatusContext fields
+	Context string `json:"context"`
+	State   string `json:"state"`
+}
+
+// FetchPRForBranch returns the open PR for the given branch, or nil if none exists.
+// Returns nil, nil when the repo has no open PR for that branch.
+func (c *Client) FetchPRForBranch(ctx context.Context, owner, repo, branch string) (*PRStatusInfo, error) {
+	var resp graphQLResponse[branchPRData]
```
michael-webster, line 96: **Suggestion — `FetchPRForBranch` has no tests.**

The function dispatches on `__typename` to handle the `CheckRun`/`StatusContext` union, silently handles `hasResolutionError`, and parses timestamps with a fallback — none of which are exercised. An `httptest.NewServer` returning canned GraphQL JSON would cover these paths end-to-end without any mocks. Both correctness bugs identified in this review (`ChangesRequested` and the `states` key) would have surfaced under such a test.

internal/cmd/watchdaemon.go
```diff
@@ -38,7 +39,18 @@ func newWatchDaemonCmd() *cobra.Command {
...
+				ghClient, _ = github.New(github.Config{
```
michael-webster, line 47: **Nit — redundant `config.Resolve` call.**

`rc` from `config.ResolveCircleCI` is already in scope. `ResolveCircleCI` intentionally skips GitHub/Anthropic keychain work, but since `config.Resolve` is called unconditionally immediately after, that optimisation has no effect here — both calls run every time. Calling `config.Resolve` once and passing `rc.CircleCIToken` (or equivalent) to `authprompt.ResolveCircleCIClient` would eliminate the extra I/O round-trip.

michael-webster (dismissed):
Multi-agent review (6 reviewers + adversarial verification). Two correctness blockers confirmed; all other flagged issues verified and refuted or downgraded.

**Sidecar validate:** changed-file tests pass; one pre-existing flaky failure in `TestAssemblePool_OutdatedExisting_ReplacedAutomatically` (temp dir cleanup, unrelated to this PR).

danmux (approved):

danmux (approved):

PR #584, opened 2026-09-11
Measure trees git can't answer for, and run checks in a snapshot

jesseworld22:

1. #584 – Measure trees git can't answer for, and run checks in a snapshot  ← **this PR**

---

## Summary

- **Measure a tree git can't answer for** – a repo with no commits has no `HEAD` to diff against, and a directory that was never a repo has nothing at all, so every run in them blocked. `internal/filestate` walks and hashes instead, with git still answering first where it can. Hashing counts a modified file whole, which over-measures and makes you wait, rather than under-measuring and waving a change through.
- **Keep a result the tree moved past** (`asyncValidateWorktree`, off by default) – checking the snapshot out into its own worktree means the copy can't move, so the verdict stays true about the state the turn produced. Off by default because a snapshot holds nothing gitignored, and those projects would see environment failures reported as code failures.
- **The trap, found by running it rather than testing it** – a fresh checkout is a *clean* tree, and the Stop hook skips clean trees, so the first honest run executed nothing and reported a pass. `--attribute-to` exempts it and keeps the event log under the real repo.
- **Stop wasting work on obsolete runs** – two runs for one project could be in flight at once, the first validating code that had already moved while holding the lock the second wanted. `taskStore.supersede` cancels it and records nothing, since a cancelled run concluded nothing and shouldn't owe a blocking run. Snapshot runs are exempt.
- **Known gap** – remote commands in a snapshot run register their output under the copy, so the `chunk watch` logs pane won't show it.

## Test plan

- [x] `filestate`: whole-tree and incremental diffs, a modified file counting whole, `.git` skipped, simple gitignore forms honoured and unclear ones walked anyway, binary content, the budget, digest stability.
- [x] Snapshots: uncommitted work carried in, the copy not following the live tree, ignored content absent, cleanup of directory and git metadata; at the daemon, attributed back to the real project and a result surviving an edit mid-run while a live-tree result is still discarded.
- [x] Superseding: the in-flight run cancelled, nothing owed afterwards, snapshot runs left alone.
- [x] `TestASnapshotRunIsNotSkippedForBeingClean` confirmed to fail without the exemption.
- [x] By hand against a real daemon: `CWD=/tmp/chunk-shadow-…` with `CONTENT=ORIGINAL` while the live tree said `EDITED`, the failure reported and qualified, nothing left behind.

### DIFF
--- a/internal/filestate/filestate.go
+++ b/internal/filestate/filestate.go
@@ -1,0 +1,315 @@
+// Package filestate measures a working tree without asking git anything.
+//
+// It exists because git is not always there to ask. A repository with no
+// commits has no HEAD to diff against, and a directory that was never a
+// repository has nothing at all — and both are trees an agent can be working
+// in. More of them will be: keeping track of what changed without a commit
+// history is becoming its own problem, and a resident daemon watching a
+// filesystem is the shape of the answer.
+//
+// Where git can answer, it should: it knows what is ignored, and it knows how
+// much of a modified file an edit actually touched. This does not. See Changes
+// for what that costs and which way it errs.
+package filestate
+
+import (
+	"bufio"
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/CircleCI-Public/chunk-cli/internal/changeset"
+	"github.com/CircleCI-Public/chunk-cli/internal/hashutil"
+)
+
+// Walk budgets. Past either of them Build fails rather than returning a partial
+// index, because a partial index measures a smaller change than the real one —
+// and a change that reads smaller than it is, is a change that gets waved
+// through. Failing makes the caller wait instead.
+var (
+	maxFiles       = 20000
+	maxBytes int64 = 64 << 20
+)
+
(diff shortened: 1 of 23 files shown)


internal/filestate/filestate.go
```diff
@@ -0,0 +1,315 @@
...
+	if !info.IsDir() {
+		return nil, fmt.Errorf("read %s: not a directory", root)
+	}
+
+	skip := loadIgnores(root)
+	idx := make(Index)
+
+	var (
+		files int
+		bytes int64
```
hanabel1: the `bytes` package is only used in `hashAndCount`, not in `Build`'s closure. But any future edit that calls `bytes.Contains` etc. inside the closure will either ail to compile or silently misbehave. Rename to `bytesRead` or `totalBytes`.

danmux (approved):
Three non-blocking notes:

**internal/gitutil/shadow.go** — TOCTOU: `os.Remove(path)` then `git worktree add path` has a race window. If worktree add fails, the caller silently falls back to the live tree — intended, or should it surface?

**internal/watchd/validate.go** — First use of stdlib `log.Printf` in this file. Consistent with rest of daemon?

**internal/watchd/risk.go** — When `filestate.Build` fails, falls through and tries it a second time, losing the incremental baseline silently on double failure. Intended fallback, or should the first failure be logged?

PR #585, opened 2026-09-14
Move `chunk validate` to pools.

z00b:
This PR is broken into 3 commits to make it logical, but intermediate states are not useful in the product.

1. Update pools to handle current state of main
2. Move `chunk validate` to pools
3. Update `watch` to use the new data structures under pools

Can be reviewed separately or together. More details in each commit message.


schurchleycci:
Some Claude findings:

Correctness

1. validate/plan.go:35 — PlacementConfigured routes anything without local: true remotely, and there's no migration. Every pre-branch config has format as role: autofix with no local key, so gofmt -w . / task fmt now runs on the sidecar — the developer's tree never gets formatted. Hits all existing users on first validate after upgrade.
2. validate/ciconfig.go:141 — roleSpec (:132) has no Local field, so formatters derived from CI config can't be marked local like the setup.go:193,215 defaults. Same bug as 1, for new repos.
3. sidecar/pool.go:419 — passes false for retryOn404. FreshIDs exempts a just-created sidecar from the staleness probe (:152), so it lands in aliveExisting and this becomes its first sync; a transient post-creation 404 then reads as SidecarGone and the new sidecar is deleted and replaced. :465 passes true for exactly this case.
4. sidecar/pool.go:459, :346 — replacements use p.image, which is "" because the snapshot picked by autoSelectSnapshotImage (cmd/validate.go:1024) never leaves resolveOrCreateSidecarID. Replacements boot the bare image and fail on missing tooling. Compounds 3.
5. cmd/validate.go:888 — remoteWorkDir = entry.RepoPath passes straight through ResolveWorkspace (sync.go:46), making its active.Workspace branch unreachable. A sidecar with a custom workdir silently gets a second checkout.
6. watchd/load.go:102 — pool state is appended with blank SessionID/Workspace, and savePoolState runs every validate so its mtime always wins. In chunk watch the name becomes validate and session/workspace go blank.
7. cmd/validate.go:905 — OnSubmitted fires only on successful submit (target.go:86); main's setCommandID("") error-path clear was dropped, so a failed submit leaves a stale command ID on later event-log entries.

Error handling

8. sidecar/target.go:98 — ReadyExecRunner, WorkspaceExists, and the deleted unreachableSidecar/missingWorkspace are all unreferenced now. A missing workspace shows as exit code 1 plus raw cd: no such file, misreported as a test failure instead of the old actionable error.

Messaging / observability

9. cmd/validate.go:116 — "left local (autofix rewrites files)" is untrue under the new default; --mark-remote misreports what it did.
10. validate/validate.go:176 — the $ <command> echo is gone, so expanded commands (notably {{CHANGED_PACKAGES}}) are invisible.

Dead code

11. cmd/validate.go:198 — opts.remote is written at :398, read nowhere; --remote does nothing.

z00b:
fixed all the comments from @schurchleycci 's review other than the backwards compatibility which i intentionally ignored. moving to default remote is fine and preserving the default local is not worth the complexity.

fixed a couple more things while i was at it. one CI test failed that passed locally, but it looks like it's on its way to passing now.

schurchleycci (approved):
Couple small cleanup things but can always do as a followup:

1. Remote→local fallback plumbing is never exercised — internal/validate/distributed.go

DistributedJobResult.Fallback (:17), .UnavailableErr (:19), DistributedRunResult.FellBack (:32), and .UnavailableErr (:35) have no producer. The only caller that builds a DistributedJobResult is runPooledValidateCommand (internal/cmd/validate.go:882), which returns {Err} on its three failure paths and {Passed, Err} on success. Branch-wide, those fields are set only in distributed_test.go:106-108.

2. HasRemoteCommands() is now dead and misnamed — internal/config/project.go:88

With the predicate changed to cmd.RunsRemotely() (i.e. !cmd.Local), this is equivalent to HasCommands() — the only real difference is the nil-receiver guard. Both callers were deleted in this PR (checkRemoteConfigured and validateNeedsSidecar, previously cmd/validate.go:275,300), leaving one assertion at cmd/validate_test.go:826.

PR #586, opened 2026-09-14
Route pre-commit hooks through chunk validate for daemon + cache benefits

hanabel1:
## Summary

- Pre-commit (PreToolUse) hooks previously ran configured commands directly (e.g. `task test`), bypassing `chunk validate` entirely — no daemon delegation, no result caching
- Now hooks generate `chunk validate <name>` instead of the raw command, so pre-commit and Stop hooks share the same execution path
- `tryHookDelegate` routes to the watch daemon when running; the result cache skips re-runs when the working tree fingerprint matches a prior success

## Events

- `chunk init` generates PreToolUse hooks calling `chunk validate <name>` (e.g. `chunk validate test`)
- `chunk init` generates Stop hook calling `chunk validate` (unchanged)
- Both hook types now get daemon delegation and result caching automatically

## Not collected

- No behavior change when the daemon is not running — falls back to inline execution as before
- No change to the Stop hook command itself

## Opt-out

- `chunk hook disable` still disables the Stop hook
- `CHUNK_HOOKS_DISABLED` still suppresses Stop hook validation

## Implementation

- `Build()` and `BuildCodex()` in `settings.go`: generate `chunk validate <name>` for PreToolUse hook entries instead of `cmd.Run`
- Updated tests to assert `chunk validate <name>` as the hook command

## Test plan

- [x] `go test -race ./internal/settings/... ./internal/cmd/...` passes


schurchleycci (approved):
LGTM!

PR #589, opened 2026-09-15
Make background validation thresholds configurable

### LINKED ISSUE #588: Make background validation thresholds configurable
Follows #581, which added the daemon's decision about which changes are validated in the background. Two of its inputs are not something a project can set today.

## What's already configurable

`asyncValidate` (`auto`/`always`/`never`) and `asyncValidateMaxLines` are project config in `.chunk/config.json`, so the line threshold is covered.

## What isn't

- **The extension lists are hardcoded** in `internal/watchd/risk.go` – `inertExtensions` (`.md`, `.markdown`, `.txt`, `.rst`, `.adoc`) and `inertNames` (`LICENSE`, `NOTICE`, `AUTHORS`, `CHANGELOG`). A repo that lints its markdown, or one where a `.sql` or `.json` change matters, has no way to say so.
- **Nothing lets a project name extensions that must always block.** That's the direction worth having: it only ever tightens, so it can't wave a change through by mistake.

Worth preserving: `inertExtensions` is an allowlist on purpose, not a list of source extensions. An unfamiliar extension currently makes the developer wait; a denylist would background the run and hope. Config shouldn't be able to flip that direction by accident.

## Scope

- [ ] Project config for the inert extensions and names – add to the defaults, or replace them.
- [ ] Project config for extensions that always require a blocking run.
- [ ] Both validated on load, with a clear error rather than a silently ignored key.
- [ ] Skills updated so an agent can set this up from a conversation, not by hand-editing JSON.
- [ ] `docs/CLI.md`, `docs/HOOKS.md` and `docs/SKILLS.md` updated.

## Notes

No skill mentions `asyncValidate` yet – only `docs/CLI.md` and `docs/HOOKS.md` do – so the skills piece is new work rather than an edit.

jesseworld22:

1. #589 – Make background validation thresholds configurable  ← **this PR**

---

Closes #588.

## Summary

#581 made the line threshold configurable and left the other half of the judgement – *which paths* it applies to – hardcoded. A repo that lints its markdown had no way to say the default was wrong for it.

- **`asyncValidateInert`** – extensions (`.sql`) or exact file names (`NOTICE`) this project also counts as prose, released at any size. It adds to the built-in list rather than replacing it: that list is an allowlist, and dropping an entry silently would release a change instead of holding it.
- **`asyncValidateBlocking`** – paths that always block, over an inert default, over the line threshold, and over `asyncValidate: always`. It is the narrower instruction and it only ever makes somebody wait, so it wins. `never` is unaffected, having already blocked everything.
- **`chunk config set` takes all five `asyncValidate*` keys now** – it previously accepted only `orgID` and `validation.sidecarImage`, so an agent had to hand-edit JSON.
- **Bad rules are refused when set, not ignored** – a path, a glob, an unknown mode, a non-numeric line count, or an entry in both lists gets an error naming what to write instead. A misspelt rule otherwise fails in the direction nobody checks: the project believes it is waiting for its migrations and nothing says otherwise.
- **New `chunk-validate-config` skill**, installed by `chunk init`. It asks what the user wants, reads the repo rather than guessing, sets the narrowest key that does it, and refuses to put source extensions in the inert list or flip `asyncValidateWorktree` without asking.

## Test plan

- [x] Units: a configured extension treated as prose; blocking beating the inert defaults, a small diff, and `always`; saying nothing under `never`; `policyFor` reading both lists, case-insensitively.
- [x] Config: each unusable rule shape rejected on load – a path, a glob, an empty entry, a bare dot, and the same entry in both lists.
- [x] Acceptance: the keys driven through the real binary the way the skill drives them, including a refused path.
- [x] Against a real daemon: 1 line of `.sql` with `asyncValidateBlocking` blocked with "a path this project always validates changed"; 3,000 lines of `.sql` with `asyncValidateInert` released as prose; the same 3,000 lines with no config blocked as a control.

## Notes

- `skills_test.go` now derives its fixture from `skills.All` rather than a hardcoded list of five names, which was failing four tests that have nothing to say about a new skill.
- `ProjectConfig.Validate()` is exported so `config set` can reject input before `SaveProjectConfig` wraps the error in a "check file permissions" suggestion.


internal/watchd/score_test.go
```diff
@@ -25,22 +25,22 @@ func TestScoreRisesWithSize(t *testing.T) {
...
+	at := scoreChange(500, inertRules{}, false, change(500, "main.go"), nil, historyEvidence{})
+	way := scoreChange(500, inertRules{}, false, change(50000, "main.go"), nil, historyEvidence{})
 	assert.Equal(t, at.Score, way.Score)
 }
 
 func TestScoreRisesWithBreadth(t *testing.T) {
-	narrow := scoreChange(500, false, change(100, "a.go"), nil, historyEvidence{})
-	wide := scoreChange(500, false, change(100, "a.go", "b.go", "c.go", "d.go", "e.go"), nil, historyEvidence{})
+	narrow := scoreChange(500, inertRules{}, false, change(100, "a.go"), nil, historyEvidence{})
+	wide := scoreChange(500, inertRules{}, false, change(100, "a.go", "b.go", "c.go", "d.go", "e.go"), nil, historyEvidence{})
```
hanabel1, line 35: `wide` wasn't updated to the new `scoreChange` signature - `false` (bool) is passed where inertRules is expected. It won't compile!

fix: add inertRules{} as a second arg

internal/config/project.go
```diff
@@ -213,6 +232,42 @@ func (c *ProjectConfig) validate() error {
...
+		case strings.TrimSpace(entry) == "":
+			return fmt.Errorf("%s has an empty entry", key)
+		case entry == ".":
+			return fmt.Errorf("%s: %q is not an extension or a file name", key, entry)
+		case strings.ContainsAny(entry, `/\`):
+			return fmt.Errorf("%s: %q looks like a path; use an extension (\".md\") or a file name (\"NOTICE\")", key, entry)
+		case strings.Contains(entry, "*"):
+			return fmt.Errorf("%s: %q looks like a glob; use an extension (\".md\") or a file name (\"NOTICE\")", key, entry)
+		}
+	}
```
danmux, line 270: The comment warns about bare extensions (sql without dot) never matching, but there is no check that catches it. "sql" passes all guards and is stored as a literal filename match, silently never matching .sql files.

internal/watchd/risk.go
```diff
@@ -48,13 +48,75 @@ var inertNames = map[string]bool{
...
+func newInertRules(inert, blocking []string) inertRules {
+	return inertRules{inert: ruleSet(inert), blocking: ruleSet(blocking)}
+}
+
+func ruleSet(entries []string) map[string]bool {
+	if len(entries) == 0 {
+		return nil
+	}
+	set := make(map[string]bool, len(entries))
+	for _, e := range entries {
```
danmux, line 75: Extensions are lowercased in ruleSet but file names are not. A config entry "notice" will not match the file "NOTICE". Is case-folding for file names intentionally omitted?

danmux (approved):

PR #591, opened 2026-09-15
Allow daemon to run remotely in a sandbox

hanabel1:
**Summary**

- Part one of https://linear.app/circleci/issue/FACT-475/allow-daemon-to-run-remotely-in-a-sandbox. This PR is just for transport and auth. Part two would be the remote daemon spawning other sidecars. 
- Adds TCP transport support to the watchd client and daemon so chunk can connect to a daemon running inside a remote sandbox rather than only over a local Unix socket
- CHUNK_WATCHD_TCP_ADDR tells the daemon to bind a TCP listener in addition to the Unix socket
- CHUNK_WATCHD_REMOTE_ADDR tells the client to dial that address instead of the local socket
- All existing client call sites (FetchSnapshot, RegisterCommand, FetchOutput, RunValidate, IsDaemonRunning, IsDaemonCompatible) route through the new daemonClient/longDaemonClient helpers automatically — no changes needed at call sites
**Remote:**
Screenshot 2026-09-15 at 2 37 11 PM

**Local:**
Screenshot 2026-09-15 at 2 37 28 PM

Implementation

- paths.go: two env-var accessors (TCPListenAddr, TCPRemoteAddr)
- ipc.go: tcpClient and longTCPClient — TCP equivalents of the existing Unix socket clients; custom DialContext ignores the URL hostname so callers keep using http://watchd/... URLs unchanged
- client.go: daemonClient/longDaemonClient choose Unix or TCP based on CHUNK_WATCHD_REMOTE_ADDR; pingDaemon replaces duplicated ping logic across IsDaemonRunning and IsDaemonCompatible
- daemon.go: conditionally starts a TCP listener when CHUNK_WATCHD_TCP_ADDR is set

Test plan

- [x] task build and cross-compile Linux binary
- [x] Start a sidecar, scp the binary in, run the daemon with CHUNK_WATCHD_TCP_ADDR=0.0.0.0:7777
- [x] Set CHUNK_WATCHD_REMOTE_ADDR=<sandbox-host>:7777 locally and run chunk sidecar status / a validation — confirms client routes over TCP
- [x] Unset CHUNK_WATCHD_REMOTE_ADDR and confirm Unix socket path still works


schurchleycci:
A few claude findings:

1. The TCP listener serves the same mux as the Unix socket with no authentication and no TLS, and /validate executes caller-supplied args through the cobra root — which includes sidecar exec and sidecar ssh. Anyone who can reach the port gets arbitrary command execution, plus env injection into the run.
2. RunValidate puts CircleCIToken and the client's entire os.Environ() in the request body. Over a Unix socket that's harmless; over plain HTTP to a sandbox host it's every local secret on the wire.
3. longTCPClient sets no dial, response-header, or total timeout, so an unreachable or mid-run-dead sandbox stalls chunk validate for an unbounded, OS-determined period instead of falling back to inline execution. Separately, tcpClient's 5s total budget was sized for a local socket and is tight for a remote round trip.
4. EnsureRunning, EnsureLaunched, and stopDaemon still operate on the local pid file and socket while FetchSnapshot and pingDaemon go remote. In remote mode loadWithRelaunch starts a stray local daemon and then retries a remote fetch that can't succeed, and maybeAutoLaunchDaemon starts a local daemon on nearly every command.
5. StopForCredentialChange only targets the local daemon, so after auth login/logout a remote daemon keeps serving with a revoked token.
6. CHUNK_WATCHD_TCP_ADDR and CHUNK_WATCHD_REMOTE_ADDR are missing from the env-var table in docs/ARCHITECTURE.md.
7. The new transport ships with no tests, despite an existing internal/watchd suite that already covers the IPC layer.

z00b:
## Review Summary

**Files reviewed**: 6
**Issues identified**: 4

---

### 1. Bearer token transmitted in plaintext over TCP
**`internal/watchd/ipc.go:139–175`** | **High**

The TCP transport (`tcpClient`, `longTCPClient`) uses plain HTTP with no TLS. The `Authorization: Bearer <token>` header — and all daemon responses, which may include validated command output — are visible to any observer on the network path. The daemon binding `0.0.0.0:7777` as the docs suggest is reachable from anywhere that can route to the sandbox host.

The `RunValidate` change already shows awareness of this (it withholds `circleCIToken` and `os.Environ()` over TCP), but the token used to authenticate those requests is still sent unencrypted and can be captured and reused.

If the intended deployment model is an SSH tunnel to the sandbox (which would encrypt the traffic), that assumption should be explicitly documented — in the env var table in ARCHITECTURE.md and ideally enforced or warned on at runtime. If direct TCP over the public internet is in scope, TLS is needed before this ships beyond manual testing.

---

### 2. Non-constant-time token comparison in `withBearerAuth`
**`internal/watchd/ipc.go:91–103`** | **High**

```go
if r.Header.Get("Authorization") != want {
```

String `!=` is not constant-time. A close network observer could use timing differences to recover the token byte by byte. Since the token is the sole authentication mechanism, this matters. Fix:

```go
if subtle.ConstantTimeCompare([]byte(r.Header.Get("Authorization")), []byte(want)) != 1 {
```

`crypto/subtle` is in the standard library — one-line change.

---

### 3. Security policy buried inside `RunValidate` via env-var read
**`internal/watchd/client.go:261–267`** | **Medium**

```go
if TCPRemoteAddr() == "" {
    req.CircleCIToken = circleCIToken
    req.Env = os.Environ()
}
```

Per the team's standards, env-var reads belong in `cmd/`, not business-logic functions. More concretely: the caller passes `circleCIToken` as an argument already. Whether to include it in the request is now an invisible side effect — callers can't observe or control it, and it's hard to test in isolation.

Consider an explicit `forwardEnv bool` parameter (or an opts struct), decided at the `cmd/` layer, so `RunValidate` is a pure "send this request" function.

---

### 4. TOCTOU port race in `startTestDaemonTCP`
**`internal/watchd/ipc_test.go:218–224`** | **Medium**

```go
ln, err := net.Listen("tcp", "127.0.0.1:0")
addr := ln.Addr().String()
assert.NilError(t, ln.Close())
// ← another goroutine/process can grab the port here
t.Setenv("CHUNK_WATCHD_TCP_ADDR", addr)
```

Between `ln.Close()` and the daemon's `net.Listen(addr)`, another process can claim the port — a classic source of intermittent CI failures. Cleaner fix: keep the listener open and pass it directly into the daemon (or expose a `RunDaemonWithListener` test variant), removing the race entirely.

internal/watchd/ipc.go
```diff
@@ -86,6 +86,36 @@ func newServer(d *daemon) *http.Server {
 	}
 }
 
+// withBearerAuth wraps h so that every request must carry a matching
+// Authorization: Bearer <token> header. When token is empty the handler is
+// returned unchanged (auth disabled).
+func withBearerAuth(h http.Handler, token string) http.Handler {
```
schurchleycci, line 110: **Blocker — auth fails open, exposing arbitrary command execution.**

When `token == ""` this returns the handler unwrapped, so setting `CHUNK_WATCHD_TCP_ADDR` without `CHUNK_WATCHD_TCP_TOKEN` binds the *entire* mux with no credential. This is worse than `/validate` being exposed: `internal/cmd/watchdaemon.go` builds a fresh root command from the request's `Args`, so **every** `chunk` subcommand is reachable, and `req.Env` flows through `envctx.WithEnv` into `cmd.Env`.

The docs table added in this PR presents exactly this configuration as valid — `0.0.0.0:7777` as the address example, directly above a token row whose default is documented as _(no auth)_.

The zero value of a security control shouldn't mean "disabled". Suggest failing closed in `RunDaemon`:

```go
if addr := TCPListenAddr(); addr != "" {
	if TCPToken() == "" {
		return fmt.Errorf("CHUNK_WATCHD_TCP_ADDR requires CHUNK_WATCHD_TCP_TOKEN")
	}
	...
```

Note `TestTCPTransport_NoTokenAllowsAll` currently pins the fail-open behaviour as intended, so it would need to change too.

internal/watchd/ipc.go
```diff
@@ -109,3 +139,37 @@ func longUnixClient(sockPath string) *http.Client {
...
+// longTCPClient is like tcpClient but with no total-request timeout, suitable
+// for long-running operations like /validate. A 10 s dial timeout bounds how
+// long an unreachable sandbox can stall the caller before ErrDaemonUnavailable
+// triggers a fallback to inline execution.
+func longTCPClient(addr string) *http.Client {
+	transport := &http.Transport{
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			return (&net.Dialer{Timeout: 10 * time.Second}).DialContext(ctx, "tcp", addr)
+		},
+		ResponseHeaderTimeout: 30 * time.Second,
```
schurchleycci: **Blocker — this 30s budget becomes a hard cap on the whole validate run.**

The doc comment above promises "no total-request timeout, suitable for long-running operations like `/validate`", but `handleValidate` buffers stdout/stderr into `bytes.Buffer` and writes the first byte of the response only *after* the runner returns. `ResponseHeaderTimeout` measures from "request fully written" until response headers arrive, so with a small JSON request body the clock effectively starts at send — 30s for the entire validation.

When it trips, the consequences compound:

- `RunValidate` wraps the transport error as `ErrDaemonUnavailable`, and `internal/cmd/validate.go` treats that as "no daemon" and **silently falls through to inline local execution** — the exact thing remote mode exists to avoid, with no message to the user.
- `handleValidate` uses `context.WithoutCancel`, so the abandoned remote run continues **while still holding `validateMu`** — the suite runs twice, and the next remote `/validate` blocks until the orphan finishes.

`longUnixClient` sets no equivalent timeout, so local and remote diverge with no stated reason. If the intent was only to bound an unreachable peer, the 10s `Dialer.Timeout` already does that — dropping `ResponseHeaderTimeout` looks like the fix.

internal/watchd/ipc.go
```diff
@@ -109,3 +139,37 @@ func longUnixClient(sockPath string) *http.Client {
 		},
 	}
 }
+
+// tcpClient returns an *http.Client that dials addr over TCP. The custom
+// DialContext ignores the URL hostname (kept as "watchd" for consistency with
+// the unix variants) and always connects to addr.
+func tcpClient(addr string) *http.Client {
```
schurchleycci, line 164: The bearer token and every payload cross the network in cleartext (plain `net.Dialer`, no `TLSClientConfig`, `http://` throughout), and there is no server-identity check in either direction.

I grepped for a mitigating deployment and couldn't find one: nothing in the repo sets `CHUNK_WATCHD_TCP_ADDR` or `CHUNK_WATCHD_REMOTE_ADDR`, and the existing SSH machinery (`startSSHProxy` in `internal/sidecar/rsync.go`) is not reused here — so there's no tunnel implied by the code, only whatever an operator assembles by hand.

The reverse direction matters too: `RunValidate`'s response is written straight to the user's terminal and, on the hook path, into an AI agent's context, with no signature or nonce — so a spoofed peer can assert "all checks passed".

This reads as a missing design decision rather than a code defect. If cleartext is an accepted constraint for this stage, could the docs say so explicitly and steer toward a loopback/tunnelled bind rather than offering `0.0.0.0` as the example?

internal/watchd/ipc.go
```diff
@@ -86,6 +86,36 @@ func newServer(d *daemon) *http.Server {
...
+// withBearerAuth wraps h so that every request must carry a matching
+// Authorization: Bearer <token> header. When token is empty the handler is
+// returned unchanged (auth disabled).
+func withBearerAuth(h http.Handler, token string) http.Handler {
+	if token == "" {
+		return h
+	}
+	want := "Bearer " + token
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Authorization") != want {
```
schurchleycci: Nit: `!=` on a secret short-circuits at the first differing byte (and on length first), leaking a prefix-match oracle to anyone who can reach the listener — and there's no rate limit or lockout in front of it. Given this guard is the only thing standing in front of command execution:

```go
if subtle.ConstantTimeCompare([]byte(r.Header.Get("Authorization")), []byte(want)) != 1 {
```

internal/watchd/ipc.go
```diff
@@ -109,3 +139,37 @@ func longUnixClient(sockPath string) *http.Client {
...
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			return (&net.Dialer{Timeout: 10 * time.Second}).DialContext(ctx, "tcp", addr)
+		},
+		ResponseHeaderTimeout: 30 * time.Second,
+	}
+	var rt http.RoundTripper = transport
+	if token := TCPToken(); token != "" {
+		rt = &bearerTransport{inner: transport, token: token}
+	}
+	return &http.Client{Transport: rt}
```
schurchleycci, line 195: With no `Client.Timeout` here and an unbounded `json.NewDecoder(resp.Body).Decode` in `RunValidate`, a peer that is alive but stalls mid-body blocks the CLI with no deadline — and unlike a dial failure, a mid-body read error is *not* wrapped in `ErrDaemonUnavailable`, so there's no inline fallback either.

Worth noting Ctrl-C doesn't rescue the user: `RunValidate` uses `client.Post` with no context, so `main.go`'s `signal.NotifyContext` never reaches this request, and after the first SIGINT Go's default die-on-interrupt is already replaced.

A verifier pushed back usefully here — TCP keepalives are enabled by default, so a *dead* peer is detected in ~15s–2min; genuinely unbounded needs a live-but-stalled peer (suspended sandbox, blackholed NAT entry). Also fair: `longUnixClient` is strictly less bounded and predates this PR. Lower priority than the rest, but a total timeout on the remote path would make the failure recoverable.

internal/watchd/daemon.go
```diff
@@ -115,6 +114,32 @@ func RunDaemon(ctx context.Context, client *circleci.Client, authMessage string,
...
+		if err != nil {
+			return fmt.Errorf("listen tcp on %s: %w", addr, err)
+		}
+		log.Printf("watch daemon started pid=%d socket=%s tcp=%s", os.Getpid(), sockPath, addr)
+		// The TCP server wraps the same handler with bearer-token auth so the
+		// Unix socket (bound to the local user's filesystem) stays unauthenticated
+		// while the TCP listener enforces a shared secret.
+		tcpSrv := &http.Server{
+			Handler:           withBearerAuth(srv.Handler, TCPToken()),
+			ReadHeaderTimeout: 5 * time.Second,
```
schurchleycci, line 181: `ReadHeaderTimeout` is the only timeout set on either server, and it stops applying once headers are parsed. Combined with `handleValidate` taking the process-wide `validateMu` *before* `json.NewDecoder(r.Body).Decode`, a client that sends complete headers and then stalls blocks inside the decode while holding the lock — wedging `/validate` for every other caller, including local Unix-socket ones, since both listeners share the same `*daemon`.

A verifier reproduced this against a standalone copy of the pattern (victim still blocked after 20s). There's also no `http.MaxBytesReader` on the body: a 200 MiB body measured ~456 MiB heap, and `Content-Length` is attacker-chosen.

Two cheap mitigations: decode the body *before* taking `validateMu`, and wrap `r.Body` in a `MaxBytesReader`. Setting `ReadTimeout`/`IdleTimeout` would help, though it conflicts with genuinely long validates on the same server.

internal/watchd/client.go
```diff
@@ -189,26 +231,30 @@ func IsDaemonRunning() bool {
 // not support all API endpoints (e.g. /validate), so delegation should be
 // skipped and the operation run inline instead.
 func IsDaemonCompatible() bool {
-	sockPath, err := SocketPath()
-	if err != nil {
-		return false
-	}
-	ok, build := ping(sockPath)
+	ok, build := pingDaemon()
```
schurchleycci, line 296: **`IsDaemonCompatible` can essentially never return true in remote mode.**

`BuildID()` is `version|os.Executable() path|size|mtime.UnixNano()`. A daemon in a Linux sandbox has a different path, a different size, and a different nanosecond mtime — a verifier cross-compiled this tree and got three of four fields differing, and `mtime.UnixNano()` alone makes a match essentially impossible even for a byte-identical binary at an identical path.

The consequence is an invisible split between invocation paths:

- `internal/cmd/validate.go` `shouldUseDaemon` → `IsDaemonCompatible()` → false, so **terminal `chunk validate` always runs inline locally**, with no message.
- `tryHookDelegate` gates on `IsDaemonRunning()` (ping only, no build check) → **hook-invoked validates do go remote**.

So the remote `/validate` path is only ever exercised from hooks, where nobody is watching. A build-identity check is meaningful for a daemon spawned from the same binary; remote mode needs a capability/version check instead.

(If remote direct-validate delegation is intentionally out of scope for this PR and lands in #593, saying so in the commit or docs would help — a verifier flagged that it couldn't rule that out.)

internal/watchd/client.go
```diff
@@ -189,26 +231,30 @@ func IsDaemonRunning() bool {
...
 // circleCIToken is forwarded to the subprocess as CIRCLE_TOKEN.
 func RunValidate(args []string, circleCIToken string) (ValidateResponse, error) {
-	sockPath, err := SocketPath()
+	client, err := longDaemonClient()
 	if err != nil {
 		return ValidateResponse{}, err
 	}
-	body, err := json.Marshal(ValidateRequest{Args: args, CircleCIToken: circleCIToken, Env: os.Environ()})
+	req := ValidateRequest{Args: args}
+	if TCPRemoteAddr() == "" {
```
schurchleycci: Withholding the local token and env over TCP is the right call, but this branch has **zero test coverage** — `RunValidate` has exactly one caller and no test anywhere calls it (I grepped `*.go` including `acceptance/`). Deleting this `if` would ship the developer's CircleCI token and full `os.Environ()` to a remote host with the suite green.

Related gap: `/validate` is never exercised over TCP at all — the test helper passes `nil` for both client and runner, so such a request returns 503 regardless of auth.

One test that posts to `/validate` over the TCP daemon with a real runner and asserts the received `ValidateRequest` has empty `CircleCIToken`/`Env` would close both holes at once.

internal/watchd/client.go
```diff
@@ -19,19 +19,75 @@ import (
...
+
+// doPing sends a /ping request with the given client and returns reachability
+// and the daemon's build identity.
+func doPing(client *http.Client) (bool, string) {
+	resp, err := client.Get("http://watchd/ping")
+	if err != nil {
+		return false, ""
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode != http.StatusOK {
```
schurchleycci, line 63: Collapsing every non-200 into `(false, "")` means a `401` from a mismatched `CHUNK_WATCHD_TCP_TOKEN` is reported as "no daemon exists", so `chunk` silently runs everything inline locally with nothing pointing at the token.

The status check moved here verbatim from the old `ping(sockPath)`, so the logic isn't new — but before this PR a 401 from watchd was unreachable, and `withBearerAuth` is the only thing that can produce one. The docs say to "set the same value on the daemon and all clients" with no tooling doing it, so set-on-daemon-forgotten-on-client needs no typo at all.

I'll note the counter-argument, which is decent: a daemon you can't authenticate to is functionally absent, and inline fallback is the documented resilient behaviour. `chunk watch` does surface it (`FetchSnapshot` preserves `resp.Status`) — it's specifically the validate path that's silent. Distinguishing 401 and emitting a one-line hint would be enough.

internal/watchd/client.go
```diff
@@ -19,19 +19,75 @@ import (
...
+	}
+	return doPing(client)
+}
+
 // FetchSnapshot connects to the running watch daemon and returns the current
 // snapshot for the given project roots. If roots is empty all known projects
 // are returned.
 func FetchSnapshot(roots []string) (Snapshot, error) {
-	sockPath, err := SocketPath()
+	client, err := daemonClient()
```
schurchleycci, line 87: Project identity is an absolute path string canonicalized on whichever host holds it, and there's no translation layer across the TCP boundary — so client-side roots and sandbox-side roots are different strings for the same project.

To be fair about scope, a verifier refuted the broadest version of this: plain `chunk watch` sends `nil` roots (`internal/ui/watch/fetch.go` only populates them when `!watchAll`), so the daemon's return-everything path applies. What survives:

- **`chunk watch --focus`** sends locally-canonicalized roots, matched by exact string equality in `daemon.snapshot` → empty `Projects`, HTTP 200, no error.
- **Command output, regardless of `--focus`**: `RegisterCommand` goes over TCP with a local `ProjectRoot`; the daemon keys it via `CanonicalProjectRoot`, which returns the path unchanged when `EvalSymlinks` fails on the sandbox — so `commandsFor(ps.root)` never matches and the logs pane stays blank.

Separately, `sidecar.AllProjectRoots()` drops roots failing `os.Stat`, so a sandbox daemon that hasn't run anything itself discovers zero projects — worth confirming the remote dashboard is expected to show anything at all yet.

internal/watchd/client.go
```diff
@@ -176,11 +222,7 @@ func StopForCredentialChange() {
 // from the same binary as the caller. A daemon from an older build is treated
 // as absent: it may not serve routes added since it was compiled.
 func IsDaemonRunning() bool {
-	sockPath, err := SocketPath()
-	if err != nil {
-		return false
-	}
-	ok, _ := ping(sockPath)
+	ok, _ := pingDaemon()
```
schurchleycci, line 286: Nit: this doc comment still says the daemon "was built from the same binary as the caller" and that "a daemon from an older build is treated as absent", but `IsDaemonRunning` is now ping-only — the build comparison lives in `IsDaemonCompatible` below. Stale in a way that matters, since the difference between these two is exactly what decides whether a hook delegates remotely.

internal/watchd/ipc_test.go
```diff
@@ -208,6 +209,113 @@ func TestSnapshotReportsAuthErrorWhenCredentialsAreMissing(t *testing.T) {
...
+	wrongClient := &http.Client{
+		Timeout:   5 * time.Second,
+		Transport: &bearerTransport{inner: transport, token: wrongToken},
+	}
+
+	ok, _ := doPing(wrongClient)
+	assert.Check(t, !ok, "expected ping to fail with wrong token")
+}
+
+func TestEnsureRunning_SkipsInRemoteMode(t *testing.T) {
```
schurchleycci, line 326: Nit (test isolation): these two set only `CHUNK_WATCHD_REMOTE_ADDR` — no `CHUNK_WATCHD_DIR`, no `HOME`, and there's no `TestMain` — so `watchdDir()` resolves to the developer's real `~/.chunk/watchd`.

As written they short-circuit and touch nothing, and to be fair they *do* fail if the guard regresses (`BuildID` can never match the test binary, so the nil-return path is unreachable) — a reviewer claim that they'd pass vacuously was refuted. But the latent hazard is real: without the guard, `EnsureRunning` would either SIGTERM a live local daemon via `stopDaemon` on build mismatch, or reach `launchDaemon` and spawn the *test binary* detached as `watch _daemon` against the real `watchd.log` — which re-runs the test and spawns its own child.

`t.Setenv("CHUNK_WATCHD_DIR", t.TempDir())` in each, plus asserting no pid file or socket appears there, fixes both.

internal/watchd/ipc_test.go
```diff
@@ -208,6 +209,113 @@ func TestSnapshotReportsAuthErrorWhenCredentialsAreMissing(t *testing.T) {
...
+	t.Cleanup(func() {
+		cancel()
+		select {
+		case <-errCh:
+		case <-time.After(5 * time.Second):
+			t.Error("daemon did not shut down within 5s")
+		}
+	})
+
+	deadline := time.Now().Add(5 * time.Second)
```
schurchleycci, line 250: Nit: this wait loop never selects on `errCh`, so a `RunDaemon` startup failure is discarded and the test fails 5s later with "tcp daemon did not become reachable within 5s" — pointing at the daemon rather than the real cause.

The bind-then-close port selection above is the case that motivates it, though a verifier found that race to be near-theoretical (no `t.Parallel()` in this package, ephemeral ports allocate monotonically, and `SO_REUSEADDR` means re-binding your own released port succeeds). The unread `errCh` is the part worth fixing — it improves diagnostics for *every* startup failure, not just port collisions.

schurchleycci (changes requested):
## Multi-agent review

6 independent review agents produced 26 raw findings; after dedup (13 unique claims) each was put to an adversarial verifier that had to construct a concrete failure scenario to confirm it. 1 claim was fully refuted and 5 partially refuted — those are not reported here.

The direction of this PR looks right, and the remote-mode guards in `EnsureRunning`/`EnsureLaunched` in particular held up under attack (a reviewer claim that they should ping first was refuted: the contract is launch-oriented, and adding a ping would put a network round trip in front of every command).

Requesting changes on two blockers:

1. **`withBearerAuth` fails open** — a TCP listener with no token set serves the full mux unauthenticated, and `/validate` runs a full root cobra command with caller-supplied args and env.
2. **`ResponseHeaderTimeout: 30s` on `longTCPClient`** — silently caps every remote validate at 30s and falls back to inline local execution while the remote run keeps going, holding `validateMu`.

The remaining comments are strong suggestions and nits. Happy to be wrong on any of these — where a verifier was less than certain I've said so inline.

danmux (changes requested):
Three issues before merging:

**ipc_test.go** — TOCTOU: port released with `ln.Close()` before the daemon rebinds. Another process can claim it in the window. Pass the already-open listener into the daemon, or retry on EADDRINUSE.

**client.go** — Remote daemon version mismatch silently falls back to inline local execution with no user signal. For local daemons a mismatch triggers a restart via EnsureRunning — for remote there is no equivalent. Should this be an explicit error?

**client.go** — CircleCIToken and Env are silently dropped from ValidateRequest when talking to a remote daemon. If the sidecar was launched without the needed token, the validate job fails inside the sandbox with no hint. Worth a warning or doc comment?

danmux (approved):

PR #592, opened 2026-09-16
Add pooled mutation testing for Go

z00b:
- enumerate deterministic Go mutation candidates with built-in AST operators
- create or reuse sidecar pools and schedule mutations as members become ready
- apply isolated patches, run configurable tests with timeouts, and restore workers
- report ordered killed, survived, and infrastructure-error results
- document the command and cover enumeration, patching, and bounded output


internal/mutate/mutate.go
```diff
@@ -0,0 +1,68 @@
...
+	Col      int    `json:"col"`
+	Operator string `json:"operator"`
+	Status   string `json:"status"` // RUNNABLE or NOT COVERED
+	Before   string `json:"before"` // original token, e.g. "<"
+	After    string `json:"after"`  // mutated token, e.g. "<="
+}
+
+// Enumerator produces a list of candidate mutations for a directory.
+type Enumerator interface {
+	Enumerate(ctx context.Context, dir string) ([]Mutation, error)
```
danmux: `Enumerator` has only one implementation. Project rule: no interface without two or more implementations. Pass `*GoEnumerator` directly or use a func parameter.

internal/cmd/mutate.go
```diff
@@ -0,0 +1,210 @@
...
+			stack := mutate.DetectedStack(workDir, cfg)
+			if stack == "" {
+				return &userError{
+					msg:    "Could not detect project language.",
+					errMsg: "language detection failed",
+					suggestion: "Set \"environment\": {\"stack\": \"go\"} in .chunk/config.json " +
+						"or ensure a marker file (go.mod, package.json, Cargo.toml) exists.",
+				}
+			}
+
```
danmux, line 57: Type-asserting the returned `Enumerator` back to `*mutate.GoEnumerator` to set Paths breaks the abstraction entirely. Make it a parameter to EnumeratorFor or Enumerate instead.

internal/cmd/mutate.go
```diff
@@ -0,0 +1,210 @@
...
+		Short: "Find test coverage gaps via mutation testing",
+		Long: `Enumerate candidate mutations in production code and validate them against
+the test suite using chunk sidecars. Survivors — mutations the suite fails to
+catch — reveal real coverage gaps.
+
+Without --parallelism, prints the mutation inventory and exits.
+With --parallelism N, creates a pool of N sidecars and runs each mutation
+against the test suite in parallel.`,
+		Args: cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
danmux, line 37: `cfg, _ :=` silently discards the LoadProjectConfig error. A malformed .chunk/config.json produces a confusing message instead of reporting the actual parse error.

danmux (changes requested):
Three issues to resolve before merging.

danmux (approved):

danmux (approved):

PR #593, opened 2026-09-16
remote daemon: spawn and use sidecars for validate runs

hanabel1:
## Summary

- Adds a `provisioner` to the daemon that creates and tracks sidecars owned by this daemon instance, with `stopAll` cleanup on shutdown
- New HTTP endpoints `POST /sidecar` and `DELETE /sidecar/{id}` let callers ask the daemon to create/delete sidecars using its own CircleCI credentials
- Wires `handleValidate` to auto-provision a fresh sidecar when `OrgID` is provided, injecting `--sidecar-id` into the validate args so the in-process runner syncs and uses it

## Events

- `POST /sidecar` → daemon calls `circleci.CreateSidecar`, tracks the ID internally, returns `{sidecar_id}` (503 when no credentials)
- `DELETE /sidecar/{id}` → daemon calls `circleci.DeleteSidecar` and untracks
- `POST /validate` with `org_id` set → daemon provisions a sidecar, appends `--sidecar-id <id>` to args, defers deletion after the run

## Not collected

- No image selection yet — provisioned sidecars use the org default (`image: ""`)
- `tryHookDelegate` passes `orgID: ""` so the hook early-exit path does not provision (degrades gracefully to existing behaviour)

## Opt-out

Provisioning only activates when both conditions are true: the daemon has credentials (`prov != nil`) and the validate request includes a non-empty `OrgID`. All existing code paths are unchanged.

## Implementation

`provisioner` is a thin wrapper around `circleci.Client` with a mutex-protected set of owned IDs. `stopAll` iterates the set and deletes everything — sidecars never outlive the daemon that created them.

`RunValidate` gains an `orgID string` parameter, forwarded in the request body only over TCP (the Unix-socket path already carries full env and credentials and does not need it).

## Test plan

- [ ] `go test -race ./internal/watchd/...` — provision/deprovision endpoint tests pass
- [ ] `go test -race ./internal/cmd/...` — validate call-site compilation and tests pass
- [ ] Manual: start daemon in sandbox with credentials, run `chunk validate` from local machine over TCP, confirm sidecar is created and cleaned up


internal/watchd/provision.go
```diff
@@ -0,0 +1,70 @@
...
+	p.owned[sc.ID] = true
+	p.mu.Unlock()
+	return sc.ID, nil
+}
+
+func (p *provisioner) delete(ctx context.Context, id string) error {
+	p.mu.Lock()
+	delete(p.owned, id)
+	p.mu.Unlock()
+	return p.client.DeleteSidecar(ctx, id)
```
michael-webster, line 53: 🔴 **Blocker: sidecar leaks permanently when `DeleteSidecar` fails**

The ID is removed from `p.owned` (line 51) _before_ calling `p.client.DeleteSidecar`. If the API call fails (network timeout, 5xx), the ID is already gone — `stopAll` on shutdown iterates `p.owned` and will never retry it.

The call site in `handleValidate` discards the error entirely (`_ = d.prov.delete(...)`), so this failure is completely silent. Confirmed by all 6 reviewers.

```go
func (p *provisioner) delete(ctx context.Context, id string) error {
    if err := p.client.DeleteSidecar(ctx, id); err != nil {
        return err
    }
    p.mu.Lock()
    delete(p.owned, id)
    p.mu.Unlock()
    return nil
}
```

internal/watchd/provision.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+func (p *provisioner) create(ctx context.Context, orgID, name, image string) (string, error) {
+	sc, err := p.client.CreateSidecar(ctx, orgID, name, image)
+	if err != nil {
+		return "", fmt.Errorf("provision sidecar: %w", err)
+	}
+	p.mu.Lock()
+	p.owned[sc.ID] = true
```
michael-webster, line 44: 🔴 **Blocker: race between `create` and `stopAll` — sidecar escapes cleanup on SIGINT**

`create` calls `CreateSidecar` outside the lock (line 39), then acquires the lock here to record the ID. There is no done flag or context check between the API return and `p.mu.Lock()`.

Concrete race: SIGINT arrives after `CreateSidecar` succeeds → `srv.Close()` fires → `prov.stopAll` acquires the lock, calls `clear(p.owned)`, releases it → the HTTP handler goroutine (still alive after `srv.Close()`) acquires the lock here and writes the new ID into the now-empty map. No subsequent cleanup sweep ever runs — the sidecar is permanently orphaned on CircleCI.

Fix: record the ID and check for a shutdown flag atomically, e.g. hold the lock across the API call, or track a `stopped bool` that `create` checks under the lock before writing.

internal/watchd/ipc.go
```diff
@@ -173,3 +176,57 @@ func longTCPClient(addr string) *http.Client {
...
+	if d.prov == nil {
+		http.Error(w, "daemon has no credentials", http.StatusServiceUnavailable)
+		return
+	}
+	id := strings.TrimPrefix(r.URL.Path, "/sidecar/")
+	if id == "" {
+		http.Error(w, "sidecar id required", http.StatusBadRequest)
+		return
+	}
+	if err := d.prov.delete(r.Context(), id); err != nil {
```
michael-webster, line 227: 🔴 **Blocker (Security): `DELETE /sidecar/{id}` has no ownership check — any daemon client can destroy arbitrary sidecars**

The caller-supplied `id` is passed directly to `d.prov.delete` with no check against `p.owned`. `delete` unconditionally calls `p.client.DeleteSidecar` for any ID, including ones the daemon never provisioned.

On the TCP path (`CHUNK_WATCHD_REMOTE_ADDR`), the bearer token is shared across all callers of the same daemon. Any authenticated client can destroy another user's sidecar by supplying an arbitrary ID (IDs are returned in API responses).

```go
// Check ownership before forwarding to the API:
if !p.owned[id] {
    http.Error(w, "sidecar not found", http.StatusNotFound)
    return
}
```
(Combine with the ordering fix on `delete`: ownership check first, map removal after successful API call.)

internal/watchd/client.go
```diff
@@ -277,6 +281,58 @@ func RunValidate(args []string, circleCIToken string) (ValidateResponse, error)
 	return result, nil
 }
 
+// ProvisionSidecar asks the remote daemon to create a new sidecar and return
+// its ID. Returns an error if the daemon is unavailable or unauthenticated.
+//
+// Only meaningful when CHUNK_WATCHD_REMOTE_ADDR is set; against a local Unix
+// socket the caller can create sidecars directly without going through the daemon.
+func ProvisionSidecar(req ProvisionRequest) (ProvisionResponse, error) {
+	client, err := daemonClient()
```
michael-webster, line 290: 🟡 **Strong suggestion: `ProvisionSidecar` uses the short 5s `daemonClient` — caller times out while daemon retains an orphaned sidecar**

`daemonClient()` sets `http.Client.Timeout: 5s`. The daemon handler calls `CreateSidecar` synchronously before responding. If CircleCI takes > 5s, the client transport drops the connection and the caller gets an error with no sidecar ID — but the daemon may have already called `CreateSidecar` and written the ID into `p.owned`. The daemon will clean it up via `stopAll` on shutdown, but the caller cannot use or explicitly delete it.

The analogous `RunValidate` explicitly uses `longDaemonClient()` for this reason (line 238). Should `ProvisionSidecar` get the same treatment?

internal/watchd/validate.go
```diff
@@ -62,8 +68,20 @@ func (d *daemon) handleValidate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	args := req.Args
+	if d.prov != nil && req.OrgID != "" {
+		name := fmt.Sprintf("validate-%x", time.Now().UnixNano())
+		id, err := d.prov.create(ctx, req.OrgID, name, "")
```
michael-webster, line 74: 🟡 **Strong suggestion: `validateMu` is now held across a remote API call**

`handleValidate` acquires `validateMu` exclusively (line 45) and holds it for the entire body, which now includes `d.prov.create` here — a synchronous CircleCI API call. Concurrent `/validate` requests from multiple agents hitting the same daemon queue behind each other for the full provisioning + run cycle.

The comment at lines 54–56 confirms serialisation is intentional. Worth calling out explicitly now that the critical section includes a remote API round-trip: if this daemon is ever intended to serve concurrent callers, provisioning could be moved before acquiring `validateMu`.

internal/watchd/provision_test.go
```diff
@@ -0,0 +1,192 @@
...
+	owned := len(p.owned)
+	p.mu.Unlock()
+	assert.Check(t, cmp.Equal(owned, 1))
+
+	p.stopAll(ctx)
+
+	p.mu.Lock()
+	remaining := len(p.owned)
+	p.mu.Unlock()
+	assert.Check(t, cmp.Equal(remaining, 0))
```
michael-webster, line 191: 🔵 **Nit: `stopAll` test only asserts in-memory state — server-side deletion is not verified**

The test asserts `len(p.owned) == 0` after `stopAll` but never checks that the fake CircleCI server actually received the delete requests. `fakes.NewFakeCircleCI()` exposes a public `Sidecars []Sidecar` field that `handleDeleteSidecar` updates — a regression that cleared the map without calling `DeleteSidecar` would pass undetected.

```go
p.stopAll(ctx)
assert.Check(t, cmp.Equal(len(cci.Sidecars), 0), "expected all sidecars deleted from API")
```

michael-webster (changes requested):
Multi-agent review (6 reviewers + 7 adversarial verifiers). Three blockers confirmed — requesting changes before merge.

🔴 **Blockers (3)**: sidecar leak on delete failure, create/stopAll race on shutdown, unowned sidecar deletion
🟡 **Strong suggestions (2)**: ProvisionSidecar timeout, validateMu scope
🔵 **Nit (1)**: stopAll test assertion gap

internal/watchd/ipc.go
```diff
@@ -80,12 +81,44 @@ func newServer(d *daemon) *http.Server {
...
+// withBearerAuth wraps h so that every request must carry a matching
+// Authorization: Bearer <token> header. When token is empty the handler is
+// returned unchanged (auth disabled).
+func withBearerAuth(h http.Handler, token string) http.Handler {
+	if token == "" {
+		return h
+	}
+	want := "Bearer " + token
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Header.Get("Authorization") != want {
```
danmux: Plain  comparison for a shared secret is vulnerable to timing side-channels. Use subtle.ConstantTimeCompare([]byte(got), []byte(want)) != 1?

internal/watchd/ipc.go
```diff
@@ -109,3 +142,91 @@ func longUnixClient(sockPath string) *http.Client {
...
+		rt = &bearerTransport{inner: transport, token: token}
+	}
+	return &http.Client{Timeout: 5 * time.Second, Transport: rt}
+}
+
+// longTCPClient is like tcpClient but with no total-request timeout, suitable
+// for long-running operations like /validate. A 10 s dial timeout bounds how
+// long an unreachable sandbox can stall the caller before ErrDaemonUnavailable
+// triggers a fallback to inline execution.
+func longTCPClient(addr string) *http.Client {
```
danmux, line 184: ResponseHeaderTimeout: 30s kills any validate run that takes longer than 30s to write response headers - the server only writes headers after the subprocess finishes. This triggers ErrDaemonUnavailable and silent fallback, defeating the remote daemon for slow checks. Should be unset like longUnixClient.

internal/watchd/ipc_test.go
```diff
@@ -206,6 +207,113 @@ func TestSnapshotReportsAuthErrorWhenCredentialsAreMissing(t *testing.T) {
...
+// TCP, and verify that the bearer-token guard is enforced.
+func startTestDaemonTCP(t *testing.T, token string) string {
+	t.Helper()
+	dir, err := os.MkdirTemp("", "wd-tcp")
+	assert.NilError(t, err)
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	t.Setenv("CHUNK_WATCHD_DIR", dir)
+	t.Setenv("XDG_DATA_HOME", t.TempDir())
+
+	// Pick a free port by binding on :0 and immediately releasing it.
```
danmux, line 224: TOCTOU: port released with ln.Close() before the daemon binds. Another process can claim it in the window. Can the daemon accept a net.Listener directly to hand over the already-bound socket?

danmux (changes requested):
Two issues to fix before merging.

danmux:
Temporarily closing to unstack from #591 so it can merge. Will reopen after rebasing on main.

PR #602, opened 2026-09-21
Use session-scoped anonymous ID for telemetry

schurchleycci:
## Summary

- Adds `config.SessionTrackingID(sessionID string) uuid.UUID` — derives a stable, anonymous UUID from a session ID string via UUID v5, so the same session always yields the same tracking UUID
- Adds `SessionTrackingID` field to `telemetry.Meta`; when set, it is used as `AnonymousId` in Segment events instead of the per-install `InstanceID` (which remains in `context.device.id` for install-level attribution)
- In `setupTelemetry`, reads the active session ID from `session.IDFromEnv()` (respects `CHUNK_SESSION_ID` and `CLAUDE_CODE_SESSION_ID`) and populates `SessionTrackingID` when a session is active

The effect: multiple `chunk` invocations within the same Claude Code (or manually-set) session now share a common anonymous identifier, making it possible to correlate events at the session level rather than only at the install level.

## Test plan

- `TestSender_Track_WithSessionTrackingID` — verifies the session ID is used as `AnonymousId` and the install ID stays in `context.device.id`
- `TestSessionTrackingID_Deterministic` — same input always yields the same UUID
- `TestSessionTrackingID_DifferentInputsDifferentIDs` — distinct sessions get distinct IDs
- `TestSessionTrackingID_NonNil` — result is never the zero UUID
- All 1914 existing tests pass; lint clean

Closes FACT-479


internal/config/config.go
```diff
@@ -300,6 +300,18 @@ func IsTelemetry(cfg UserConfig) bool {
 	return *cfg.Telemetry
 }
 
+// chunkTelemetryNamespace is a fixed UUID used as the v5 namespace when
+// deriving per-session anonymous tracking IDs. It uniquely scopes the hash
+// domain to chunk-cli so session IDs from other tools never collide.
+var chunkTelemetryNamespace = uuid.MustParse("f47ac10b-58cc-4372-a567-0e02b2c3d479")
```
danmux, line 306: The namespace UUID is the RFC 4122 Appendix C example — it appears verbatim in countless tutorials and docs. The uniqueness guarantee only holds if no other tool grabbed it too. Shouldn't this be a freshly generated UUID specific to chunk-cli?

danmux (approved):

danmux (approved):

internal/config/config.go
schurchleycci: I'll fix this up now!

PR #603, opened 2026-09-21
Use a chunk-specific UUID namespace for session tracking IDs

schurchleycci:
## Summary

Follow-up to [#602](https://github.com/CircleCI-Public/chunk-cli/pull/602), addressing @danmux's review comment on `internal/config/config.go`.

The v5 namespace used to derive session tracking IDs was `f47ac10b-58cc-4372-a567-0e02b2c3d479` — the RFC 4122 Appendix C example UUID, which appears verbatim in countless tutorials and docs. Namespacing the hash domain only buys uniqueness if the namespace is actually ours, so reusing a published example defeats the point.

Replaced with a UUID generated specifically for chunk-cli, and expanded the comment to record why it's tool-specific and that changing it would orphan previously reported session IDs.

No behavioral change beyond the namespace value: `SessionTrackingID` remains deterministic and non-reversible.

## Test plan

- `go test -race ./internal/config/... ./internal/telemetry/...` passes
- No test or doc hardcoded the old namespace value, so nothing else needed updating

## Note

`task lint` and the build currently fail on `main` for an unrelated reason: `internal/cmd/mutate.go:129` still passes `AuthSock: os.Getenv(config.EnvSSHAuthSock)` to `sidecar.PoolOptions`, but neither `PoolOptions.AuthSock` nor `config.EnvSSHAuthSock` exists anymore (looks like fallout from the pools cleanup in #587). That break is independent of this change but will fail CI here until it's fixed on `main`.


danmux (approved):
Rebased onto main (picks up the prm fix from #607). Tests green.

PR #605, opened 2026-09-21
"sql" in a config is now an error, not a silent no-op

jesseworld22:
## Summary

Follow-up to #589, resolving both of @danmux's review comments.

- **`"sql"` is now rejected.** `validatePathRules` named a bare extension as a mistake in its own doc comment but never checked for it — no slash, no star, so it was stored as an exact file name and waited for a file called `sql` while every migration passed unnoticed. It's a list of extension names rather than a shape test, because `sql` and `mvnw` have the same shape and a shape test would lock out the real file name.
- **Extensions fold case, file names don't — on purpose, and now said so.** `.MD` and `.md` are the same kind of file by convention. `notice` and `NOTICE` are two different paths to git, so folding them would match files the project never named, and for the inert list that's the direction that releases a change and loses a failure.
- **Plus a build fix `main` needs.** #574 deleted `PoolOptions.AuthSock` and `config.EnvSSHAuthSock`; the pooled mutation command was written before that landed and merged after, so `internal/cmd/mutate.go:129` referenced both and `main` stopped compiling. Happy to split this into its own PR if you'd rather it land separately.

## Test plan

- [x] `task test` — 2055 tests, `-race`, green
- [x] `task lint` — 0 issues
- [x] `task ci:diff` — clean
- [x] `task build` — binary runs
- [x] New cases: bare `sql` and `SQL` rejected; `Makefile` and `mvnw` still load; `.SQL` matches `schema.sql`, `notice` doesn't match `NOTICE`


internal/config/project.go
```diff
@@ -266,11 +266,36 @@ func validatePathRules(key string, entries []string) error {
...
 
+// knownExtensions are extension names, without their leading dot, that are
+// rejected when written bare. An entry with no dot is stored as an exact file
+// name, so "sql" waits for a file called "sql" and never sees a migration.
+//
+// It is a list of names rather than a test of shape because the two shapes are
+// the same: "sql" and "mvnw" are both lowercase letters with no dot, and a shape
+// test would reject the second with no way for a project to insist. A list only
+// ever misses one, and the ones it holds are the ones somebody writes.
+var knownExtensions = map[string]bool{
```
tshen815, line 308: It doesnt look like we're trimming the whitespace for these extensions so `"sql  " ` would slip past this check. 

we could add `strings.TrimSpace` over both lists in `LoadProjectConfig` before `validate()` runs. I know you didnt introduce this change but just wanted to call it out.

danmux (approved):
Factory: rebased (3 commits, 1 dropped as already upstream), remote validated clean — 2154 tests pass.

